Issue #314, opened 2014-11-17
The minified file used for the client side should be included in this lib


totty90:
As every proper repo, the min file should be already included. Don't expect every dev can build it (might use win and missing some build tools). 

Put the min file in this folder '/dist' and can be named 'bitcoinjs-lib.min.js'.

Edit: Seems like you have already talked about this: https://github.com/bitcoinjs/bitcoinjs-lib/pull/252 but you closed. I have to say that I've never found a bower component that requires any build step before use. Take a look at every client side javascript lib (not only bower) and you will see that they all have a min file bundled. You can create a script to run every time you want to publish a new version that creates the min file and do other stuff you need. 

If you don't put a min file then many client side dev will not use this (I don't know if you care or not, but I hope you do).

davidapple:
I understand your security concerns. However, having to compile this locally is a massive pain. I am a proficient web developer, however, I routinely work on machines that throw errors when running npm. This is a massive barrier to entry for a lot of developers and a huge time sponge.

Please make a compiled version of bitcoinjs-lib v4.0.2 available from somewhere. Even if it's from another repository. Nearly all other js libraries include a dist folder with a minified version of the library pre-compiled. It's standard and saves a huge amount of time.

I'm sure people using this for serious financial applications are aware of the risks. But forcing people to compile every time is blocking the majority of developers that want to play.

dcousens:
Half the libraries interface is using a Buffer.  How do you expect to use it?

dcousens:
I don't think any new argument has been presented.
Should we lock this one?

jprichardson:
> Should we lock this one?

Yes. 

@davidapple A new issue should be created to discuss this. Including a bundled version carries an additional audit burden for consumers of the package. Also, there's a lot of packaged overlap due to the reliance upon Node.js specific libs which means the payload size would generally be quite large and unnecessary for most. I realize it's painful for you, but these days it's inescapable for web devs to avoid using npm; so much tooling depends upon it.

PR #315, opened 2014-11-18
README: improve browser setup advice

dcousens:
As per https://github.com/bitcoinjs/bitcoinjs-lib/issues/314, this updates the README to include more up-to-date and idiomatic build advice for users who want to use the library from the browser.

Suggestions welcome and appreciated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,29 +56,27 @@ These steps are advisory only and allow you to use the API to its full extent.
 
 [Browserify](https://github.com/substack/node-browserify) is assumed to be installed for these steps.
 
-From your repository, create a `bitcoin.js` file
+From your repository, create a `foobar.js` file
 
 ``` javascript
-var bitcoin = {
+var foobar = {
   base58: require('bs58'),
+  bitcoin: require('bitcoinjs-lib'),
   ecurve: require('ecurve'),
   BigInteger: require('bigi'),
   Buffer: require('buffer')
 }
 
-var bitcoinjs = require('bitcoinjs-lib')
-for (var a in bitcoinjs) bitcoin[a] = bitcoinjs[a]
-
-module.exports = bitcoin
+module.exports = foobar
 ```
 
-Then, using browserify, compile `bitcoin.js` for use in the browser:
+Then, using browserify, compile `foobar.js` for use in the browser:
 
-    $ browserify bitcoin.js -s bitcoin > dist/bitcoin.js
+    $ browserify foobar.js -s foobar > dist/foobar.js
 
-You will then be able to load `dist/bitcoin.js` into your browser, with each of the dependencies above accessible from the global `bitcoin` object.
+You will then be able to load `dist/foobar.js` into your browser, with each of the dependencies above accessible from the global `foobar` object.
 
-**NOTE**: See the package.json for the currently supported version of browserify used by this repository.
+**NOTE**: See our package.json for the currently supported version of browserify used by this repository.
 
 
 ## Examples
--- a/README.md
+++ b/README.md
@@ -66,11 +66,14 @@ var foobar = {
 module.exports = foobar
 ```
 
-Then, using browserify, compile `foobar.js` for use in the browser:
+Each of these included packages are seperate to `bitcoinjs-lib`, and must be installed separately.
+They are however used in the bitcoinjs-lib public API.
 
-    $ browserify foobar.js -s foobar > dist/foobar.js
+Using browserify, compile `foobar.js` for use in the browser:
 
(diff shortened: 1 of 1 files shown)


dcousens:
@weilu I've updated the instructions.

I believe I've kept true to your intentions,

> the use of browserify directly without having its version managed by a dependency manager.
see https://github.com/bitcoinjs/bitcoinjs-lib/commit/df50393727979008098362b8ae9360686907cdcf

> buffer is a generic function that should not be living under the bitcoin namespace
see https://github.com/bitcoinjs/bitcoinjs-lib/commit/1d9bc2cc4c3468eb18de7dcc975963f7156cc7f0

I think the above makes it abundantly clear we are just helping the user understand how to user browserify, and not just providing a convenient hack.

README.md
```diff
@@ -51,16 +51,32 @@ From the repo:
...
 
-From NPM:
+From your repository, create a `foobar.js` file
 
-    $ npm -g install bitcoinjs-lib browserify uglify-js
-    $ browserify -r bitcoinjs-lib -s bitcoin | uglifyjs > bitcoinjs.min.js
+``` javascript
+var foobar = {
+  base58: require('bs58'),
+  bitcoin: require('bitcoinjs-lib'),
```
weilu: it will complain that `bitcoinjs-lib` is not found as it has not yet been installed through npm

README.md
```diff
@@ -51,16 +51,32 @@ From the repo:
...
+  ecurve: require('ecurve'),
+  BigInteger: require('bigi'),
+  Buffer: require('buffer')
+}
 
-After loading this file in your browser, you will be able to use the global `bitcoin` object.
+module.exports = bitcoin
```
weilu: `foobar` you mean? You also don't have to give it such an awkward name. 

```
module.exports = {
  base58: require('bs58'),
  bitcoin: require('bitcoinjs-lib'),
  ecurve: require('ecurve'),
  BigInteger: require('bigi'),
  Buffer: require('buffer')
}
```

README.md
dcousens: Same with `bs58`, and any others.

weilu: So? the instruction is still incomplete

dcousens: Actually, it has been installed.

This is under the `## Setup` header, with `## Installation` above it.
Therefore, at this stage, it is assumed they have `bitcoinjs-lib` installed.

README.md
```diff
@@ -44,23 +44,35 @@ If you are looking for the original, it is tagged as `0.1.3`. Unless you need it
...
 
-From the repository: Compile `bitcoinjs-min.js` with the following command:
+module.exports = foobar
+```
 
-    $ npm run-script compile
+Then, using browserify, compile `foobar.js` for use in the browser:
 
-From NPM:
+    $ browserify foobar.js -s foobar > dist/foobar.js
```
weilu:
```
mkdir node_sandbox
cd node_sandbox/
npm init
npm install bitcoinjs-lib
browserify foobar.js -s foobar > dist/foobar.js
-bash: dist/foobar.js: No such file or directory
```

weilu:
```
browserify foobar.js -s foobar > bundle.js
Error: Cannot find module 'bs58' from '/Users/luwei/workspace/node_sandbox'
    at /usr/local/lib/node_modules/browserify/node_modules/resolve/lib/async.js:50:17
    at process (/usr/local/lib/node_modules/browserify/node_modules/resolve/lib/async.js:119:43)
    at /usr/local/lib/node_modules/browserify/node_modules/resolve/lib/async.js:128:21
    at load (/usr/local/lib/node_modules/browserify/node_modules/resolve/lib/async.js:60:43)
    at /usr/local/lib/node_modules/browserify/node_modules/resolve/lib/async.js:66:22
    at /usr/local/lib/node_modules/browserify/node_modules/resolve/lib/async.js:21:47
    at Object.oncomplete (fs.js:107:15)
```

weilu: @dcousens Did I fail at reading the readme?

dcousens: @weilu try now after 642315e

dcousens:
> Again, how do you control versions of those libraries?
That is a good point.  I guess it becomes obvious in your examples, but this problem is independent of this PR as it also affects the node set up advice.

How do we control users using `bigi@latest` in conjunction with our API instead of what we use internally?
The best we can do is either simplify our API such that common JS types are used or we ask the users to verify their dependencies against what we use internally.

dcousens:
I'm going to merge this, but the point brought up by @weilu is quite valid.
I'll summarize and it make another issue.

``` javascript
How do we control users using bigi@latest in conjunction with our API instead of what we use internally?
The best we can do is either simplify our API such that common JS types are used for data interchange or we ask the users to verify their dependencies against what we use internally.
```

PR #336, opened 2015-01-02
Make RFC6979 compliant + loop through bad r or s.

bip32JP:
I have made the same pull request to bitcore in regards to the missing HMAC.
As well as I have also made an additional request to blockchain.info MyWallet to account for looping Step H when bad r or s value is there.

Please check here for an explanation for the extra V = HMAC_K(V).
bitpay/bitcore#884

As for the bad r or s: (I also explain this here blockchain/My-Wallet#115 )
RFC specifies that if the k value is out of bounds OR if r or s is 0 you must loop through Step h until a proper value is found. Not throw an error.

The way bitcore implements this is by including an integer called "badrs" that will force entering the re-hashing loop of Step h. The incrementing of badrs is done in a do while loop around k generation and r and s calculation.

This is not a large issue, as the only possible vector of attack for it as is... is if someone had a transaction + privkey pair that just happened to produce an out of bounds k... AND they decided to sign that transaction using both bitcoinjs-lib AND Electrum (uses python-ecdsa, 100% RFC6979 compliant) and then for some reason posted those two signed transactions in the public space somehow.

Either way, this needs to be fixed in order to be RFC 6979 compliant.

Thank you.

### DIFF
--- a/src/ecdsa.js
+++ b/src/ecdsa.js
@@ -9,12 +9,15 @@ var ZERO = new Buffer([0])
 var ONE = new Buffer([1])
 
 // https://tools.ietf.org/html/rfc6979#section-3.2
-function deterministicGenerateK(curve, hash, d) {
+function deterministicGenerateK(curve, hash, d, badrs) {
   typeForce('Buffer', hash)
   typeForce('BigInteger', d)
 
   // sanity check
   assert.equal(hash.length, 32, 'Hash must be 256 bit')
+
+  // badrs forces Step H loop when bad r or s value is given
+  badrs = badrs || 0
 
   var x = d.toBuffer(32)
   var k = new Buffer(32)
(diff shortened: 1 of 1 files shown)


dcousens:
Can't accept this until we have some test cases... but I'll see if I can produce them.  Reading through https://github.com/bitpay/bitcore/pull/884 at the moment.

dcousens:
I think https://github.com/indutny/elliptic/blob/master/lib/elliptic/hmac-drbg.js captures this process a bit nicer than providing the external `badrs` state; it might be better to focus on moving to that instead.

dcousens:
I think @indutny 's solution by using the HmacDRBG [path](https://github.com/indutny/elliptic/blob/master/lib/elliptic/hmac-drbg.js) is far superior in this case,  and better captures the carried state of `K`, `V` and `T` in the bad states.
For now, I might merge this or something similar, depending on how much time it takes.

bip32JP:
Closed, continue discussion in #337

PR #337, opened 2015-01-04
RF6979 compliance in regards to invalid ECDSA signatures

dcousens:
This pull request fixes a bug presented in #336 such that an out of bounds `k` would not be fully compliant with other RFC6979 implementations, and therefore would present an opportunity (in very irregular circumstances) for the `k` values produced by both implementations to be different.
As pointed out in #336, this could lead to an attack vector if both signatures were exposed in the public space somehow.

~~This PR is not yet complete, as it is awaiting test cases.~~

### DIFF
--- a/src/ecdsa.js
+++ b/src/ecdsa.js
@@ -20,6 +20,7 @@ function deterministicGenerateK(curve, hash, d) {
   var k = new Buffer(32)
   var v = new Buffer(32)
 
+  // Step A, ignored as hash already provided
   // Step B
   v.fill(1)
 
--- a/src/ecdsa.js
+++ b/src/ecdsa.js
@@ -12,9 +12,10 @@ var ONE = new Buffer([1])
 function deterministicGenerateK(curve, hash, d, checkSig) {
   typeForce('Buffer', hash)
   typeForce('BigInteger', d)
+
+  // FIXME: remove/uncomment for 2.0.0
 //  typeForce('Function', checkSig)
 
-  // FIXME: remove in 2.0.0
   if (typeof checkSig !== 'function') {
     console.warn('deterministicGenerateK requires a checkSig callback in 2.0.0, see #337 for more information')
 
--- a/src/ecdsa.js
+++ b/src/ecdsa.js
@@ -9,9 +9,10 @@ var ZERO = new Buffer([0])
 var ONE = new Buffer([1])
 
 // https://tools.ietf.org/html/rfc6979#section-3.2
-function deterministicGenerateK(curve, hash, d) {
+function deterministicGenerateK(curve, hash, d, checkSig) {
   typeForce('Buffer', hash)
   typeForce('BigInteger', d)
+  typeForce('Function', checkSig)
 
   // sanity check
   assert.equal(hash.length, 32, 'Hash must be 256 bit')
@@ -55,8 +56,8 @@ function deterministicGenerateK(curve, hash, d) {
 
   var T = BigInteger.fromBuffer(v)
 
-  // Step H3, repeat until T is within the interval [1, n - 1]
-  while ((T.signum() <= 0) || (T.compareTo(curve.n) >= 0)) {
+  // Step H3, repeat until T is within the interval [1, n - 1] and is suitable for ECDSA
+  while ((T.signum() <= 0) || (T.compareTo(curve.n) >= 0) || !checkSig(T)) {
     k = crypto.createHmac('sha256', k)
       .update(v)
       .update(ZERO)
@@ -64,6 +65,9 @@ function deterministicGenerateK(curve, hash, d) {
 
     v = crypto.createHmac('sha256', k).update(v).digest()
 
+    // Step H1/H2a, again, ignored as tlen === qlen (256 bit)
+    // Step H2b again
+    v = crypto.createHmac('sha256', k).update(v).digest()
     T = BigInteger.fromBuffer(v)
   }
 
@@ -71,18 +75,28 @@ function deterministicGenerateK(curve, hash, d) {
 }
 
 function sign(curve, hash, d) {
-  var k = deterministicGenerateK(curve, hash, d)
+  var r, s
 
+  var e = BigInteger.fromBuffer(hash)
   var n = curve.n
   var G = curve.G
-  var Q = G.multiply(k)
-  var e = BigInteger.fromBuffer(hash)
 
-  var r = Q.affineX.mod(n)
-  assert.notEqual(r.signum(), 0, 'Invalid R value')
+  deterministicGenerateK(curve, hash, d, function(k) {
+    var Q = G.multiply(k)
 
-  var s = k.modInverse(n).multiply(e.add(d.multiply(r))).mod(n)
-  assert.notEqual(s.signum(), 0, 'Invalid S value')
+    if (curve.isInfinity(Q))
+      return false
+
+    r = Q.affineX.mod(n)
+    if (r.signum() === 0)
+      return false
+
+    s = k.modInverse(n).multiply(e.add(d.multiply(r))).mod(n)
+    if (s.signum() === 0)
+      return false
+
+    return true
+  })
 
   var N_OVER_TWO = n.shiftRight(1)
 
--- a/src/ecdsa.js
+++ b/src/ecdsa.js
@@ -12,7 +12,33 @@ var ONE = new Buffer([1])
 function deterministicGenerateK(curve, hash, d, checkSig) {
   typeForce('Buffer', hash)
   typeForce('BigInteger', d)
-  typeForce('Function', checkSig)
+//  typeForce('Function', checkSig)
+
+  // FIXME: remove in 2.0.0
+  if (typeof checkSig !== 'function') {
+    console.warn('deterministicGenerateK requires a checkSig callback in 2.0.0, see #337 for more information')
+
+    checkSig = function(k) {
+      var G = curve.G
+      var n = curve.n
+      var e = BigInteger.fromBuffer(hash)
+
+      var Q = G.multiply(k)
+
+      if (curve.isInfinity(Q))
+        return false
+
+      var r = Q.affineX.mod(n)
+      if (r.signum() === 0)
+        return false
+
+      var s = k.modInverse(n).multiply(e.add(d.multiply(r))).mod(n)
+      if (s.signum() === 0)
+        return false
+
+      return true
+    }
+  }
 
   // sanity check
   assert.equal(hash.length, 32, 'Hash must be 256 bit')
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "bs58check": "1.0.3",
     "crypto-browserify": "^3.2.6",
     "ecurve": "1.0.0",
-    "typeforce": "0.0.2"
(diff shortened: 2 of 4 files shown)


dcousens:
@ryanxcharles, @bip32JP

I've posted this solution as a clearer solution which IMHO better represents my interpretation of RFC6979 without following the HmacDBRG route of maintaining state.

This solution also avoids the situation that the entire loop is maintained, and instead goes for a 'yield' approach by using a `checkSig` function.

dcousens:
@ryanxcharles, @bip32JP

I've posted this solution as a clearer solution which IMHO better represents my interpretation of RFC6979 without following the HmacDBRG route of maintaining state.

This solution also avoids the situation that the entire process is repeated for each failure, and instead uses a yield-like approach by using a `checkSig` callback.

dcousens:
@weilu this is a breaking change, but also a fix.  Whats the result for semver?

bip32JP:
These tests are using the i < badrs method. Generated on python-ecdsa and checked on bitcore's implementation (where I forced badrs in each case to 0, 1, and 15 respectively.

One drawback to the badrs method is that while a badrs of 1 SHOULD mean just 1 re-loop of Step H, in the (extremely unlikely) event of an out of bounds k on the last loop (if badrs = 1, then if the resulting k is out of bounds) then it would re-loop 2 times even though badrs is 1.

However, if, for instance, the 3rd reloop of the badrs 15 test was an out of bounds k, it wouldn't matter, as i is incremented either way, so we can be certain that badrs = 15 means that it was relooped 15 times as long as the 15th reloop did not result in out of bounds k..

But to be honest, RFC6979 is almost impossible to test for errors anyways with secp256k1 as hlen == qlen and not only that but the digest range and the order are so close... so a 1 in 2^128 chance that badrs x may not represent x loops, but rather x + 1 loops, is better than "we can't test it, so let it fail." imo.

I am not sure if you can forcibly cause your checkSig to fail x times... but if you can, here are the vectors I have made using python-ecdsa.

```javascript
[
    {
        'message': 'test data',
        'privkey': 'fee0a1f7afebf9d2a5a80c0c98a31c709681cce195cbcd06342b517970c0be1e',
        'k_bad00': 'fcce1de7a9bcd6b2d3defade6afa1913fb9229e3b7ddf4749b55c4848b2a196e',
        'k_bad01': '727fbcb59eb48b1d7d46f95a04991fc512eb9dbf9105628e3aec87428df28fd8',
        'k_bad15': '398f0e2c9f79728f7b3d84d447ac3a86d8b2083c8f234a0ffa9c4043d68bd258'
    },
    {
        'message': 'Everything should be made as simple as possible, but not simpler.',
        'privkey': '0000000000000000000000000000000000000000000000000000000000000001',
        'k_bad00': 'ec633bd56a5774a0940cb97e27a9e4e51dc94af737596a0c5cbb3d30332d92a5',
        'k_bad01': 'df55b6d1b5c48184622b0ead41a0e02bfa5ac3ebdb4c34701454e80aabf36f56',
        'k_bad15': 'def007a9a3c2f7c769c75da9d47f2af84075af95cadd1407393dc1e26086ef87'
    },
    {
        'message': 'Satoshi Nakamoto',
        'privkey': '0000000000000000000000000000000000000000000000000000000000000002',
        'k_bad00': 'd3edc1b8224e953f6ee05c8bbf7ae228f461030e47caf97cde91430b4607405e',
        'k_bad01': 'f86d8e43c09a6a83953f0ab6d0af59fb7446b4660119902e9967067596b58374',
        'k_bad15': '241d1f57d6cfd2f73b1ada7907b199951f95ef5ad362b13aed84009656e0254a'
    },
    {
        'message': 'Diffie Hellman',
        'privkey': '7f7f7f7f7f7f7f7f7f7f7f7f7f7f7f7f7f7f7f7f7f7f7f7f7f7f7f7f7f7f7f7f',
        'k_bad00': 'c378a41cb17dce12340788dd3503635f54f894c306d52f6e9bc4b8f18d27afcc',
        'k_bad01': '90756c96fef41152ac9abe08819c4e95f16da2af472880192c69a2b7bac29114',
        'k_bad15': '7b3f53300ab0ccd0f698f4d67db87c44cf3e9e513d9df61137256652b2e94e7c'
    },
    {
        'message': 'Japan',
        'privkey': '8080808080808080808080808080808080808080808080808080808080808080',
        'k_bad00': 'f471e61b51d2d8db78f3dae19d973616f57cdc54caaa81c269394b8c34edcf59',
        'k_bad01': '6819d85b9730acc876fdf59e162bf309e9f63dd35550edf20869d23c2f3e6d17',
        'k_bad15': 'd8e8bae3ee330a198d1f5e00ad7c5f9ed7c24c357c0a004322abca5d9cd17847'
    },
    {
        'message': 'Bitcoin',
        'privkey': 'fffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364140',
        'k_bad00': '36c848ffb2cbecc5422c33a994955b807665317c1ce2a0f59c689321aaa631cc',
        'k_bad01': '4ed8de1ec952a4f5b3bd79d1ff96446bcd45cabb00fc6ca127183e14671bcb85',
        'k_bad15': '56b6f47babc1662c011d3b1f93aa51a6e9b5f6512e9f2e16821a238d450a31f8'
    },
    {
        'message': 'i2FLPP8WEus5WPjpoHwheXOMSobUJVaZM1JPMQZq',
        'privkey': 'fffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364140',
        'k_bad00': '6e9b434fcc6bbb081a0463c094356b47d62d7efae7da9c518ed7bac23f4e2ed6',
        'k_bad01': 'ae5323ae338d6117ce8520a43b92eacd2ea1312ae514d53d8e34010154c593bb',
        'k_bad15': '3eaa1b61d1b8ab2f1ca71219c399f2b8b3defa624719f1e96fe3957628c2c4ea'
    },
    {
        'message': 'lEE55EJNP7aLrMtjkeJKKux4Yg0E8E1SAJnWTCEh',
        'privkey': '3881e5286abc580bb6139fe8e83d7c8271c6fe5e5c2d640c1f0ed0e1ee37edc9',
        'k_bad00': '5b606665a16da29cc1c5411d744ab554640479dd8abd3c04ff23bd6b302e7034',
        'k_bad01': 'f8b25263152c042807c992eacd2ac2cc5790d1e9957c394f77ea368e3d9923bd',
        'k_bad15': 'ea624578f7e7964ac1d84adb5b5087dd14f0ee78b49072aa19051cc15dab6f33'
    },
    {
        'message': '2SaVPvhxkAPrayIVKcsoQO5DKA8Uv5X/esZFlf+y',
        'privkey': '7259dff07922de7f9c4c5720d68c9745e230b32508c497dd24cb95ef18856631',
        'k_bad00': '3ab6c19ab5d3aea6aa0c6da37516b1d6e28e3985019b3adb388714e8f536686b',
        'k_bad01': '19af21b05004b0ce9cdca82458a371a9d2cf0dc35a813108c557b551c08eb52e',
        'k_bad15': '117a32665fca1b7137a91c4739ac5719fec0cf2e146f40f8e7c21b45a07ebc6a'
    },
    {
        'message': '00A0OwO2THi7j5Z/jp0FmN6nn7N/DQd6eBnCS+/b',
        'privkey': '0d6ea45d62b334777d6995052965c795a4f8506044b4fd7dc59c15656a28f7aa',
        'k_bad00': '79487de0c8799158294d94c0eb92ee4b567e4dc7ca18addc86e49d31ce1d2db6',
        'k_bad01': '9561d2401164a48a8f600882753b3105ebdd35e2358f4f808c4f549c91490009',
        'k_bad15': 'b0d273634129ff4dbdf0df317d4062a1dbc58818f88878ffdb4ec511c77976c0'
    }
]
```

bip32JP:
Also, @dcousens the title should be `...in regards to invalid r and s values`

dcousens:
@bip32JP thanks for the test vectors.  I've added them in 0e5c7b2 with the hopes I'll refactor those tests at some point to be a little clearer.  In any case, seems pretty solid now :+1:

dcousens:
Good to merge, just want 1 more ACK.

ryanxcharles:
Agreed that this solution flows better than the badrs solution. However, it is still a bit hard to read, IMO. Though I think that may just be RFC 6979's fault.

src/ecdsa.js
```diff
@@ -70,18 +75,28 @@ function deterministicGenerateK(curve, hash, d) {
...
 
+  var e = BigInteger.fromBuffer(hash)
   var n = curve.n
   var G = curve.G
-  var Q = G.multiply(k)
-  var e = BigInteger.fromBuffer(hash)
 
-  var r = Q.affineX.mod(n)
-  assert.notEqual(r.signum(), 0, 'Invalid R value')
+  deterministicGenerateK(curve, hash, d, function(k) {
```
weilu: Can we make this function the default callback function for `deterministicGenerateK ` if none is provided? That way we can be backward compatible.

dcousens: @weilu that would give people the same behavior as before, which is incorrect.  A `checkSig` callback *is* required, that is the point.

weilu: Different behavior but a same interface. A default `checkSig` is still a check. What's wrong with that?

dcousens: @weilu oh, you mean put the entire of the signature generation function in `deterministicGenerateK`?  That is ugly as hell. :/

weilu: Right, as `deterministicGenerateK` is defined and used in the context of this library. I don't see there's anything wrong with supplying a default signature validation function in order to preserve backward compatibility

src/ecdsa.js
```diff
@@ -9,9 +9,36 @@ var ZERO = new Buffer([0])
...
+function deterministicGenerateK(curve, hash, d, checkSig) {
   typeForce('Buffer', hash)
   typeForce('BigInteger', d)
+//  typeForce('Function', checkSig)
+
+  // FIXME: remove in 2.0.0
+  if (typeof checkSig !== 'function') {
+    console.warn('deterministicGenerateK requires a checkSig callback in 2.0.0, see #337 for more information')
+
+    checkSig = function(k) {
```
weilu: It's not ideal that this function is repeated here with minor modification. I'm not if this is an abstraction problem as the function below used in `sign` has the side effect of modifying `s` and `r`. Maybe it's not worth the effort to refactor this as it will be removed in 2.0.0.

src/ecdsa.js
```diff
@@ -9,9 +9,36 @@ var ZERO = new Buffer([0])
 var ONE = new Buffer([1])
 
 // https://tools.ietf.org/html/rfc6979#section-3.2
-function deterministicGenerateK(curve, hash, d) {
+function deterministicGenerateK(curve, hash, d, checkSig) {
   typeForce('Buffer', hash)
   typeForce('BigInteger', d)
+//  typeForce('Function', checkSig)
```
weilu: Move this to the comment 2 lines done so peeps won't wonder why this is commented out

src/ecdsa.js
dcousens: Definitely not worth any refactoring effort.
I assume you have no other issues with this PR?

PR #343, opened 2015-01-27
integration: avoid excessive blockchain requests

dcousens:
This should fix the related integration test occasionally failing due to a timeout.
Currently 900+ transactions are returned unnecessarily.


dcousens:
@weilu post ACK if necessary

weilu:
nice.

dcousens:
Technically this is less malleability resistant, but I don't think that is
a major concern compared for this test case.

weilu:
Malleability isn't what the tests here care about, so I think it's totally fine.

PR #349, opened 2015-02-03
TransactionBuilder P2SH Multisig order of signatures (alternative implementation)

rubensayshi:
Alternative (IMO better) implementation of https://github.com/bitcoinjs/bitcoinjs-lib/pull/348

### DIFF
--- a/src/transaction_builder.js
+++ b/src/transaction_builder.js
@@ -178,7 +178,7 @@ TransactionBuilder.prototype.__build = function(allowIncomplete) {
     var scriptSig
     var scriptType = input.scriptType
 
-    var signatures = input.signatures.map(function(signature) {
+    var signatures = input.signatures.filter(function(signature) { return !!signature }).map(function(signature) {
       return signature.toScriptSignature(input.hashType)
     })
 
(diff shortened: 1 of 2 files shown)


src/transaction_builder.js
```diff
@@ -268,8 +275,25 @@ TransactionBuilder.prototype.sign = function(index, privKey, redeemScript, hashT
...
+    input.signatures.forEach(function(signature) {
+        pubKeys.forEach(function(pubKey, i) {
+            if (!multisigSignatures[i] && pubKey.verify(hash, signature)) {
+                multisigSignatures[i] = signature
+              }
+          })
+      })
+    input.signatures = pubKeys.map(function(pubKey, i) { return multisigSignatures[i] || null })
+    var signatureIndex = pubKeys.map(function(pubKey) { return pubKey.toBuffer().toString('hex') }).indexOf(privKey.pub.toBuffer().toString('hex'))
+    assert.notEqual(signatureIndex, -1)
```
dcousens: See https://github.com/bitcoinjs/bitcoinjs-lib/pull/329/files#diff-f18685c2b076ef917fa091a6f60a0dd4R345

dcousens: There is a tonne of fun to deal with `pubKey` inputs properly as well.

PR #352, opened 2015-02-08
Updated address.js javascript styling - now uses semicolons

benignbemine:
I did not modify any logic, all I have done is update the javascript statements to use semicolons. Note, semicolons after function declarations like on lines 14 and 24 are not necessary, so I did not add them there. It seems as though this project places a lot of importance on code style, so I am just hoping to help out! Let me know if you agree with this change, and I will update all of your JS files. 

Also, semicolons are super important, see javascript example below.

>   var x = 3 + f
>  (a+b).toString()

By missing the semicolons, the interpreter may confuse f to be a function: 

> var x = 3 + f(a+b).toString()

The npm test commands you suggested were not working on my computer, I received this message: 

"Failed at the bitcoinjs-lib@1.4.4 coverage script. This is most likely a problem with the bitcoinjs-lib package, not with npm itself. Tell the author that this fails on your system: istanbul cover _mocha -- test/*.js"

### DIFF
--- a/src/address.js
+++ b/src/address.js
@@ -1,64 +1,64 @@
-var assert = require('assert')
-var base58check = require('bs58check')
-var typeForce = require('typeforce')
-var networks = require('./networks')
-var scripts = require('./scripts')
+var assert = require('assert');
+var base58check = require('bs58check');
+var typeForce = require('typeforce');
+var networks = require('./networks');
+var scripts = require('./scripts');
 
 function findScriptTypeByVersion(version) {
   for (var networkName in networks) {
-    var network = networks[networkName]
+    var network = networks[networkName];
 
-    if (version === network.pubKeyHash) return 'pubkeyhash'
-    if (version === network.scriptHash) return 'scripthash'
+    if (version === network.pubKeyHash) return 'pubkeyhash';
+    if (version === network.scriptHash) return 'scripthash';
   }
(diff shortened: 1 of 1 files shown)


dcousens:
> I understand that there are some conventions this project is deliberately and intentionally not following (semicolons, by the way, are not one of them, there is no convention on the use of semicolons and a lot of projects don't use them, including some very big ones you're probably using right now). But there are almost always reasons we do some things differently than others. We don't bundle pre-compiled javascript files because it's dangerous and we can't guarantee their safety. We don't use semicolons because they are a cargo cult that are not required for statement end termination with JavaScript in almost all cases (if developers are working on JavaScript Bitcoin software, they should understand how JavaScript syntax works). And due to the low level of this library, it happens that integration tests are working better than the now-broken examples I wrote when we launched 1.0, because we can constantly monitor them to make sure they continue to work.

Also see https://github.com/bitcoinjs/bitcoinjs-lib/pull/96#issuecomment-38752164, https://github.com/bitcoinjs/bitcoinjs-lib/pull/326#issuecomment-65176637

dcousens:
On a side note, I was thinking of adopting https://github.com/feross/standard this @weilu. Thoughts?

weilu:
@dcousens seems reasonable :)

PR #354, opened 2015-02-11
scripts: add allowIncomplete for multisig scripts

dcousens:
Thanks to @bpdavenport for the initial implementation of this, however there was some issues with the existing implementation in https://github.com/bitcoinjs/bitcoinjs-lib/commit/f7604b81035839153725815dbf19a54e48c5ca94 that meant all multisig signatures were considered complete even though they weren't. 

To allow for this without breaking backwards compatibility, I have added the `allowIncomplete` flag and appropriate tests.

@bpdavenport does this look good to you?

This should bode well with the `txbmerge` changes and ideally be merged in the next 24 hours.

### DIFF
--- a/test/fixtures/scripts.json
+++ b/test/fixtures/scripts.json
@@ -53,6 +53,51 @@
       "type": "nulldata",
       "data": "deadffffffffffffffffffffffffffffffffbeef",
       "scriptPubKey": "OP_RETURN deadffffffffffffffffffffffffffffffffbeef"
+    },
+    {
+      "type": "nonstandard",
+      "typeIncomplete": "multisig",
+      "pubKeys": [
+        "0279be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798",
+        "02b80011a883a0fd621ad46dfc405df1e74bf075cbaf700fd4aebef6e96f848340",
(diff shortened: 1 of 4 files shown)


test/fixtures/scripts.json
```diff
@@ -53,6 +53,40 @@
...
+      "type": "multisig",
+      "typeIncomplete": "nonstandard",
+      "allowIncomplete": true,
+      "pubKeys": [
+        "0279be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798",
+        "02b80011a883a0fd621ad46dfc405df1e74bf075cbaf700fd4aebef6e96f848340",
+        "024289801366bcee6172b771cf5a7f13aaecd237a0b9a1ff9d769cabc2e6b70a34"
+      ],
+      "signatures": [
+        null,
```
dcousens: `null` is re-interpreted as `OP_0` as seen on line 148.

test/fixtures/scripts.json
```diff
@@ -53,6 +53,36 @@
       "type": "nulldata",
       "data": "deadffffffffffffffffffffffffffffffffbeef",
       "scriptPubKey": "OP_RETURN deadffffffffffffffffffffffffffffffffbeef"
+    },
+    {
+      "type": "nonstandard",
+      "typeIncomplete": "multisig",
```
dcousens: this type is used when the `allowIncomplete` flag is true

PR #355, opened 2015-02-11
Transaction: get normalized hash / id

mhanne:
Function to get the normalized tx hash / id.

This is ~~needed~~ convenient to deal with transaction malleability and is useful to keep track of partially signed multisig transactions where the regular hash changes after each signature is added.

### DIFF
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -185,6 +185,9 @@ Transaction.prototype.clone = function () {
  * method copies the transaction, makes the necessary changes based on the
  * hashType, serializes and finally hashes the result. This hash can then be
  * used to sign the transaction input in question.
+ * When a negative inIndex and no prevOutScript are given, it computes the hash
+ * with all transaction inputs set to zero. This can be used to identify a
+ * transaction independently of its signing status.
  */
 Transaction.prototype.hashForSignature = function(inIndex, prevOutScript, hashType) {
   // FIXME: remove in 2.x.y
@@ -198,20 +201,27 @@ Transaction.prototype.hashForSignature = function(inIndex, prevOutScript, hashTy
   }
 
   typeForce('Number', inIndex)
-  typeForce('Script', prevOutScript)
+  if (prevOutScript) {
+    typeForce('Script', prevOutScript)
+  }
   typeForce('Number', hashType)
 
-  assert(inIndex >= 0, 'Invalid vin index')
+  if (prevOutScript) {
+    assert(inIndex >= 0, 'Invalid vin index')
+  }
   assert(inIndex < this.ins.length, 'Invalid vin index')
 
   var txTmp = this.clone()
-  var hashScript = prevOutScript.without(opcodes.OP_CODESEPARATOR)
 
-  // Blank out other inputs' signatures
+  // Blank out all inputs' signatures
   txTmp.ins.forEach(function(txIn) {
     txIn.script = Script.EMPTY
   })
-  txTmp.ins[inIndex].script = hashScript
+
+  // Set stripped prevOutScript for the given input, unless we want an "empty" hash
+  if (prevOutScript && inIndex >= 0) {
+    txTmp.ins[inIndex].script = prevOutScript.without(opcodes.OP_CODESEPARATOR)
+  }
 
   var hashTypeModifier = hashType & 0x1f
   if (hashTypeModifier === Transaction.SIGHASH_NONE) {
@@ -240,6 +250,14 @@ Transaction.prototype.getHash = function () {
 Transaction.prototype.getId = function () {
   // TxHash is little-endian, we need big-endian
   return bufferutils.reverse(this.getHash()).toString('hex')
+}
+
+Transaction.prototype.getNormalizedHash = function () {
+  return this.hashForSignature(-1, null, Transaction.SIGHASH_ALL)
+}
+
+Transaction.prototype.getNormalizedId = function () {
+  return bufferutils.reverse(this.getNormalizedHash()).toString('hex');
 }
 
 Transaction.prototype.toBuffer = function () {
--- a/test/fixtures/transaction.json
+++ b/test/fixtures/transaction.json
@@ -4,6 +4,8 @@
       "description": "Standard transaction (1:1)",
       "id": "a0ff943d3f644d8832b1fa74be4d0ad2577615dc28a7ef74ff8c271b603a082a",
       "hash": "2a083a601b278cff74efa728dc157657d20a4dbe74fab132884d643f3d94ffa0",
+      "normalizedId": "321caa7f62477e0c2a8efe97374eb7e2d963f4aa9cdf471821086d788bcc60b3",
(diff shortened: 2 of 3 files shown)


src/transaction.js
```diff
@@ -198,20 +198,23 @@ Transaction.prototype.hashForSignature = function(inIndex, prevOutScript, hashTy
...
-  var hashScript = prevOutScript.without(opcodes.OP_CODESEPARATOR)
+  var hashScript = prevOutScript ? prevOutScript.without(opcodes.OP_CODESEPARATOR) : null
 
   // Blank out other inputs' signatures
   txTmp.ins.forEach(function(txIn) {
     txIn.script = Script.EMPTY
   })
-  txTmp.ins[inIndex].script = hashScript
+  if(hashScript != null && inIndex >= 0)
+    txTmp.ins[inIndex].script = hashScript
```
dcousens: Nit: please adhere to project style by adding whitespace after the `if` keyword, and always use strict equality where possible.

``` javascript
if (hashScript !== null && inIndex >= 0) {
  txTmp.ins[inIndex].script = hashScript
}
```

src/transaction.js
```diff
@@ -242,6 +245,14 @@ Transaction.prototype.getId = function () {
   return bufferutils.reverse(this.getHash()).toString('hex')
 }
 
+Transaction.prototype.getNormalizedHash = function () {
+  return this.hashForSignature(-1, null, Transaction.SIGHASH_ALL)
+}
+
+Transaction.prototype.getNormalizedId = function () {
+  return this.getNormalizedHash().toString('hex');
```
dcousens: This will need to be reversed

test/fixtures/transaction.json
```diff
@@ -27,6 +29,8 @@
       "description": "Standard transaction (2:2)",
       "id": "fcdd6d89c43e76dcff94285d9b6e31d5c60cb5e397a76ebc4920befad30907bc",
       "hash": "bc0709d3fabe2049bc6ea797e3b50cc6d5316e9b5d2894ffdc763ec4896dddfc",
+      "normalizedId": "c2745efcb4cef1a9005a54dedcf4a42bc417ea9a9413f488d0ca528b84937af7",
```
dcousens: This should be something like `f77a93 ...`

dcousens:
Everything looks great after you fix the above comments :).

The only remaining question is,  what problems does this solve?
What this transaction is doing is exactly as the implementation looks like:
``` javascript
return this.hashForSignature(-1, null, Transaction.SIGHASH_ALL)
```

It returns a hash which accounts for all inputs disregarding their script signatures.
This allows you to identify transactions that will potentially have different script signatures (malleability), but it also stops you from recognizing transactions that may have varied inputs.

I feel like there is a hidden gotcha in what might allow unknowing users to use this without knowing its consequence... however, granted, they might use the standard Id/Hash also, so I'm not sure if I completely agree with that.

The Bitcoin core team eventually [voted against merging it](https://github.com/bitcoin/bitcoin/pull/3656),  but it should be noted large companies like Coinbase mentioned that they did use a normalized hash internally.  

[bitcoinj](https://github.com/bitcoinj/bitcoinj) have no such thing.

[The Bitcore team](https://github.com/bitpay/bitcore/pull/249) merged it.

@ryanxcharles, thoughts on whether you felt normalized hash has had a positive net gain for Bitcore? (or even https://github.com/ryanxcharles/fullnode)

dcousens:
Also pinging @jprichardson, @locksley, @weilu

dcousens:
@mhanne I'm almost tempted to not add the methods, but allow for the parameter support of:

``` javascript
return this.hashForSignature(-1, null, Transaction.SIGHASH_ALL)
```

Which you have only just added.

mhanne:
Alright, I fixed the issues you mentioned, thanks for pointing it out :)

I think there is a valid use for an identifier of a specific set of changes to the current state of the blockchain.

If you look at it like that, it's not *supposed* to be the same hash for different stages of ANYONECANPAY, nor should it care about how a transaction is signed (ie. who signed it):

- If the actual inputs (not just the signatures) change, it would result in a different state change in the blockchain

- Likewise, it doesn't matter who signed it or if it even has a valid signature at all, it represents the same (potential) change to the utxo set.

The "sighash with all inputs zeroed out" kind of suggests itself, though I agree that "normalized hash/id" probably isn't the best name. Maybe "empty sighash" or "null sighash" or something like that? "normalized sighash"? That would make it clearer and hint at what it actually is.


Unless I'm mistaken, @gmaxwell is only arguing why it wouldn't be a workable solution to *exclusively* use the normalized hash for tracking, not how it could hurt to have it additionally.

It's not meant to replace the regular tx hash, nor is it a simple drop-in replacement to magically fix all code that is vulnerable to malleability. It's also not *required* to deal with it, but it can be handy. I don't see how it not being applicable to every situation would be a reason not to use it when it is.

mhanne:
To clarify, the specific usecases I'm thinking of are:

1) when sending (or receiving) an unconfirmed transaction, use it to check if the tx - or its equivalent - is being included in a block

2) when passing a multisig tx around between signers, use it to identify the transaction while the regular hash changes

Is there anything fundamentally wrong with that?

I'm not actually using bitcoinjs to send/receive transactions, so to me the first point is more theoretical, but I'm working on something involving case two right now.

Personally, I'd be fine with calling hashForSignature() directly, but having a shortcut/label for it would be nice.

dcousens:
ACK on all changes, good to merge.  Just a matter of whether I agree with the fundamentals now :).  Still thinking it over and waiting for others feedback if possible.  Thanks @mhanne

src/transaction.js
```diff
@@ -242,6 +252,14 @@ Transaction.prototype.getId = function () {
   return bufferutils.reverse(this.getHash()).toString('hex')
 }
 
+Transaction.prototype.getNormalizedHash = function () {
```
weilu: I honestly think this function can be achieved much cleaner without using and modifying `hashForSignature`:

```js
function getNormalizedHash = function(tx) {
  var txTmp = tx.clone()

  txTmp.ins.forEach(function(txIn) {
    txIn.script = Script.EMPTY
  })

  var hashTypeBuffer = new Buffer(4)
  hashTypeBuffer.writeInt32LE(Transaction.SIGHASH_ALL, 0)

  return crypto.hash256(Buffer.concat([txTmp.toBuffer(), hashTypeBuffer]))
}
```

@dcousens put up a very nice summary of the whole debate. IMHO, aligning bitcoinjs' features with bitcoin core as much as possible provides consistency and least surprise for devs. I'm not comfortable merging a feature explicitly rejected by the core team after much debate, especially when this feature is very easy to implement when one indeed needs it for their own purposes.

jprichardson:
@weilu said it best:

>  I'm not comfortable merging a feature explicitly rejected by the core team after much debate, especially when this feature is very easy to implement when one indeed needs it for their own purposes.

:+1:

mhanne:
Alright then, thanks for taking the time to consider it.

PR #356, opened 2015-02-12
ECSignature: enforce valid hashType in toScriptSignature

dcousens:
This PR just adds an assert to enforce valid `hashType`s are used in `ECSignature.prototype.toScriptSignature`, along with tests.


weilu:
ack

PR #357, opened 2015-02-12
doc: no section headers

dcousens:
This was agreed to a while back, just making this consistent across the repository.


weilu:
ack

PR #358, opened 2015-02-12
package: remove uglify-js

dcousens:
It is not our responsibility to uglify the code.

If they want to uglify it, there are plenty of tools to which they could use, just like the rest of their build process.

@weilu thoughts?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,3 @@
-node_modules
-bitcoinjs-min.js
+bitcoin.js
 coverage
+node_modules
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     }
   ],
   "scripts": {
-    "compile": "browserify ./src/index.js -s bitcoin | uglifyjs > bitcoinjs-min.js",
+    "compile": "browserify ./src/index.js -s bitcoin > bitcoin.js",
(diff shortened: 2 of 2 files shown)


dcousens:
@weilu still happy to hear feedback, but I can't foresee any real issue with this.

weilu:
Here's a reason why this change is a bad idea:

```
±  |master ✗| → ll -h bitcoin.js 
-rw-r--r--@ 1 wlu  401060983   620K Feb 16 19:47 bitcoin.js

±  |master ✗| → ll -h bitcoinjs-min.js 
-rw-r--r--@ 1 wlu  401060983   149K Oct  7 15:34 bitcoinjs-min.js
```

weilu:
After I posted it I realized that the minified bitcoinjs was of a much older version. Current master yields:

```
±  |master ✗| → ll -h bitcoinjs-min.js 
-rw-r--r--@ 1 wlu  401060983   424K Feb 16 19:52 bitcoinjs-min.js
```

@dcousens I'm curious to what makes bitcoinjs-lib 3x more bloated than 4 months ago

dcousens:
Crypto-browserify has grown substantially in size.  That may accommodate
for some of it.

Issue #362, opened 2015-02-17
throw new RangeError('Trying to access beyond buffer length')


p2pmining:
When creating a transaction from coinbase transaction of block `0000000000000000186bf9453bd37528e7caf0cec4ce46d9fa5d3ce7bead507d` , Range error is thrown when bitcoinjs-lib creates the input script.
Code
```javascript
var hex = "01000000010000000000000000000000000000000000000000000000000000000000000000ffffffff29032832051c4d696e656420627920416e74506f6f6c20626a343a45ef0454c5de8d5e5300004e2c0000ffffffff01414f1995000000001976a914b05793fe86a9f51a5f5ae3a6f07fd31932128a3f88ac00000000"
var tx = new bitcoin.Transaction.fromHex(hex)
```

Error
`
RangeError: Trying to access beyond buffer length
    at checkOffset (buffer.js:582:11)
    at Buffer.readUInt32LE (buffer.js:609:5)
    at Object.readPushDataInt (/home/jtserafin_gmail_com/wwBackEnd/bitcoinjs-lib/src/bufferutils.js:42:21)
    at Function.Script.fromBuffer (/home/jtserafin_gmail_com/wwBackEnd/bitcoinjs-lib/src/script.js:39:27)
    at readScript (/home/jtserafin_gmail_com/wwBackEnd/parse.js:126:27)
    at Object.<anonymous> (/home/jtserafin_gmail_com/wwBackEnd/parse.js:93:13)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
`

weilu:
Same issue as #344?

dcousens:
Apparently not.

socr102:
same issue

socr102:
RangeError: Trying to access beyond buffer length
    at checkOffset (index.js:1127:1)
    at Uint8Array.readUIntLE (index.js:1134:1)
    at UInt.decode (Layout.js:570:1)
    at Structure.decode (Layout.js:1234:1)
    at WrappedLayout.decode (index.ts:102:1)
    at WrappedLayout.decode (index.ts:102:1)
    at Structure.decode (Layout.js:1234:1)
    at Sequence.decode (Layout.js:1090:1)
    at Structure.decode (Layout.js:1234:1)
    at WrappedLayout.decode (index.ts:102:1)

PR #363, opened 2015-02-19
Coinbase Transaction parsing

dcousens:
Fixes https://github.com/bitcoinjs/bitcoinjs-lib/issues/362

Naturally, this bug would also be present in the `Transaction` parsing code in `block.js`, so for now I've just put in f136505 to allow us to bypass that.

As discussed with @weilu (somewhere?), I'm thinking for `2.0.0`, that `Block` will be headers only anyway.


dcousens:
Despite TRAVIS complaining about coveralls, all tests are passing.

PR #366, opened 2015-02-23
Allow constructing HDNode from an ECPubKey.

Sjors:
Currently you can only construct an HD node using either a private key or an ECPoint. When using an ECPoint it is turned into an ECPubKey. This pull request makes it possible to use an ECPubKey directly.


dcousens:
This is already covered in `2.0.0`, however, it is useful and for those pre `2.0.0`, it'll benefit.
I'll add it in as `1.4.6`.

dcousens:
Will be adding tests and an assertion that the public key is compressed, post-merge.

PR #368, opened 2015-03-02
add `fixSignatureOrder` option to txb.sign

rubensayshi:
add `fixSignatureOrder` option to `txb.sign` for when trying to sign transactions that don't have `OP_0` in place of missing signatures.

It's not doing this by default atm, but I think it would be a good idea to do this by default in `txb.sign` since asuming that partial transactions will always have `OP_0` to fill missing signatures is not very safe 
but if you guys prefer not doing this by default then at least having the option to do it in bitcoinjs-lib wouldn't hurt imo

this has also broken backward compatibily completely with being able to sign a partially signed transaction from a v1.4.4 bitcoinjs-lib with a v1.4.5+ bitcoinjs-lib.

### DIFF
--- a/src/transaction_builder.js
+++ b/src/transaction_builder.js
@@ -271,7 +271,7 @@ TransactionBuilder.prototype.__build = function (allowIncomplete) {
   return tx
 }
 
-TransactionBuilder.prototype.sign = function (index, privKey, redeemScript, hashType) {
+TransactionBuilder.prototype.sign = function (index, privKey, redeemScript, hashType, fixSignatureOrder) {
   assert(index in this.inputs, 'No input at index: ' + index)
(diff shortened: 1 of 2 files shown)


rubensayshi:
hmm, I'm also seeing issues with signing 2of3 multisigs, if the middle signature is left as an `OP_0` because the 1st and 3rd key are used to sign then there's this `OP_0` in the middle messing everything up 

`bitcoin-cli sendrawtransaction` gives "16: mandatory-script-verify-flag-failed (Dummy CHECKMULTISIG argument must be zero)" when this happens

rubensayshi:
closing this one since I found out the 2of3 issue, alternative; https://github.com/bitcoinjs/bitcoinjs-lib/pull/369

PR #369, opened 2015-03-02
stop filling P2SH multisig with OP_0 when there's no signature.

rubensayshi:
better alternative to https://github.com/bitcoinjs/bitcoinjs-lib/pull/368
 - only fill P2SH multisig with `OP_0` in `txb.Incomplete`. 
 - fix the order of the signature in `txb.sign` if there's not enough `OP_0`s.

we can't leave the `OP_0`s in the transaction when trying to broadcast it, `bitcoin-cli sendrawtransaction` gives `6: mandatory-script-verify-flag-failed (Dummy CHECKMULTISIG argument must be zero)` when this happens.

no tests were failing because there are only 2of2 tests, they never have this issue and most likely noone noticed in production because if it's the 3rd key that is missing a signature it's also trimmed off (by coincidence)

### DIFF
--- a/src/transaction_builder.js
+++ b/src/transaction_builder.js
@@ -238,8 +238,13 @@ TransactionBuilder.prototype.__build = function (allowIncomplete) {
         case 'multisig': {
           // Array.prototype.map is sparse-compatible
           var msSignatures = input.signatures.map(function (signature) {
-            return signature.toScriptSignature(input.hashType)
-          }).filter(function(signature) { return !!signature })
+            return signature && signature.toScriptSignature(input.hashType) || ops.OP_0
+          })
+
+          // filter out OP_0s when we're not building an incomplete TX
+          if (!allowIncomplete) {
+            msSignatures = msSignatures.filter(function (signature) { return !!signature })
+          }
 
           var redeemScript = allowIncomplete ? undefined : input.redeemScript
           scriptSig = scripts.multisigInput(msSignatures, redeemScript)
@@ -268,7 +273,7 @@ TransactionBuilder.prototype.__build = function (allowIncomplete) {
   return tx
 }
 
-TransactionBuilder.prototype.sign = function (index, privKey, redeemScript, hashType) {
+TransactionBuilder.prototype.sign = function (index, privKey, redeemScript, hashType, forceFixSignatures) {
   assert(index in this.inputs, 'No input at index: ' + index)
   hashType = hashType || Transaction.SIGHASH_ALL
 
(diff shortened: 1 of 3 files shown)


ryanxcharles:
I'm not completely sure how bitcoinjs-lib handles multisig, but it is better practice to fill in missing signatures with an OP_0, because this let's the signers put their signature in the correct place without having to validate each other signature in order to determine the correct order (since multisig signatures must be in the same order as the public keys).

rubensayshi:
maybe it should only be done with doing `txb.buildIncomplete` because bitcoind doesn't like it when we keep a `OP_0` around

ryanxcharles:
Again, I'm not exactly sure how bitcoinjs-lib handles this, but any superfluous OP_0s need to be removed before broadcasting the transaction. e.g., a 3-of-5 multisig tx should start with 5 OP_0s, then once 3 signatures are added, the last two OP_0s should be dropped, then the tx can be broadcast. This strategy for filling in signatures minimizes the amount of computation each signer has to do.

rubensayshi:
but you can't asume that every library does that, so it greatly depends on weither or not your use case involves having control over where the partially signed TXs come from or not.

ryanxcharles:
Agreed. If insufficient OP_0s are present, you have to determine the order of the signatures somehow. However, putting the OP_0s in there really should be the standard, and everyone should do it that way.

rubensayshi:
okay, but the biggest problem right now is that the `OP_0`s shouldn't be there when you want to broadcast, and they are.

ryanxcharles:
Yep ... those need to be removed. There should be a "finalize signatures" step that removes the superfluous OP_0s before broadcasting. (Again ... I'm not very familiar with bitcoinjs-lib, so I don't know how it does or does not handle that at present.)

rubensayshi:
PS. are you guys aware that when doing `mocha` without any arguments it will only run tests in `test/*.js` and not `test/integration/*.js` ?

dcousens:
> PS. are you guys aware that when doing mocha without any arguments it will only run tests in test/*.js and not test/integration/*.js ?

`npm run integration`

vs

`npm run unit`

For fidelity over both.
Perhaps we should make `npm test` run both, and not just `unit` by default. 

But yes, we are aware.

dcousens:
@rubensayshi could you rebase this PR, then re-open it and apply just the fixes for:

> fix the order of the signature in txb.sign if there's not enough OP_0s.

Also, try and use the transaction fixtures if possible.

Thanks for your work on this :)

src/transaction_builder.js
```diff
@@ -238,14 +238,12 @@ TransactionBuilder.prototype.__build = function (allowIncomplete) {
...
 
-          // fill in blanks with OP_0
-          for (var i = 0; i < msSignatures.length; ++i) {
-            if (msSignatures[i]) continue
-
-            msSignatures[i] = ops.OP_0
+          // filter out OP_0s when we're not building an incomplete TX
+          if (!allowIncomplete) {
+            msSignatures = msSignatures.filter(function (signature) { return !!signature })
           }
```
dcousens: Identical to https://github.com/bitcoinjs/bitcoinjs-lib/commit/bcf8d0177dfb9b7c859b010aec695293372bf764, except isolated.

dcousens:
For what its worth, `fromTransaction` should be the only place that handles the possibility of `OP_0` not being present, but not enough signatures existing.

It should re-order the signatures (if possible) and create the sparse signatures array.

rubensayshi:
@dcousens I disagree, that would move the performance heavy parts into `fromTransaction`, while I'd greatly prefer to have those in `sign` instead because I wouldn't expect `fromTransaction` to do much other than initialize

PR #370, opened 2015-03-02
Change 2-of-2 integration test to 2-of-3

rubensayshi:
add failing unittest to prove the error I'm trying to fix in https://github.com/bitcoinjs/bitcoinjs-lib/pull/369

### DIFF
--- a/test/integration/multisig.js
+++ b/test/integration/multisig.js
@@ -19,18 +19,19 @@ describe('bitcoinjs-lib (multisig)', function () {
     assert.equal(address, '36NUkt6FWUi3LAWBqWRdDmdTWbt91Yvfu7')
   })
 
-  it('can spend from a 2-of-2 multsig P2SH address', function (done) {
+  it('can spend from a 2-of-3 multsig P2SH address', function (done) {
     this.timeout(20000)
 
     var privKeys = [
       '91avARGdfge8E4tZfYLoxeJ5sGBdNJQH4kvjJoQFacbgwmaKkrx',
-      '91avARGdfge8E4tZfYLoxeJ5sGBdNJQH4kvjJoQFacbgww7vXtT'
+      '91avARGdfge8E4tZfYLoxeJ5sGBdNJQH4kvjJoQFacbgww7vXtT',
+      '91avARGdfge8E4tZfYLoxeJ5sGBdNJQH4kvjJoQFacbgx3cTMqe'
     ].map(bitcoin.ECKey.fromWIF)
     var pubKeys = privKeys.map(function (x) {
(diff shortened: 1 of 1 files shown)


test/integration/multisig.js
```diff
@@ -77,4 +77,63 @@ describe('bitcoinjs-lib (multisig)', function () {
...
+        var txb = new bitcoin.TransactionBuilder()
+        txb.addInput(unspent.txId, unspent.vout)
+        txb.addOutput(targetAddress, 1e4)
+
+        // sign with 1st and 3rd key
+        txb.sign(0, privKeys[0], redeemScript)
+        txb.sign(0, privKeys[2], redeemScript)
+
+        // broadcast our transaction
+        blockchain.transactions.propagate(txb.build().toHex(), function (err) {
```
dcousens: This should fail.

dcousens:
I can't add this test as is, because technically the test is taking advantage of code that should be failing elsewhere.

We do however need to add a test to ensure that `txb.build().toHex()` fails when a signature is missing.  Infact, I'm not sure why it isn't already.

dcousens:
Disregard previous comments, accidentally read this as `3 of 3`, not `2 of 3`.

dcousens:
@rubensayshi I have added https://github.com/bitcoinjs/bitcoinjs-lib/commit/c79fecffa3ded9ee26db8c777b72a58d09ed8a48 to show case this test without the need for it to be an integration test.

What is shown there is the expected `txHex`,  it verifies that no OP_0 should be present in the scriptSig; and as such, before https://github.com/bitcoinjs/bitcoinjs-lib/commit/707376c81eefea312392c0dc9f6868e9aff3eb3d was a failing test. 

Please ACK this and I'll close this PR. :)

rubensayshi:
hmm, I changed the 2of2 integration test to 2of3, I think that's a better test case since it has a greater chance of failing (for various reasons)

dcousens:
@rubensayshi I'm OK with this change, but I don't think it covers any less or more than `2-of-2`, it simply covers the other case (fully signed vs partially signed). 

However this does show case a better example of multisig to the typical user, so I'll update the README example after to merge.

dcousens:
@rubensayshi to follow that line of thought, [I changed it](https://github.com/bitcoinjs/bitcoinjs-lib/commit/29b93b8a5c393a21865a359d2491faec5664d2e8) to a less traditional `2-of-4` which would better illustrate some parts of the API and provide a larger contrast to the example above it, the `2-of-2`.

rubensayshi:
awesome!

PR #372, opened 2015-03-02
TransactionBuilder multisig OP_0 fix

dcousens:
Replaces #370, and introduces a fix.

PR #373, opened 2015-03-03
TransactionBuilder P2SH Multisig order of signatures

rubensayshi:
this PR is based off https://github.com/bitcoinjs/bitcoinjs-lib/tree/op0fix (https://github.com/bitcoinjs/bitcoinjs-lib/pull/372) and I also merged in https://github.com/bitcoinjs/bitcoinjs-lib/pull/370 because both of those are needed for me to continue upon in this PR.

 - fix the order of the signature in txb.sign if there's not enough `OP_0`s (or `forceFixMultisigSigOrder` is specified when the user knows they're getting a messed up TX).
 - changed tests to use fixtures.

Like I said in https://github.com/bitcoinjs/bitcoinjs-lib/pull/369 when doing `txb = new TransactionBuilder(); txb.addInput(); txb.sign()` the `sign` is the first time we actually learn about the redeemScript so it's not possible to do this stuff in `fromTransaction` but it has to go into `sign`.


rubensayshi:
hey @dcousens if you use the 'Files Changed' tab of the PR and do comments there instead of going commit for commit than it's a bit easier imo to reply to the stuff than going to the commits.

also because you went commit for commit, you basicly commented a lot of a commit I actually took the code out for in a later commit ;-)

dcousens:
@rubensayshi no worries, not sure why I commented on the commits.  My bad.

dcousens:
Will review tonight.

dcousens:
Sorry, was a bit delayed on reviewing this.  If you could please rebase on master :).  Will begin review anyway.

rubensayshi:
rebased

src/transaction_builder.js
```diff
@@ -238,7 +238,7 @@ TransactionBuilder.prototype.__build = function (allowIncomplete) {
         case 'multisig': {
           // Array.prototype.map is sparse-compatible
           var msSignatures = input.signatures.map(function (signature) {
-            return signature.toScriptSignature(input.hashType)
+            return signature && signature.toScriptSignature(input.hashType)
```
dcousens: My only qualm was that I don't think this is necessary.  However I know it is necessary given the current solution of pre-emptively always putting in the `OP_0`'s.  I might change that and submit a PR.

dcousens:
Awesome work @rubensayshi .   Thanks so much for this.

Sorry it took me a while to get around to it.

PR #377, opened 2015-03-06
Be able to get an address from a pay to publickey script.

p2pmining:
I was parsing the blockchain with bitcoinjs-lib, I found this change helped when I was indexing output script addresses.  It is common for a block explorer to show the public key as an address.


dcousens:
I don't think we should construe the meaning behind these scripts any more than necessary. 

A `pay-to-pubkey` script does not have an address, therefore, we should not confuse developers into thinking it does.

An address is an encoded script pub key such that it is human readable, it has a clear 1:1 mapping such that an address can be decoded to exactly the data that made it up.

Assuming we ignore that `pay-to-pubkey` is not intended to be addressable, and add the code as above, suddenly we will have a situation where someone can do the following:

`pay-to-pubkey scriptPubKey -> address`, when they attempt to 'reverse' it, instead they get the following `address -> pay-to-pubkeyhash`. 
This makes no sense, and instead is throwing away information for the sake of misleading convenience.

fanatid:
Look like solved #273 :)
But I'm agree with Daniel, it make no sense to add `pay-to-pubkey` to `Address`.
If your need this, you could write own function in the project for transformation pubkey `Script` type to `Address` or multisig `Script` type to array of `Address`. At least I did so in [coloredcoinjs-lib](https://github.com/chromaway/coloredcoinjs-lib/blob/aafa36c69f8d7a2ebb4bc70386dfdd0e26dde1d4/src/bitcoin.js#L33).

dcousens:
Thanks for digging up that issue @fanatid.

Closing as duplicate of #273.   Happy to continue the discussion however :)

PR #378, opened 2015-03-10
#373 clean up

dcousens:
#373 introduced some changes to `TransactionBuilder` that solved a few multisig bugs.
Thanks to @rubensayshi :)

However, there a few things that just need to be cleaned up before publish :)

**edit:** WIP PR, ready for merge by tomorrow hopefully


dcousens:
@rubensayshi on further testing,  I was optimizing the `sign` from `O(n*m)` to make use of the guaranteed order of the public keys, and I've noticed you are stlil re-ordering the public keys if they are shuffled.

Was there a reason for this?  I thought we agreed to not include that since that would mean a client is incorrectly signing transactions.
It also means the current solution cannot handle the case (AFAIK) where a multi-signature address is made up of the same public key more than once. 
I'll have to add a test for that to find out :)

rubensayshi:
it actually does make use of the order of pubKeys and signatures a bit, just not for the pubKeys for which there's no signature yet.

whenever a signature is matched it's `splice`d and the next pubKey won't be checking it, so if we have
```
pubKey1  |  pubKey2  |  pubKey3   |  pubKey4
sig1  |  sig3  |  sig4
```

then 
 - `pubKey1` will match `sig1` and break,  
 - `pubKey2` will try to match `sig3` and `sig4` and fail
 - `pubKey3` will match `sig3` and break
 - `pubKey4` will match `sig4` and we're done

I guess you want `pubKey2` to break if it doesn't match `sig3` and not try `sig4`.

test/transaction_builder.js
```diff
@@ -250,7 +250,7 @@ describe('TransactionBuilder', function () {
           f.inputs.forEach(function (input, index) {
             var privKey = bitcoin.ECKey.fromWIF(input.signs[i].privKey)
             var redeemScript = bitcoin.Script.fromASM(input.redeemScript)
-            txb.sign(index, privKey, redeemScript, null, forceFixMultisigSigOrder)
```
rubensayshi: ah sorry for leaving that in

dcousens:
I'll have to read through it again.  Thanks for the explanation.

src/transaction_builder.js
```diff
@@ -372,22 +372,19 @@ TransactionBuilder.prototype.sign = function (index, privKey, redeemScript, hash
...
+    var signatures = []
+
+    input.pubKeys.forEach(function (pubKey, i) {
+      var allowed = input.signatures.slice(i)
 
-      return signature || ops.OP_0
+      allowed.some(function (signature) {
+        if (!pubKey.verify(signatureHash, signature)) return
+
+        signatures[i] = signature
```
rubensayshi: the tests are failing because with this change if a 2of3 is signed key1 and key2 (or a 2of2 with just key1) you won't have `OP_0`s for key3, imo you should stick to `input.signatures = input.pubKeys.map` to make sure that `input.signatures.length == input.pubKeys.length`.

dcousens: Stupid battery died,  charger at work. Will follow up tomorrow.

dcousens:
> I guess you want pubKey2 to break if it doesn't match sig3 and not try sig4.

IMHO, yes.  It overall means less verifies, and should never be a valid case unless someone is using buggy software.

dcousens:
@rubensayshi (ref https://github.com/bitcoinjs/bitcoinjs-lib/commit/3070a4857b711aea2166075e9cf5446d06c042ce), please ACK

If you feel these tests are beneficial, could you outline what they are testing, maybe we could add them to the scripts fixtures?

In the end, anything *really* odd would be picked up by the final `txHex` comparison anyway.

dcousens:
@rubensayshi I've made some changes to the testing code, could you verify?

The tests should be identical in terms of what is tested, however slightly more robust in that they also check the `txHexIncomplete` after each signature.

Other than that, all tests should pass and I've taken into account some of the points you made above.

dcousens:
@rubensayshi I cleaned up the tests a lot, and there were a few comments I made along the way that may not make sense anymore, please verify by checking this thread (not your email haha).

I have opted for https://github.com/bitcoinjs/bitcoinjs-lib/commit/69eb58c78374a25c41fd4986b474bedc5376747f instead of `txHexIncomplete`, it is more obvious what is happening and very easy to debug.

dcousens:
Perhaps for easy test transparency, instead of (or at least, in addition to) the final `txHex`, we could instead just iterate over each `input` and have a final ASM `scriptSig` with which to verify against?

dcousens:
@rubensayshi probably easiest to review commit-by-commit.

dcousens:
Tests are only failing because helloblock is experiencing technical difficulties.

rubensayshi:
@dcousens we have a different taste for unittesting, but that's fine and the changes look okay

PR #380, opened 2015-03-15
Magic constants

dcousens:
This pull request does two things.

* It adds the `magic` [protocol](https://en.bitcoin.it/wiki/Protocol_specification#Message_structure) constant to the `networks` object, renaming `magicPrefix` to `messagePrefix` (with `1.5.x` compatibility added for now)

* It changes `estimateFee` to use a new function `Transaction.prototype.byteLength` which is the size calculation functionality extracted out of `Transaction.prototype.toBuffer`, this saves unnecessary building of a `Buffer` in `estimateFee`.

All changes are backwards compatible (excepting 5b4e762d2aa3fa3412d8dc9e9e4d511445584242), and could be directly applied to `1.5.x` via cherry pick.

### DIFF
--- a/src/message.js
+++ b/src/message.js
@@ -11,11 +11,11 @@ var ecurve = require('ecurve')
 var ecparams = ecurve.getCurveByName('secp256k1')
 
 function magicHash (message, network) {
-  var magicPrefix = new Buffer(network.magicPrefix)
+  var messagePrefix = new Buffer(network.messagePrefix)
   var messageBuffer = new Buffer(message)
   var lengthBuffer = bufferutils.varIntBuffer(messageBuffer.length)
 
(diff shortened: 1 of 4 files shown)


dcousens:
@p2pmining and @rubensayshi ACK appreciated if you could :)

dcousens:
ping @jprichardson

p2pmining:
ACK

rubensayshi:
ACK

jprichardson:
ACK. But I'm still unsure why you do `=== undefined` instead of the `==` to leverage type coercian to catch `null` too...

dcousens:
@jprichardson what semantic meaning would a `null` have?  In any case, the function will hopefully be gone soon, so this will just be a bike shed :)

PR #384, opened 2015-03-17
crypto: use packages directly

dcousens:
This reduces the browserify build size from 600k to ~300k.   While most apps may include `crypto` anyway, they all fall back to the same modules as this is what `crypto` uses in browserify.

Not indifferent to using `require('inherits')` vs `require('events').inherit`


dcousens:
Will merge after TRAVIS passes

rubensayshi:
awesome!

PR #385, opened 2015-03-19
ECPair

dcousens:
This pull request (finally) integrates `ECPair` across the project and pulls it into master.

Aside from the discussions we've had where this is proposed as a much better, more coherent API for users, thanks to 3183229, the total test time is also almost 40% shorter.


test/fixtures/transaction_builder.json
```diff
@@ -723,11 +723,11 @@
             "vout": 1,
             "signs": [
               {
-                "privKey": "KwDiBf89QgGbjEhKnhXJuH7LrciVrZi3qYjgd9M7rFU73sVHnoWn",
+                "keyPair": "5HpHagT65TZzG1PH3CSu63k8DbpvD8s5ip4nEB3kEsreAnchuDf",
```
dcousens: These are the same private keys, just different compression flags (they were wrong previously)

jprichardson:
As far as I can tell, to summarize this PR, this pretty much renames `ECKey` to `ECPair` and lazy calculates `Q`, correct? Either way, I like this change a lot.

dcousens:
It merges `ECKey` and `ECPubKey`

dcousens:
Merging today.

PR #393, opened 2015-04-09
add a note about uglying (to avoid issues with typeForce)

rubensayshi:


dcousens:
I made a few changes, mainly around the wording.

Also, `BigInteger` was [accidentally](https://github.com/bitcoinjs/bitcoinjs-lib/commit/b9f9620207e0d7b7ccff1ff197171c5df6ad7cf9) spelled as `BitInteger`.

dcousens:
I used the following command to generate the current list of type names to avoid uglifying.
Probably excessive since some are built ins, but, alas.

``` bash
grep "typeforce(" src/* -hir | cut -c 13- | grep "'[A-Za-z]\+'" -o | sort -u | tr "'" "\`" | tr "\n" ",  "
```

PR #397, opened 2015-04-22
bufferutils: remove equal, use Buffer.compare

dcousens:


dcousens:
@jprichardson merge after ACK

src/transaction_builder.js
```diff
@@ -408,7 +408,7 @@ TransactionBuilder.prototype.sign = function (index, keyPair, redeemScript, hash
     input.signatures[i] = signature
 
     return true
-  }, this), 'key pair cannot sign for this input')
```
dcousens: `this` context here was superfluous, probably shouldn't be in this commit, sorry.  Will remove if necessary.

jprichardson:
ACK. `Buffer.compare` is not in Node v0.10 though. Do we want to give up compatibility for that version?

dcousens:
@jprichardson we have to keep moving forward as a community.
I'm OK with dropping 0.10 support. 

This doesn't affect browserify builds in any way, as they are already 0.12 compatible (in fact, they are following iojs for the most part).

dcousens:
Only tests that are failing are the integration tests, and they are failing because of https://github.com/bitcoinjs/bitcoinjs-lib/issues/399.

dcousens:
For further posterity, this PR was reverted in https://github.com/bitcoinjs/bitcoinjs-lib/commit/d7019e7492e3675d9d90158e42a90808f11ef7c3

PR #400, opened 2015-04-28
Add non SIGHASH_ALL flag support

dcousens:
Thanks to @dthorpe for the initial implementation of this.

I had to re-write it essentially from the core client, it is 1:1 now.
I am only testing it against the core tests [for now], with hope of pulling in some of @dthorpe's test fixtures.

### DIFF
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -177,6 +177,8 @@ Transaction.prototype.clone = function () {
   return newTx
 }
 
+var one = new Buffer('0000000000000000000000000000000000000000000000000000000000000001', 'hex')
+
 /**
  * Hash transaction for signing a specific input.
  *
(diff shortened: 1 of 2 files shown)


src/transaction.js
```diff
@@ -267,7 +307,12 @@ Transaction.prototype.toBuffer = function () {
 
   writeVarInt(this.outs.length)
   this.outs.forEach(function (txOut) {
-    writeUInt64(txOut.value)
+    if (!txOut.valueBuffer) {
+      writeUInt64(txOut.value)
+    } else {
+      writeSlice(txOut.valueBuffer)
+    }
```
dcousens: Admittedly, this does suck a little bit.  We almost got away with never needing `>2^53`, but I think this covers the use case sufficiently.

I don't like that it is no longer symmetrical with `fromBuffer`.
Open to ideas on a better solution here.

dcousens:
All tests passing.  Will merge on first ACK.

jprichardson:
ACK. I looked over the code and did not notice anything that stood out as an issue or concern. However, it may be good to get an extra set of eyes on it.

dcousens:
@weilu would you be able to check over it?

dcousens:
Merging by Friday.  @weilu if no time,  review post merge please :)

src/transaction.js
```diff
@@ -177,6 +177,8 @@ Transaction.prototype.clone = function () {
   return newTx
 }
 
+var one = new Buffer('0000000000000000000000000000000000000000000000000000000000000001', 'hex')
```
weilu: Really? How about

```
var ONE = new Buffer(64)
ONE.fill(0)
ONE[63] = 1
```

src/transaction.js
```diff
@@ -191,33 +193,71 @@ Transaction.prototype.hashForSignature = function (inIndex, prevOutScript, hashT
...
+    txTmp.outs = txTmp.outs.slice(0, nOut + 1)
 
-  if (hashTypeModifier === Transaction.SIGHASH_NONE) {
-    assert(false, 'SIGHASH_NONE not yet supported')
-  } else if (hashTypeModifier === Transaction.SIGHASH_SINGLE) {
-    assert(false, 'SIGHASH_SINGLE not yet supported')
+    // blank all other outputs (clear scriptPubKey, value === -1)
+    var stubOut = {
+      script: Script.EMPTY,
+      valueBuffer: new Buffer('ffffffffffffffff', 'hex')
```
weilu: ditto

src/transaction.js
```diff
@@ -191,33 +193,71 @@ Transaction.prototype.hashForSignature = function (inIndex, prevOutScript, hashT
...
+        input.sequence = 0
+      }
+    })
+
+  } else if ((hashType & 0x1f) === Transaction.SIGHASH_SINGLE) {
+    var nOut = inIndex
+
+    // only lock-in the txOut payee at same index as txIn
+    // https://github.com/bitcoin/bitcoin/blob/master/src/test/sighash_tests.cpp#L60
+    if (nOut >= this.outs.length) return one
```
weilu: log a warning before returning one here?

src/transaction.js
```diff
@@ -191,33 +193,71 @@ Transaction.prototype.hashForSignature = function (inIndex, prevOutScript, hashT
...
+      if (i !== inIndex) {
+        input.sequence = 0
+      }
+    })
+
+  } else if ((hashType & 0x1f) === Transaction.SIGHASH_SINGLE) {
+    var nOut = inIndex
+
+    // only lock-in the txOut payee at same index as txIn
+    // https://github.com/bitcoin/bitcoin/blob/master/src/test/sighash_tests.cpp#L60
```
weilu: Link to the sha instead of master: https://github.com/bitcoin/bitcoin/blob/90c37bc16c3416f133fb15b726873960e00f2262/src/test/sighash_tests.cpp#L60

src/transaction.js
dcousens: In bitcoind,  this can be enabled/disabled,  we don't have a logging
infrastructure for doing that,  should we?  IMHO no.

src/transaction.js
dcousens: Will do

src/transaction.js
dcousens: Is the current so bad?  Its not like this hex is parsing at run time?

PR #401, opened 2015-04-28
Remove instantiable Address

dcousens:
~~WIP pull request.~~ Ready to merge

### DIFF
--- a/src/address.js
+++ b/src/address.js
@@ -1,4 +1,3 @@
-var assert = require('assert')
 var base58check = require('bs58check')
 var typeForce = require('typeforce')
 var networks = require('./networks')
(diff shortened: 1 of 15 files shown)


src/address.js
```diff
@@ -1,62 +1,67 @@
...
-  var scriptType = findScriptTypeByVersion(this.version)
+    var version = payload.readUInt8(0)
 
-  if (scriptType === 'pubkeyhash') return scripts.pubKeyHashOutput(this.hash)
-  if (scriptType === 'scripthash') return scripts.scriptHashOutput(this.hash)
+    assert(version === network.pubKeyHash || version === network.scriptHash)
 
-  assert(false, this.toString() + ' has no matching Script')
+  } catch (e) {
+    throw new Error(address + ' is not a valid ' + network + ' address')
```
weilu: Does validate have to throw on invalid? Consider returning a boolean instead?

src/address.js
```diff
@@ -1,62 +1,67 @@
...
-function Address (hash, version) {
-  typeForce('Buffer', hash)
+  } else if (scripts.isScriptHashOutput(script)) {
+    version = network.scriptHash
+    hash = script.chunks[1]
 
-  assert.strictEqual(hash.length, 20, 'Invalid hash length')
-  assert.strictEqual(version & 0xff, version, 'Invalid version byte')
+  } else {
+    assert(false, script.toASM() + ' has no matching Address')
```
weilu: s/Address/address

src/address.js
dcousens: Sure.

weilu:
@dcousens Let's finish this.

dcousens:
@weilu sounds good,  I'll get back on this today.

dcousens:
Tests failing for unrelated reasons.  I intend to fix that.

@weilu please review.
Should be quickest to review commit-by-commit.

The API could maybe change,  but lets do that in a different PR,  I want to get this merged and leave the API bike shedding separate. 

@jprichardson feel free to review also. :)

dcousens:
Intend to merge ASAP.  Any comments/qualms?

jprichardson:
Most of it looks good. I like this change. My only thought is not removing the tests (you'd still keep this commit as is https://github.com/bitcoinjs/bitcoinjs-lib/commit/16a711ba3309e0a4b8f0751cab5b1ab6e7366c40 i.e. remove these tests as you have). But by not removing the other tests and maybe just changing them to errors helps to communicate how the expectations have changed. I'm fine with it as is, just a thought.

Also, what's with bundling the commit to add the network into TxBuilder into this PR?

dcousens:
@jprichardson please see above,  I rebased the `TransactionBuilder` commit so it is inline with the other changes which I hadn't pushed! (Good spot! :))

dcousens:
@jprichardson never mind,  I rebased without that commit and will submit in a different PR.

src/address.js
```diff
@@ -13,50 +12,55 @@ function findScriptTypeByVersion (version) {
...
-Address.fromBase58Check = function (string) {
+function fromBase58Check (string) {
   var payload = base58check.decode(string)
+  if (payload.length !== 21) throw new TypeError('Invalid address length')
+
   var version = payload.readUInt8(0)
   var hash = payload.slice(1)
 
-  return new Address(hash, version)
+  return { hash: hash, version: version }
```
weilu: Not a big fan of the return value of this function. How about splitting this into two functions: `hashFromBase58Check` and `versionFromBase58Check`

src/address.js
```diff
@@ -13,50 +12,55 @@ function findScriptTypeByVersion (version) {
...
 }
 
-Address.fromOutputScript = function (script, network) {
+function fromOutputScript (script, network) {
   network = network || networks.bitcoin
 
-  if (scripts.isPubKeyHashOutput(script)) return new Address(script.chunks[2], network.pubKeyHash)
-  if (scripts.isScriptHashOutput(script)) return new Address(script.chunks[1], network.scriptHash)
+  if (scripts.isPubKeyHashOutput(script)) return toBase58Check(script.chunks[2], network.pubKeyHash)
+  if (scripts.isScriptHashOutput(script)) return toBase58Check(script.chunks[1], network.scriptHash)
```
weilu: Similarly here, an object with `hash` and `version` seems like a half backed internal representation. I'd rather this returning a base58 encoded string instead.

weilu:
There will be a PR for adding deprecation warnings on 1.x branch right?

dcousens:
`fromBase58Check` is temporary IMHO.

> There will be a PR for adding deprecation warnings on 1.x branch right?

Of course. 

> Add a possibleNetworks here as the 2nd argument, and pass it to findNetworkByWIFVersion for network detection.

I think that's worth investigating,  but we need to do *consistently* this across all the network sections, and I'd be happy to do that in https://github.com/bitcoinjs/bitcoinjs-lib/pull/425.

src/address.js
dcousens: Huh?  It doesn't return an `object`,  it returns a base58check encoded string.

dcousens:
Going to merge, as the issues brought up (here and privately) were all expected:

* `fromBase58check` feels like an internal API,  but it still useful,  lets find a decent replacement (PR to come).
* Network detection cleanup (will be addressed in https://github.com/bitcoinjs/bitcoinjs-lib/pull/425)
* Deprecation warnings (will put a note on the `1.x` branch for pre-2.0.0 tasks)

dcousens:
@coveralls, you trippin

PR #403, opened 2015-04-28
ECDSA: Further cleanup

dcousens:
Removes `verifyRaw` from `ecdsa`.

It is a breaking changing, therefore can only be put in `2.0.0`.

### DIFF
--- a/src/ecdsa.js
+++ b/src/ecdsa.js
@@ -104,7 +104,7 @@ function sign (curve, hash, d) {
   return new ECSignature(r, s)
 }
 
-function verifyRaw (curve, e, signature, Q) {
+function verify (curve, hash, signature, Q) {
   var n = curve.n
   var G = curve.G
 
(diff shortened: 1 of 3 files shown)


jprichardson:
ACK

src/ecdsa.js
```diff
@@ -223,6 +223,5 @@ module.exports = {
   deterministicGenerateK: deterministicGenerateK,
   recoverPubKey: recoverPubKey,
   sign: sign,
-  verify: verify,
-  verifyRaw: verifyRaw
```
weilu: Note: major version bump

dcousens: Already noted as breaking in 1.5.x branch,  and this is 2.0.0 Dev.

dcousens:
Breaking tag removed, as this is superseded by the breaking change in https://github.com/bitcoinjs/bitcoinjs-lib/pull/420

PR #405, opened 2015-05-06
Update 1.5.x hdnode.js

jsantirso:
Informed the pubkey when building an HDNode from an ECKey (otherwise, derivation would fail on "this.pubKey.toBuffer()")


dcousens:
Thanks @jsantirso, will add tests post merge, if possible please include them next time.

jsantirso:
I will, sorry; I had just started playing with the library and was not confident enough about my understanding of it.
Thank you for your amazing work

PR #406, opened 2015-05-07
tests: use strictEqual always

dcousens:
@jprichardson please merge after ACK :)

Like always, lets enforce strict equality wherever possible.

### DIFF
--- a/test/hdnode.js
+++ b/test/hdnode.js
@@ -66,8 +66,8 @@ describe('HDNode', function () {
         var network = networks[f.network]
         var hd = HDNode.fromSeedHex(f.master.seed, network)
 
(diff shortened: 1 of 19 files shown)


test/hdnode.js
```diff
@@ -180,25 +180,25 @@ describe('HDNode', function () {
       var hd = HDNode.fromBase58(f.master.base58)
       var hdn = hd.neutered()
 
-      assert.equal(hdn.keyPair.d, null)
```
dcousens: This was the only test that broke,  the `keyPair.d` is actually `undefined`, not `null`, in this instance.

weilu:
:+1:

PR #410, opened 2015-05-29
package: specify engines

dcousens:
@rubensayshi does this resolve the issue you were having?

Will back port to 1.5.x if merged.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,10 @@
     "client",
     "library"
   ],
+  "engines" : {
+    "iojs": "^1.0.0",
+    "node": ">=0.12.0"
+  },
   "contributors": [
     {
       "name": "Daniel Cousens",


rubensayshi:
1) you made a typo, it should be:
```
  "engines" : {
    "iojs": "^1.0.0",
    "node": ">=0.12.0"
  },
```
but you have it wrapped in another pair of `{}`


2) this gives a warning: 
npm WARN engine bitcoinjs-lib@2.0.0-pre: wanted: {"iojs":"^1.0.0","node":">=0.12.0"} (current: {"node":"0.10.36","npm":"1.4.28"})

there's `"engineStrict": true,` but that's deprecated, I guess we'll have to do with the warning :)

dcousens:
Thanks @rubensayshi, fixed the syntax errors.

weilu:
Remind me, what are we using that's only supported in 0.12?

dcousens:
@weilu `Buffer.compare`,  perhaps its worth leaving the shim for it?

Reference: https://github.com/bitcoinjs/bitcoinjs-lib/commit/75540b61169a7975944d8b5bc0e12dcbf50f838d

dcousens:
@weilu note, this hasn't landed in `1.5.x` because it was a breaking change.   I'm open to keeping compatibility with `0.10` if necessary, but this was an obvious change if we didn't keep compatibility.

@rubensayshi thoughts?

dcousens:
To revert, we would roll back 75540b6 and 706666024ebfa0dfaa06484215bc0c260d2aa247

weilu:
I can't find any usage stats of v0.12 vs. 0.11 vs. 0.10. Given 0.12 is only 3 months old, I'm skeptical about dropping 0.11 and 0.10 support. I vote for leaving the shim in. Sorry that I missed those two commits earlier.

dcousens:
@weilu I'm happy with that.   Shall we close this, and revert those changes and I'll put it up in a new PR?

weilu:
@dcousens Reverted. Just open a new PR.

dcousens:
This PR is unnecessary given the reverts.  I'm happy to leave this closed.  @rubensayshi?

rubensayshi:
still doesn't hurt to specify it right now as `> 0.10.0` as a reminder to update it once it's relevant I think :)

dcousens:
@rubensayshi that is hard to quantify, when can we measurable say it is relevant?  For me, `0.10` is already the past,  but that varies for a lot of people.

rubensayshi:
what I meant was, put `"engines": {"node": ">=0.11.0"}` in the `package.json` so that we might see it at some point and be reminded to bump it when neccesary.

as for what version to support etc... idk if you guys had a discussion about that already or not ...

I agree that `0.10` is old and we should at least target `>= 0.11`  
I also agree that moving forward is good and sticking to old versions just for the convenience of users who don't feel like upgrading sucks (\*cough\* python \*cough\*),  

but for blocktrail where we use bitcoinjs-lib to provide an SDK to our customers, amongst our customers there will be people who are terribly slow at upgrading (or don't upgrade at all) and we have an interest in support as wide of a range of versions as possible.

I don't think that dropping `0.11` support over a single usage of `Buffer.compare` is worth it, it should be something 'bigger' imo.

/ping @weilu @jprichardson

jprichardson:
> I don't think that dropping 0.11 support over a single usage of Buffer.compare is worth it, it should be something 'bigger' imo.

Totally agree. A `Buffer.compare` polyfill could created easily. No reason to drop support for `0.10` just yet.

PR #411, opened 2015-05-29
fix txb.sign having issues when an incomplete TX contains OP_0s, but not enough

rubensayshi:
when I did my first PR to deal with the out of order signatures etc this was I think in there but we removed it because there was no use/test case that required it.

unfortunatly I had a client who was still using a old version of our SDK that used `bitcoinjs-lib@1.4.4` and it produces only OP_0s for the precending pubkeys, so if you sign key with 2 - out of 3 - then you get `OP_0 SIG_2` for signatures (and not `OP_0 SIG_2 OP_0`).

this fixes the signing breaking if this occurs.

testcase is hardcoded instead of fixtures because I had to start out with a partially signed half broken raw hex

you can cherry-pick https://github.com/blocktrail/bitcoinjs-lib/commit/442835b7dd1cbccaa267416971a47504c39eb35e for `1.5.x` version of this


dcousens:
Thanks for this @rubensayshi, I'll review this asap :)

weilu:
@rubensayshi Would you consider move the signature check to [ecdsa's verify function](https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/src/ecdsa.js#L107) instead? It'd make the verify function more robust, and easier to test.

dcousens:
@weilu hmmm, I can't decide if that is a good design or not?  Right now I'd prefer to see an `Error` thrown (as it does) if the signature is missing?

weilu:
@dcousens the rest of `verify` doesn't tell you why it's returning `false`, does it? :P And you know you hate "can't call foo on undefined" as much as I do. One solution I can think of is to pass an optional `debug` argument to `verify` which will populate the `reason` key when `verify` returns false.

dcousens:
> can't call foo on undefined

But, at least then I know the problem.  I'll have a think on this one.  I'm not overwhelmingly against it.

dcousens:
ACK.  Thanks @rubensayshi,

PR #418, opened 2015-06-23
Stricter typing in ECDSA

### LINKED ISSUE #414: ecdsa: handling input types
Continuation of an issue brought up in #411.

This is undoubtedly a javascript library, but I think I'd always prefer a more strongly typed response of something this critical.  Probably even an exception if undefined parameters are given (maybe we should apply `typeforce` on this?)

ping @weilu

dcousens:
Resolves #414.

This is a breaking change, only in the sense that we are more strict about *only* `ECSignature` over `{ r: r, s: s }` objects as inputs now.
I amended some fixtures that were using decimal values, when they should have been using hex.

If consensus isn't met on this,  I'll put up a separate PR for just the test fixes.

### DIFF
--- a/test/fixtures/ecdsa.json
+++ b/test/fixtures/ecdsa.json
@@ -263,8 +263,8 @@
         "d": "01",
         "message": "foo",
         "signatureRaw": {
-          "r": "-fffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364140",
-          "s": "-fffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364140"
+          "r": "-fffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364141",
+          "s": "-fffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364141"
         }
       }
     ]
--- a/test/fixtures/ecdsa.json
+++ b/test/fixtures/ecdsa.json
@@ -192,7 +192,7 @@
         "exception": "nR is not a valid curve point",
         "e": "01",
         "signature": {
-          "r": "fffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364141",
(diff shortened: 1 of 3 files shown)


test/fixtures/ecdsa.json
```diff
@@ -6,70 +6,49 @@
         "k": "ec633bd56a5774a0940cb97e27a9e4e51dc94af737596a0c5cbb3d30332d92a5",
         "message": "Everything should be made as simple as possible, but not simpler.",
         "i": 0,
-        "signature": {
-          "r": "23362334225185207751494092901091441011938859014081160902781146257181456271561",
-          "s": "50433721247292933944369538617440297985091596895097604618403996029256432099938"
```
dcousens: These were all in decimal,  which is just confusing IMHO.  I switched them to DER.

test/fixtures/ecdsa.json
```diff
@@ -191,8 +170,8 @@
         "description": "Invalid r value (nR is infinity)",
         "exception": "nR is not a valid curve point",
         "e": "01",
-        "signature": {
-          "r": "fffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364141",
+        "signatureRaw": {
+          "r": "fffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364140",
```
dcousens: And these were hex, but being interpreted as decimal.
Now they are all hex, and the test fixtures amended.

dcousens:
ping @jprichardson as well

jprichardson:
ACK. This is a very good thing IMHO.

dcousens:
Will merge in 48 hours or after positive review by @weilu.

weilu:
I don't think this is really a breaking change. API remains the same, we are just catching errors earlier.

dcousens:
@weilu I guess,  it just depends if people were relying on that flexibility previously.

dcousens:
This is not a breaking change for `2.0.0`, as `ECDSA` is not exposed.

Issue #394, opened 2015-04-10
Extract ECC operations to 3rd party module


dcousens:
As the title says.

For now, I intend to move it to https://github.com/dcousens/libsecp256k1 (I might move that under the bitcoinjs-lib organisation).

By moving it out we can focus on what is important for this library, in this library; and naturally can focus on things like https://github.com/bitcoinjs/bitcoinjs-lib/issues/361 in an external environment.

By doing this, we will have to naturalize (to built-ins or `Buffer`s) such that our API will be more inter-operable.

Also related was https://github.com/bitcoinjs/bitcoinjs-lib/pull/350.

The reason I'm thinking we don't use https://github.com/cryptocoinjs/ecdsa initially, is because, like https://github.com/bitcoin/secp256k1/,  the extracted library should be purely for secp256k1.

@jprichardson @rubensayshi @polve thoughts?

jprichardson:
> The reason I'm thinking we don't use https://github.com/cryptocoinjs/ecdsa initially, is because, like https://github.com/bitcoin/secp256k1/, the extracted library should be purely for secp256k1.

This doesn't follow. `ecdsa` is only `secp256k1`. I don't think we should use `ecdsa` though.

My position hasn't changed much regarding my comments in those other threads. In summary, we should move towards `elliptic`. However, I can definitely see the case for maximum performance. Is it worth the cost though? 

Wouldn't it make more sense to prepare to migrate towards `elliptic` first and then tackle the issue of migrating towards https://github.com/bitcoin/secp256k1/ later? Or does it make more sense to move towards https://github.com/dcousens/libsecp256k1 (marriage of  https://github.com/bitcoin/secp256k1/ +  https://github.com/cryptocoinjs/ecurve) now?

dcousens:
Oh? I wasn't aware of that (sorry).

As for the migration plan, elliptic simply isn't ready yet,  but this move
should be done in such a way that gets us API parity, so we aren't waiting
wasting effort.

rubensayshi:
I think creating some wrapper so we can easily switch which package is used would be great and we should go straight for that instead of taking intermediate steps.  
Once that is done it's also easy to add any other implementations if there's a desire for it.

I don't like the name `libsecp256k1`, maybe `secp256k1-wrapper` or something ...  
`libsecp256k1` suggests to me that it's another implementation

ryanxcharles:
[...]

PR #420, opened 2015-06-23
index: stop exposing ecdsa

dcousens:
In preparation of #394.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -3,7 +3,6 @@ module.exports = {
   Block: require('./block'),
   bufferutils: require('./bufferutils'),
   crypto: require('./crypto'),
-  ecdsa: require('./ecdsa'),
   ECPair: require('./ecpair'),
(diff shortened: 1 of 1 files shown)


jprichardson:
It's not clear to me why we would want to do this at this point? Shouldn't we get #394 resolved first and then do this?

dcousens:
@jprichardson the idea is that we can iterate on #394 without worrying about interrupting the public API.

weilu:
LGTM. Please add deprecation warning on 1.x branch though.

dcousens:
@jprichardson if you have further concerns with this,  please bring them up ASAP.

jprichardson:
The problem is that `ECDSA` shouldn't have exposed the `ECC` primitives. (I guess a case could be made for performance) But given that's what we have now, I understand the reasoning behind this PR.

PR #421, opened 2015-06-23
Add an 'id' to vin in transactions

anders94:
It is handy to have the transaction id on vins in standard format rather than having to manually reverse the byte order and stringify to hex every time. This adds an "id" to all vins inside transactions.

### DIFF
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -62,6 +62,7 @@ Transaction.fromBuffer = function (buffer, __disableAssert) {
     if (Transaction.isCoinbaseHash(hash)) {
       tx.ins.push({
         hash: hash,
+        id: bufferutils.reverse(hash).toString('hex'),
         index: readUInt32(),
         script: readGenerationScript(),
         sequence: readUInt32()
@@ -69,6 +70,7 @@ Transaction.fromBuffer = function (buffer, __disableAssert) {
     } else {
       tx.ins.push({
         hash: hash,
+        id: bufferutils.reverse(hash).toString('hex'),
         index: readUInt32(),
         script: readScript(),
         sequence: readUInt32()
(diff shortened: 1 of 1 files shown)


dcousens:
Granted,  but the `Transaction` object is meant to be as close to a serialized transaction as possible, and ideally as fast possible.

I'm not sure if this is the best approach,  I've personally not run into this problem prior.  What is your use case @anders94? (if you don't mind)

anders94:
In my case, I'm reading a Transaction from hex and looking up the inputs in a block explorer. It is part of a tool to gather the necessary data to sign a transaction.

I'm open to any suggestion for a better place. It seemed to make sense to have the common representation in the object. Your point on keeping the object as close to serializable as possible is a good one though. I had thought about a more formalized Output object but that doesn't fit either. Do you have another suggestion? A helper method somewhere?

anders94:
Might make more sense in TransactionBuilder.

weilu:
Transaction already has an `getId` method: https://github.com/bitcoinjs/bitcoinjs-lib/blob/97c4de9da62e44d49a237e933b472935fb1ea410/src/transaction.js#L268 @anders94 does it work for you? I don't think we should have the same piece of information stored as two different formats on the transaction object.

anders94:
The transaction has a .getId() but inputs inside a transaction lack that convenience function. I suppose I could reconstitute the previous transactions from each of the inputs and .getId() them. @weilu is that what you are suggesting?

dcousens:
@anders94 as annoying as it may be,  `bufferutils.reverse(hash).toString('hex')` is about as good as this *should* get IMHO.  If you find yourself doing it often,  write a small convenience function?  Or better yet, pre-process your data to add what you need.

If we add it,  we are putting the burden of its presence on *everyone* who uses this library; and I know I certainly don't want it there.

PR #423, opened 2015-07-07
adhere more closely to BIP66

dcousens:
See pseudo code for BIP66 [here](https://github.com/bitcoin/bips/blob/master/bip-0066.mediawiki#DER_encoding_reference),  with [ordering amendments](https://github.com/bitcoin/bips/pull/169).

The constant amendments are because we ignore the SIGHASH byte in this function, which is why the constants are `8, 72` for the min/max sizes, and `6` instead of `7` for the total length offset.

### DIFF
--- a/src/ecsignature.js
+++ b/src/ecsignature.js
@@ -36,53 +36,25 @@ ECSignature.parseCompact = function (buffer) {
 // NOTE: SIGHASH byte ignored
 ECSignature.fromDER = function (buffer) {
   // Format: 0x30 [total-length] 0x02 [R-length] [R] 0x02 [S-length] [S]
-
-  // Minimum and maximum size constraints.
   if (buffer.length < 8) throw new Error('DER sequence too short')
   if (buffer.length > 72) throw new Error('DER sequence too long')
-
-  // A signature is of type 0x30 (compound).
   if (buffer[0] !== 0x30) throw new Error('Not a DER sequence')
-
-  // Make sure the length covers the entire signature.
   if (buffer[1] !== buffer.length - 2) throw new Error('Invalid sequence length')
-
-  // Check whether the R element is an integer.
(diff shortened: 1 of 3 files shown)


dcousens:
ping @weilu @jprichardson

jprichardson:
Since we discussed creating modules that represent BIPs, do you think that it makes sense to turn BIP66 functionality into a module?

dcousens:
@jprichardson that sounds like a great way,  however,  lets not let it slow down this PR?  We can extract it very easily after :)

src/ecsignature.js
```diff
@@ -32,39 +32,37 @@ ECSignature.parseCompact = function (buffer) {
...
+
+  var lenS = buffer[5 + lenR]
+  if (lenS === 0) throw new Error('S length is zero')
+  if ((lenR + lenS + 6) !== buffer.length) throw new Error('Invalid DER encoding (2)')
+
+  if (buffer[4] & 0x80) throw new Error('R value is negative')
+  if (lenR > 1 && (buffer[4] === 0x00) && !(buffer[5] & 0x80)) throw new Error('R value excessively padded')
+
+  if (buffer[lenR + 6] & 0x80) throw new Error('S value is negative')
+  if (lenS > 1 && (buffer[lenR + 6] === 0x00) && !(buffer[lenR + 7] & 0x80)) throw new Error('S value excessively padded')
```
weilu: Perhaps we should extract all above validation into a function `validateDER` with reference to bip66?

weilu:
The implementation looks right to me compared to the reference code. Do we have test case for each one of them though?

src/ecsignature.js
dcousens: I will later,  in fact I might even pull this out to a `bip66` module,   that way it could be super clear.

dcousens:
@weilu we do have test cases for each.

Issue #402, opened 2015-04-28
Parameterize network information detection


dcousens:
Remaining aspects of https://github.com/bitcoinjs/bitcoinjs-lib/issues/325

We should add `network` to all parts of the API that need it~~, with no defaulting~~.

~~That means `ECPair.fromWIF` would require a target network, with an `Error` thrown if the wrong network is given.~~
Same with `HDNode.fromBase58`.

dcousens:
This part of #325 was probably the most controversial,  and likely demands more discussion.
I know as a user of this library, in any cases where I alternate between testnet and mainnet for testing anyway, my code is littered with the network information anyway, so this just enforces that behaviour for me.

But it does make the examples a little more verbose.

jprichardson:
Disagree with not having a default `network`. Maybe it's just a matter of taste for me. But it seems that it should always default to the bitcoin network. I understand that it may add more clarity, but at the cost of verbosity, that's for sure.

dcousens:
@jprichardson its an interesting problem to be sure.   Defaults are powerful.   But in this case they can also be a source of error.

I know for a fact that they would be useful in avoiding little frustrating errors that pop up often,  but they don't appear to be too harmful as the error is usually immediately obvious.

dcousens:
I think defaults are usually acceptable if the result is still correct, but maybe just not ideal.  In this case, the result is not ideal nor correct.

dcousens:
Ping @weilu, @rubensayshi for any further discussion and thoughts?  Yay or nay on this?

rubensayshi:
I like having a default network, 

but I have to admit, I've personally build everything I did so far to support both testnet and mainnet and thus specify the network all the time.  
So my regular argument of 'if 90+% of the use cases are X then it should be the default' doesn't seem to apply as much.

regardless I think a default doesn't hurt and makes life easier for a group of users, including most likely people who try out the library for the first time

weilu:
Provide a consistent default network when possible. In case of detection of a network, require a list of possible networks passed in.

dcousens:
> In case of detection of a network, require a list of possible networks passed in.

That is an interesting approach.
I guess that would absolve the use of the global networks object entirely making extending the library to alt coins quite simple to explain.

dcousens:
Unless there is further discussion,  I'm actually thinking the best approach for the auto-detected networks will be as @weilu outlined above:

[...]

PR #425, opened 2015-07-24
Remove network detection

dcousens:
See https://github.com/bitcoinjs/bitcoinjs-lib/issues/402 for discussion.


dcousens:
This does not detection entirely, yet. 

> In case of detection of a network, require a list of possible networks passed in.

Any thoughts on how this should work with `ECPair.fromWIF`?
What would be **the** best succinct syntax for that?

src/ecpair.js
```diff
@@ -88,7 +78,21 @@ ECPair.fromWIF = function (string) {
...
   }
 
-  var network = findNetworkByWIFVersion(version)
+  var network
+
+  // list of networks?
+  if (Array.isArray(networks)) {
+    network = networks.filter(function (network) {
+      return network.wif === version
+    }).pop() || {}
```
dcousens: @weilu this could probably be more efficient,  but,  it'll do for now

src/ecpair.js
```diff
@@ -67,7 +57,7 @@ ECPair.fromPublicKeyBuffer = function (buffer, network) {
   })
 }
 
-ECPair.fromWIF = function (string) {
+ECPair.fromWIF = function (string, networks) {
```
dcousens: ping @weilu

src/transaction_builder.js
```diff
@@ -188,7 +190,7 @@ TransactionBuilder.prototype.addOutput = function (scriptPubKey, value) {
 
   // Attempt to get a valid address if it's a base58 address string
   if (typeof scriptPubKey === 'string') {
-    scriptPubKey = Address.toOutputScript(scriptPubKey)
+    scriptPubKey = Address.toOutputScript(scriptPubKey, this.network)
```
weilu: There's a potential issue with this use: `var cb = TransactionBuilder.fromTransaction` always assumes bitcoin as the network. If that's called, followed by a call to `cb.addOutput([a non-bitcoin address], value)` without manually setting the network on `cb` it's no good.

weilu:
integration tests are passing locally. Not sure what's wrong with travis =/

src/transaction_builder.js
dcousens: Good point,  `fromTransaction` will need a `network` parameter also.

dcousens:
Rebased and cherry picked 299cfce1f4b2c5af149a468714421040876d9de5 on top.

dcousens:
@weilu I addressed the above issues,  looking to keep moving forward on this.   Happy for any post reviews.

weilu:
LGTM

PR #426, opened 2015-07-24
use BlockTrail for testnet faucet

dcousens:


rubensayshi:
@dcousens I suggest using `process.env.BLOCKTRAIL_API_KEY || 'c0bd8155c66e3fb148bb1664adc1e4dacd872548'` so that it's possible to set your own api key

Issue #383, opened 2015-03-16
Add example on how to use a custom network


dcousens:
Probably just need to outline something like the following:

``` javascript
var viacoin = {
  bip32: {
    public: 0x0488b21e,
    private: 0x0488ade4
  },
  dustThreshold: 560,
  feePerKb: 100000,
  magicPrefix: '\x18Viacoin Signed Message:\n',
  pubKeyHash: 0x47,
  scriptHash: 0x21,
  wif: 0xc7
}
```

The above is `1.5.x` compatible,  for `2.0.0` see [here](https://github.com/bitcoinjs/bitcoinjs-lib/issues/383#issuecomment-130850159).

ricmoo:
Another useful note would be what fields are required for which parts of the library? For example, to just encode/decode from WIF, only *wif* is necessary.

There do not seem to exist BIP32 prefixes for Namecoin (although I have generated some that generate a prefix of **n**, if interested... Is this the right thing to do? Or just use bitcoin's?), but they can be simply omitted if you are not using HDNode.

It would be nice to know when defining custom networks what is required for my feature set.

jprichardson:
@ricmoo regarding BIP32 prefixes for different coins, I wouldn't recommend doing so as it's pretty redundant (assuming you follow BIP44). See https://github.com/bitcoin/bips/blob/master/bip-0043.mediawiki#node-serialization for more detail. i.e. using `bitcoinjs-lib`with a custom network in most cases will require you making sure that you have the WIF version correct. Proof-of-stake coins do require different deserialization/serialization of transactions (due to the timestamp) - I'm working on a solution to address this, but it may be awhile.

ricmoo:
Oh! That is PERFECT. I have been looking for that info documented somewhere. I noticed a few random alt-coins used their own prefixes.

Thanks. :)

dcousens:
As of `2.0.0`,  the `network` type format would be aligned to this example:

``` javascript
  bitcoin: {
    messagePrefix: '\x18Bitcoin Signed Message:\n',
    bip32: {
      public: 0x0488b21e,
      private: 0x0488ade4
    },
    pubKeyHash: 0x00,
    scriptHash: 0x05,
    wif: 0x80,
    dustThreshold: 546
  },
```

dcousens:
@jprichardson for what its worth,  you will still be required to specify the `network` object when calling `HDNode.fromBase58`,  as it will default to bitcoin otherwise.

Issue #325, opened 2014-12-01
Optional network considered harmful


dcousens:
Reference: https://github.com/bitcoinjs/bitcoinjs-lib/pull/321

Should the interfaces `ECPair.fromWIF` and `HDNode.fromBase58` always require a `network` parameter?
We could use auto-detect based on the version flag, but there comes up the problem that networks recently added to the network constants have duplicate constants that would mean auto detection may be misleading for those unaware.

In this sense, it is considered harmful in its current state, where the bitcoin default can regularly throw off users (especially with `.getAddress()` in <2.0.0) and the auto detect can cause incredibly difficult-to-trace bugs.

dcousens:
My conclusion here was to remove `optional network parameter(s) that default to Bitcoin`, and to provide an `optional network parameter when network information may already be
present` that a user can use to force a certain network.

As for #278,  I think we should simply add instructions for how to add your own network information easily,  and then remove all non top 5? networks.

``` javascript
var networks = require('bitcoinjs-lib').networks

networks.viacoin = {
  magicPrefix: '\x18Viacoin Signed Message:\n',
  bip32: {
    public: 0x0488b21e,
    private: 0x0488ade4
  },
  pubKeyHash: 0x47,
  scriptHash: 0x21,
  wif: 0xc7,
  dustThreshold: 560,
  dustSoftThreshold: 100000,
  feePerKb: 100000
}
```

With `estimateFee` being moved to `Transaction`.

weilu:
Works for me. We should make it clear in our docs that the additional networks should be defined before other function calls to bitcoinjs-lib, alongside what are compulsory/optional fields for a network object.

dcousens:
I vote we remove `estimateFee` entirely, it is misleading, and with the introduction of `Transaction.prototype.byteLength`, you can just do the `Math.ceil(byteLength / feePerKb)` very easily yourself.

Given `feePerKb` now needs to be changed due to https://bitcoin.org/en/developer-reference#estimatefee being a protocol level, non-constant, I think this also makes sense.

This complexity isn't something that should be hidden away.

jprichardson:
> I vote we remove estimateFee entirely, it is misleading, and with the introduction of Transaction.prototype.byteLength, you can just do the Math.ceil(byteLength / 1000) * feePerKb very easily yourself.

Seems reasonable to me.

dcousens:
The remaining part of this issue should be resolved with the merging of `ECPair`.

PR #428, opened 2015-08-01
Add PoS coins (e.g. Peercoin) support

visvirial:
# About

`bitcoinjs-lib` supports alt-coins (e.g. Litecoin, Monacoin and Dogecoin) by passing coin network configurations through its API.
However, for Proof of Stake (a.k.a. "PoS") coins such as Peercoin, Blackcoin and Peercoin-derived cryptocurrencies have slightly different binary format (see below) of transactions and blocks (it is represented as `CTransaction` or `CBlock` respectively, in the original source code), which `bitcoinjs-lib` cannot understand.

The difference of binary formats of blocks and transactions between Bitcoin and Peercoin-like cryptocurrencies are:
* Creation time in transaction: `unsigned int nTime` field in `class CTransaction` ([Peercoin: src/main.h L431](https://github.com/ppcoin/ppcoin/blob/master/src/main.h#L431))
* Block signature: `std::vector<unsigned char> vchBlockSig` field in `class CBlock` ([Peercoin: src/main.h L893](https://github.com/ppcoin/ppcoin/blob/master/src/main.h#L893))

In this pull request, I have added some codes regarding PoS coins support.

# Summary of Changes:

* Added "network.isPoS" in network option
* Added Peercoin network configuration (in "networks.peercoiin")
* Can understand nTime field in CTransaction
* Can understand vchBlockSig field in CBlock
* Added test cases for PoS-type blocks, transactions
* Customized test code in order to support PoS-type coin tests
* Added "network" field in test/fixtures/(block|transaction).js


dcousens:
This is great code @visvirial,  I'm just not sure if it is in scope.
Thoughts @weilu / anyone?

We've been doing our best to remove the network specific aspects of bitcoinjs-lib over the last few months. (See https://github.com/bitcoinjs/bitcoinjs-lib/issues/325, https://github.com/bitcoinjs/bitcoinjs-lib/issues/383 and https://github.com/bitcoinjs/bitcoinjs-lib/issues/402 for a start).

I wonder if this can be managed in such a way that we can maintain modularity?

weilu:
Interesting. I do agree that this is a bit out of scope, but I don't see how we can provide a clean entry point for such extension of the library, due to the procedural nature of deserialization/serialization. To implement support for PoS networks without modifying this library itself, one'd have to extend classes like `Transaction` and `Block`. Let's leave this open and see if anyone has better ideas.

smithb1994:
When in doubt, fork and make a new project... Then just link to it in the readme. We don't have to accommodate every cryptocoin out there, but we can still point people in the right direction.

visvirial:
If the goal of this project is just providing a simple and short library for cryptocurrency,
and customizations like this pull request should be done externally, I suggest the following.

Add a new network option `network.binaryFormat.transaction` and `network.binaryFormat.block` that tells how `bitcoinjs-lib` should read the binary data.
The structure of this new option will be:

```JavaScript
networks.bitcoin.binaryFormat.transaction = [
	// nVersion
	{
		name: 'version',
		type: 'int32',
	},
	/*
	// nTime for PoS coins
	{
		name: 'time',
		type: 'uint32',
	},
	*/
	// vin
	{
		name: 'vin',
		type: 'CTxIn[]',
	},
	// vout
	{
		name: 'vout',
		type: 'CTxOut[]',
	},
	// nLockTime
	{
		name: 'locktime',
		type: 'uint32',
	},
]
```

And then, in `Transaction.fromBuffer()`:

```JavaScript
Transaction.prototype.fromBuffer = function(...) {
	
	(snip)
	
	var read = function (type) {
		switch(type) {
			case 'int32':
			case 'uint32':
				return readUint32();
				break;
			case 'CTxIn[]':
				var len = readVarInt();
				tx.ins = [];
				for(var i=0; i<len; i++) {
					tx.ins.push(readTxIn());
				}
				break;
			(... and so on)
		}
	}
	
	network.binaryFormat.transaction.forEach(field) {
		tx[field] = read(field.type);
	}
	
	(snip)
	
}
```

I think this style of modification is much simpler, easier to maintain and more flexible
than to fork `bitcoinjs-lib` to create a new project or re-defining `Transaction.fromBuffer` externally.

How do you feel my suggestion?

urbien:
another case where support for different transaction structure is needed: Sidechain Elements. I suspect Sidechains will go mainstream quickly.

@mvayngrib, @pgmemk and I started working on a project where we need to support [Asset transactions on Sidechains](https://github.com/ElementsProject/elementsproject.github.io#basic-asset-issuance). 
It would be awesome to use bitcoinlib-js for that, or at least a good way on how to patch it for this use case.

dcousens:
@urbien what features would be needed to support Elements Alpha for example?

dcousens:
> How do you feel my suggestion?

@visvirial I think that is an OK suggestion.  Maybe not `2.0.0`,  but,  `2.x.y` certainly.
I'm not 100% on the syntax though.

dcousens:
Targeting `3.0.0`,  for which we'll re-consider how we are currently doing serialization/deserialization.
Simplicity is key,  and I don't want to discount how important it is that we don't become too generic.

dcousens:
@visvirial do you use IRC?

visvirial:
@dcousens What is the channel and server?

dcousens:
freenode,  #bitcoinjs-dev

mvayngrib:
@dcousens re: elements alpha, it would likely have to be a fork, rather than augmenting this repo. Transaction and TransactionBuilder would need to change pretty seriously to support confidential transactions, and the elements built on top of it like asset issuance/transfer. Have you played with elements yet?

dcousens:
I think I'm going to lean towards closing this.  Peercoin is still around, but, to keep up with development we'd have to massively re-write the library and highly generalize it where generalization isn't really necessary.

@visvirial if you're still interested in this, please re-open,  I'm just not sure what the best approach is.

dcousens:
> It would be awesome to use bitcoinlib-js for that, or at least have a good way of patching it for this use case.

@urbien if this is still a case,  please keep don't hesitate to submit patches that may make patching simpler, provided the changes are in scope for bitcoinjs-lib too.

dcousens:
@visvirial that said, I think https://github.com/bitcoinjs/bitcoinjs-lib/pull/428#issuecomment-126999668 is spot on and might still be worth while doing.

visvirial:
Hi @dcousens .

I hope bitocinjs-lib to be more generic, so I think implementing #513 should be done first
and your decision seems reasonable.
If there are something I can help, please mention me as above !

priestc:
Someone should take these changes and port them over to bitcore, then make a pull request to the altcore project: https://github.com/priestc/altcore-lib

Altcore wants to support all crypto currencies, especially ones like PPC that have many forked currencies descended from it.

PR #431, opened 2015-08-07
use cb-http-client with the BlockTrail API

dcousens:
Reduces the amount of boilerplate code we have here for just querying the blockchain.
Will merge once the quirk RE the test failing is found.


dcousens:
Yay, @weilu all tests passing on TRAVIS.  It feels like months since that has consistently been the case.

PR #432, opened 2015-08-11
Various cleanup

dcousens:
Tiny bit of clean up.
Going to merge, @jprichardson feel free to review post merge.

### DIFF
--- a/src/ecsignature.js
+++ b/src/ecsignature.js
@@ -71,7 +71,7 @@ ECSignature.parseScriptSignature = function (buffer) {
   var hashType = buffer.readUInt8(buffer.length - 1)
   var hashTypeMod = hashType & ~0x80
 
-  assert(hashTypeMod > 0x00 && hashTypeMod < 0x04, 'Invalid hashType ' + hashType)
+  if (hashTypeMod <= 0x00 || hashTypeMod >= 0x04) throw new Error('Invalid hashType ' + hashType)
 
   return {
     signature: ECSignature.fromDER(buffer.slice(0, -1)),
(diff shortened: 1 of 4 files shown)


src/ecsignature.js
```diff
@@ -71,7 +71,7 @@ ECSignature.parseScriptSignature = function (buffer) {
   var hashType = buffer.readUInt8(buffer.length - 1)
   var hashTypeMod = hashType & ~0x80
 
-  assert(hashTypeMod > 0x00 && hashTypeMod < 0x04, 'Invalid hashType ' + hashType)
+  if (hashTypeMod <= 0x00 || hashTypeMod >= 0x04) throw new Error('Invalid hashType ' + hashType)
```
dcousens: Moving away from `assert` as the string warning messages always get built, which has lead us to have various performance issue in the past.
It was nice while it lasted.

src/networks.js
```diff
@@ -3,7 +3,6 @@
 
 module.exports = {
   bitcoin: {
-    magic: 0xd9b4bef9,
```
dcousens: We haven't stored all the `chainParams` that exist [here](https://github.com/bitcoin/bitcoin/blob/master/src/chainparams.cpp#L86-L89).
It seems inconsistent to just have this one,  which feels very out of place among constants that are actually used and tested by this library.

Open to ideas on this one,  removing it for now.  Maybe it should come back as `chain: { magicPrefix: ? }}`?

weilu: Technically yes

weilu:
ACK

Issue #407, opened 2015-05-20
Limit use of non-primitive types throughout the API


dcousens:
As a result of the discussion in https://github.com/bitcoinjs/bitcoinjs-lib/issues/359,  the aim will be to limit the use of non primitives in the API.

`Buffer` is considered a primitive in this case, and is therefore acceptable.

dcousens:
Wherever the API changes,  warnings will need to be added to `1.5.x`.

dcousens:
Best I can tell, the only non-primitive constructor that we have is in `function ECPair (d, Q, options)`.

dcousens:
I didn't realise before,  but `Script` is a major component which could be simplified down to a `Buffer` for interop.

With modules like https://github.com/bitcoinjs/bip69 coming up,  the ability to reduce this part of the API would be asimilar to the changes made to `Address`.

dcousens:
For the record, as per https://github.com/bitcoinjs/bitcoinjs-lib/issues/407#issuecomment-128611249,  `script` was transitioned to a `Buffer` in `2.0.0`.

dcousens:
Related https://github.com/bitcoinjs/bitcoinjs-lib/commit/00308542cddc84ac8417761f5deaeebca4749b4d

dcousens:
We really want to avoid the use of `bigi` in the API for the next version, IMHO

dcousens:
Closing in favour of 508

PR #434, opened 2015-08-12
Remove assert, use typeforce typing

dcousens:
Will wait on ACK before merge.

ping @weilu, @jprichardson 

Curious on your opinions about this...

### DIFF
--- a/src/address.js
+++ b/src/address.js
@@ -1,7 +1,8 @@
 var base58check = require('bs58check')
-var typeForce = require('typeforce')
+var typeforce = require('typeforce')
 var networks = require('./networks')
 var scripts = require('./scripts')
+var types = require('./types')
 
 function fromBase58Check (string) {
   var payload = base58check.decode(string)
@@ -23,10 +24,7 @@ function fromOutputScript (script, network) {
 }
 
 function toBase58Check (hash, version) {
-  typeForce('Buffer', hash)
-
-  if (hash.length !== 20) throw new TypeError('Invalid hash length')
-  if (version & ~0xff) throw new TypeError('Invalid version byte')
+  typeforce(types.tuple(types.Hash160bit, types.UInt8), arguments)
 
   var payload = new Buffer(21)
   payload.writeUInt8(version, 0)
--- a/src/ecdsa.js
+++ b/src/ecdsa.js
@@ -1,6 +1,6 @@
-var assert = require('assert')
 var createHmac = require('create-hmac')
-var typeForce = require('typeforce')
+var typeforce = require('typeforce')
+var types = require('./types')
 
 var BigInteger = require('bigi')
 var ECSignature = require('./ecsignature')
@@ -10,12 +10,12 @@ var ONE = new Buffer([1])
 
 // https://tools.ietf.org/html/rfc6979#section-3.2
 function deterministicGenerateK (curve, hash, d, checkSig) {
-  typeForce('Buffer', hash)
-  typeForce('BigInteger', d)
-  typeForce('Function', checkSig)
-
-  // sanity check
-  assert.equal(hash.length, 32, 'Hash must be 256 bit')
+  typeforce(types.tuple(
+    types.ECCurve,
+    types.Hash256bit,
+    types.BigInt,
+    types.Function
+  ), arguments)
 
   var x = d.toBuffer(32)
   var k = new Buffer(32)
@@ -75,9 +75,7 @@ function deterministicGenerateK (curve, hash, d, checkSig) {
 }
 
 function sign (curve, hash, d) {
-  typeForce('Curve', curve)
-  typeForce('Buffer', hash)
-  typeForce('BigInteger', d)
+  typeforce(types.tuple(types.ECCurve, types.Hash256bit, types.BigInt), arguments)
 
   var e = BigInteger.fromBuffer(hash)
   var n = curve.n
@@ -109,10 +107,11 @@ function sign (curve, hash, d) {
 }
 
 function verify (curve, hash, signature, Q) {
-  typeForce('Curve', curve)
-  typeForce('Buffer', hash)
-  typeForce('ECSignature', signature)
-  typeForce('Point', Q)
+  typeforce(types.tuple(types.ECCurve,
+    types.Hash256bit,
+    types.ECSignature,
+    types.ECPoint
+  ), arguments)
 
   var n = curve.n
   var G = curve.G
@@ -162,20 +161,20 @@ function verify (curve, hash, signature, Q) {
   * http://www.secg.org/download/aid-780/sec1-v2.pdf
   */
 function recoverPubKey (curve, e, signature, i) {
-  typeForce('Curve', curve)
-  typeForce('BigInteger', e)
-  typeForce('ECSignature', signature)
-  typeForce('Number', i)
-  assert.strictEqual(i & 3, i, 'Recovery param is more than two bits')
+  typeforce(types.tuple(
+    types.ECCurve,
+    types.BigInt,
+    types.ECSignature,
+    types.UInt2
+  ), arguments)
 
   var n = curve.n
   var G = curve.G
-
   var r = signature.r
   var s = signature.s
 
-  assert(r.signum() > 0 && r.compareTo(n) < 0, 'Invalid r value')
-  assert(s.signum() > 0 && s.compareTo(n) < 0, 'Invalid s value')
+  if (r.signum() <= 0 || r.compareTo(n) >= 0) throw new Error('Invalid r value')
+  if (s.signum() <= 0 || s.compareTo(n) >= 0) throw new Error('Invalid s value')
 
   // A set LSB signifies that the y-coordinate is odd
   var isYOdd = i & 1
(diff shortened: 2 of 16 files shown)


src/types.js
```diff
@@ -0,0 +1,71 @@
...
+  return typeforce.Number(value) &&
+    value >= 0 &&
+    value <= UINT53_MAX &&
+    Math.floor(value) === value
+}
+
+// external dependent types
+function BigInt (value) { return !typeforce.Null(value) && value.constructor === bigi }
+function ECCurve (value) { return !typeforce.Null(value) && value.constructor === ecurve.Curve }
+function ECPoint (value) { return !typeforce.Null(value) && value.constructor === ecurve.Point }
```
dcousens: @weilu you this does revert our duck typing changes.   I've never had that problem,  so I'm not sure why we allowed it in the first place.

`ECPair` and `ecdsa` are the only interfaces that accept `bigi` / `ECPoint`'s,  and we [intend](https://github.com/bitcoinjs/bitcoinjs-lib/issues/407) to [remove](https://github.com/bitcoinjs/bitcoinjs-lib/issues/394) their individual use entirely for `2.0.0`,  so this shouldn't even be an issue.


Various bugs where internal versions of these libraries could be inconsistent means duck typing is a very bad idea in this case.

test/fixtures/transaction.json
```diff
@@ -229,12 +229,12 @@
   "invalid": {
     "addInput": [
       {
-        "exception": "Expected hash length of 32, got 30",
+        "exception": "Expected 256-bit Buffer, got 240-bit",
```
dcousens: Should this error be `Expected 256-bit hash, got 240-bit`?

src/transaction.js
```diff
@@ -19,7 +19,7 @@ Transaction.SIGHASH_NONE = 0x02
 Transaction.SIGHASH_SINGLE = 0x03
 Transaction.SIGHASH_ANYONECANPAY = 0x80
 
-Transaction.fromBuffer = function (buffer, __disableAssert) {
+Transaction.fromBuffer = function (buffer, __disableExcess) {
```
dcousens: `__disableExcess`, if provides, disables the 'strictness' of `fromBuffer` from checking that the bytes read are equal to the bytes in `buffer`.

src/ecdsa.js
```diff
@@ -10,12 +10,12 @@ var ONE = new Buffer([1])
...
-  typeForce('Function', checkSig)
-
-  // sanity check
-  assert.equal(hash.length, 32, 'Hash must be 256 bit')
+  typeforce(types.tuple(
+    types.ECCurve,
+    types.Hash256bit,
+    types.BigInt,
+    types.Function
+  ), arguments)
```
dcousens: All these types are currently being created on the fly (though, it is pretty light weight), we can pull them out of the function and just enforce them on the first line if we feel performance is an issue,  though,  this new syntax is already 1000% faster than the previous implementation.

dcousens:
I'm happy to break out the `typeforce` typing changes if theres any issues,  as the move away from `assert` is a much simpler decision and it does have highly tangible benefits.

jprichardson:
> as the move away from assert is a much simpler decision and it does have highly tangible benefits.

I think for posterity, these benefits should be explicitly stated.

src/types.js
weilu: I remember an important reason why we went with duck typing is to deal with having multiple versions of the same library in the same app. I still think that's relevant so my preference is still duck typing.

src/transaction.js
weilu: `__lenient` or `__noCheckTransactionData`. I find negated names most of the time confusing. Given that you prepended it with __ I wonder if we can get away with flipping it to `__strict` without breaking the API.

test/fixtures/transaction.json
weilu: I think Buffer is fine

weilu:
I don't get why you are getting rid of asserts. IMHO in most of the places, assert is still more direct and semantically straightforward. It's not like we are throwing any custom Error types for better error handling.

dcousens:
> I think for posterity, these benefits should be explicitly stated.

* In a production application, with over 30 high level dependencies,  `bitcoinjs-lib`, `bs58`, `bip39`, `bs58check`,  are the only dependencies using `assert` for error handling.   All other conclusions aside,  this is a specific pattern we have adopted for our ecosystem.

* `assert` explicitly makes code coverage difficult,  because the branch is deeper down in the application.   `typeforce` changes do not help this,  however,  the majority of `typeforce` changes were `type` checking.  The majority of logic based changes that now exist as `throw` conditions are now visible in code coverage.

* `assert` only accepts value arguments (like any function), therefore error messages are evaluated/built on every call.   This has caused major performance problems for us in the past where `JSON.stringify` or `toString` has been more complex then expected and excessively slowed down applications.

* `assert.equal` is more often used over `assert.strictEqual`,  most likely because it is shorter to write.  We actually have our linter fail our tests if we wrote this code normally, and yet, we have no qualms with this.  Granted,  that could just be fixed by enforcing `strictEqual`,  it is more verbose and often longer code than just writing out the explicit branch anyway.

src/types.js
dcousens: > I remember an important reason why we went with duck typing is to deal with having multiple versions of the same library in the same app.

Right.

> I've never had the issue of de-duping where it wasn't symptomatic of something potentially disastrous anyway, so I'm not sure why we allowed it in the first place.

I feel like such a situation is dangerous enough that we shouldn't let it occur though?

src/transaction.js
dcousens: @weilu the default is strict,  it is only in [specific cases](https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/src/block.js#L47-L53) that you would want to do this.

`__noStrict` or `__lenient` works for me.

dcousens:
Rebased on master.  Added 94e16fb,  6002d9a.

dcousens:
@jprichardson @weilu,  shall I merge or?

weilu:
@dcousens Given that the types of the errors and in some cases the messages are different, I wonder if we need to bump major for this?

dcousens:
@weilu I wasn't going to bother back porting this to `1.5.x`,  the change is worth it,  but it doesn't fix/add anything such that it would be worth back porting.

test/fixtures/transaction_builder.json
```diff
@@ -780,7 +780,7 @@
       },
       {
         "description": "Wrong key pair for multisig redeemScript",
-        "exception": "key pair cannot sign for this input",
+        "exception": "Key pair cannot sign for this input",
```
dcousens: Breaking change

test/fixtures/transaction.json
```diff
@@ -229,12 +229,12 @@
...
         "hash": "0aed1366a73b6057ee7800d737bff1bdf8c448e98d86bc0998f2b009816d",
         "index": 0
       },
       {
-        "exception": "Expected hash length of 32, got 34",
+        "exception": "Expected 256-bit Buffer, got 272-bit",
```
dcousens: Breaking change

test/fixtures/transaction.json
dcousens: Breaking change

test/fixtures/ecdsa.json
```diff
@@ -188,7 +188,7 @@
       },
       {
         "description": "Invalid i value (> 3)",
-        "exception": "Recovery param is more than two bits",
+        "exception": "Expected UInt2, got Number 4",
```
dcousens: Breaking change

dcousens:
I'd say there is only a few breaking changes, outlined above.  And even then,  `ecdsa` is no longer exposed.   The error messages that are different are therefore only for `Transaction`,  the rest should be consistent as `types.tuple` should behave the same as the previous code.

src/address.js
```diff
@@ -23,10 +24,7 @@ function fromOutputScript (script, network) {
 }
 
 function toBase58Check (hash, version) {
-  typeForce('Buffer', hash)
-
-  if (hash.length !== 20) throw new TypeError('Invalid hash length')
-  if (version & ~0xff) throw new TypeError('Invalid version byte')
+  typeforce(types.tuple(types.Hash160bit, types.UInt8), arguments)
```
dcousens: This API is new,  so its not a breaking change.

src/types.js
dcousens: This is the other breaking change,  but,  that has already been pointed out.

weilu:
I'm fine with not back-porting this to 1.5.x but let's add these breaking changes into our 2.0 release note.

weilu:
Good to merge

dcousens:
Thanks @weilu,  I added the breaking change tag so we can easily build that release note by reviewing it later.

dcousens:
Follow up https://github.com/bitcoinjs/bitcoinjs-lib/commit/5d2abb523ee32c7b9039330152134825c0a7c918 for `HDNode`.

src/types.js
dcousens: This change was reverted in https://github.com/bitcoinjs/bitcoinjs-lib/pull/450 as it was still an issue.  `quacksLike` behaviour was restored.

PR #438, opened 2015-08-14
No Script

dcousens:
Ref: https://github.com/bitcoinjs/bitcoinjs-lib/issues/407#issuecomment-128611249

> I didn't realise before, but Script is a major component which could be simplified down to a Buffer for interop.
> With modules like https://github.com/bitcoinjs/bip69 coming up, the ability to reduce this part of the API would be asimilar to the changes made to Address.

Overall, this is part of #407.
The performance is the same,  we just stopped using `Script` as a data interchange.
You are now either explicitly dealing with the decompilation result of a script (aka, chunks) after a `scripts.decompile`,  or you are dealing with a `Buffer` which is a script.

This also means you can now pass your script directly to `Transaction` without having to decompile it first.

I also merged `Script` and `scripts`, which makes the interface smaller and cleaner IMHO,  open to feedback though.  

Chained on #434 , but **not** dependent.

**Comparison link:** https://github.com/bitcoinjs/bitcoinjs-lib/compare/noassert...noscript


dcousens:
My motivation for this is mostly interchange between libraries.  I wanted to be able to agnostically work with scripts,  and yet most of the `bitcoinjs-lib` interface requires me to "pre-parse" the script using `Script.fromBuffer` or what not.
This was especially annoying if I had no need to actually decompile the script anyway,  so it was just a fruitless operation for the sake of an abstraction.

jprichardson:
All of the assert changes / typeforce changes to this PR makes this more painful to review... granularity ftw :)

While I've always thought Script was a useful abstraction, I understand the motivation, especially regarding interoperability. So I am OK with this. Would like to know what @weilu thinks.

dcousens:
> All of the assert changes / typeforce changes to this PR makes this more painful to review... granularity ftw :)

Did you check the comparison link? (See https://github.com/bitcoinjs/bitcoinjs-lib/compare/noassert...noscript)

dcousens:
Waiting on ACK for #434 before merge.

src/scripts.js
```diff
@@ -1,12 +1,129 @@
...
+    // data chunk
+    } else {
+      return new Buffer(strChunk, 'hex')
+    }
+  })
+
+  return compile(chunks)
+}
+
+function compile (chunks) {
```
weilu: Where does the name `compile`/`decompile` come from? Would it help inferring the input and output types better by calling them `chunkArrayToBuffer` and `bufferToChunkArray`?

src/scripts.js
```diff
@@ -1,12 +1,129 @@
...
+for (var op in OPS) {
+  var code = OPS[op]
+  REVERSE_OPS[code] = op
+}
+
+function toASM (chunks) {
+  if (types.Buffer(chunks)) {
+    chunks = decompile(chunks)
+  }
+
```
weilu: Perhaps `typyforce(Array, chunks)` here?

src/scripts.js
```diff
@@ -1,12 +1,129 @@
...
-var ECSignature = require('./ecsignature')
-var Script = require('./script')
+var OPS = require('./opcodes')
+var REVERSE_OPS = []
+for (var op in OPS) {
+  var code = OPS[op]
+  REVERSE_OPS[code] = op
+}
+
+function toASM (chunks) {
```
weilu: For the to and from ASM function pair, perhaps they should be strict reverse of each other in terms of input and output types? Similarly to what has been suggested below, consider adding both the source type and destination types to the function name, e.g. `chunksArrayToASM`

src/scripts.js
```diff
@@ -170,96 +301,91 @@ function classifyInput (script, allowIncomplete) {
...
-
-  return Script.fromChunks([
-    ops.OP_DUP,
-    ops.OP_HASH160,
-    hash,
-    ops.OP_EQUALVERIFY,
-    ops.OP_CHECKSIG
-  ])
+function pubKeyHashOutput (pubKeyHash) {
+  typeforce(typeforce.Buffer, pubKeyHash)
```
weilu: You were checking the size of the buffer in noassert branch. Why giving it up?

src/scripts.js
```diff
@@ -170,96 +301,91 @@ function classifyInput (script, allowIncomplete) {
...
-function scriptHashOutput (hash) {
-  typeForce('Buffer', hash)
-
-  return Script.fromChunks([
-    ops.OP_HASH160,
-    hash,
-    ops.OP_EQUAL
-  ])
+function scriptHashOutput (scriptHash) {
+  typeforce(typeforce.Buffer, scriptHash)
```
weilu: ditto

src/scripts.js
dcousens: `chunks` are essentially uncompiled byte code,  which is then compiled to a `Buffer` for serialization in a `Transaction`.
Seeing as we have to parse a `Buffer` to de-serialize it, using the `OP_CODE` **byte** codes,  it seemed like a fitting analogy.

src/scripts.js
dcousens: Sure,  though `.map` is straight after,  probably not necessary.

src/transaction.js
```diff
@@ -103,19 +97,21 @@ Transaction.isCoinbaseHash = function (buffer) {
   })
 }
 
+var EMPTY = new Buffer(0)
```
weilu: Nit: EMPTY is an adjective. Empty what?

src/scripts.js
dcousens: They [are synonymous already](https://github.com/bitcoinjs/bitcoinjs-lib/pull/438/files#diff-4e2a83a598cdaf6ab71d5de299dcd1f2R27)?

src/scripts.js
dcousens: A mistake,  evidently we need a test :+1:

src/scripts.js
dcousens: A mistake, will fix

src/transaction.js
```diff
@@ -201,11 +192,13 @@ Transaction.prototype.hashForSignature = function (inIndex, prevOutScript, hashT
 
   // in case concatenating two scripts ends up with two codeseparators,
   // or an extra one at the end, this prevents all those possible incompatibilities.
-  var hashScript = prevOutScript.without(opcodes.OP_CODESEPARATOR)
+  var hashScript = scripts.compile(scripts.decompile(prevOutScript).filter(function (x) {
```
weilu: You lost the todo note with this replacement: 
`// FIXME: doesn't work for data chunks, maybe time to use buffertools.compare...`

src/transaction.js
dcousens: Will change to `EMPTY_SCRIPT`

src/transaction.js
dcousens: As a general algorithm,  `without` did not handle data chunks correctly.
Now that we can use `filter` by itself,  the semantics are clear and it is on the user to know that `Buffer`s are not comparable without using a `.compare` or a `bufferutils.equals` equivalent.

In that sense, there is no need for a `TODO` here, we only care about the `opcodes.OP_CODESEPARATOR` byte *not* being the result.

src/scripts.js
weilu: From my point of view, it's not obvious what are the input and output types just by reading the function name.

weilu:
Overall this change does enable the use of primitive/native types, but at the cost of abstraction. It used to be Script is an abstraction with two forms of representation -- chunks and buffer. Now that's no longer the case. I personally still find the version with the Script abstraction easier to reason, but it could be just me.

src/scripts.js
dcousens: A `script` in this sense should be synonymous with a `Buffer`.
In all cases a `Script` was a valid, a `Buffer` should now be used.

`compile` is the only exception to this,  it takes an `Array` of chunks to be serialized into a `Buffer`.

`chunks` are the byte codes in their uncompiled form.
`toASM` takes a `Buffer` and transpiles it back to its human readable equivalent (ASM).
`fromASM` takes a string of human readable opcodes (ASM) and compiles it to a `Buffer`.

**For internal usage and performance reasons**, it is also possible to interchangeably use a `chunks` `Array` to avoid constant decompilation/compilation.

weilu:
Granted I don't work with script enough, so if @dcousens feels strong enough to merge this, then go ahead :)

dcousens:
After further discussion of `compile` v `decompile` with @weilu,  I'm now on the fence as to whether `compileToBuffer` would be a better name to indicate the target type more clearly.

dcousens:
Rebased on master

dcousens:
All nits/tests fixed.
Last remaining question is `compile` v `compileToBuffer`.
If no response within 24 hours or so, will merge as `compileToBuffer`.

weilu:
`compileToBuffer` is better IMO

dcousens:
@weilu undecided about that, so I opened https://github.com/bitcoinjs/bitcoinjs-lib/pull/442 so I can move on and hopefully just have it resolved before `2.0.0`.

PR #441, opened 2015-08-17
address: consistent errors

dcousens:
This changes gives a more meaningful error for an invalid address,  while also making the exception message checking easier by making the exception format always contain the input value first (last would be fine too).

### DIFF
--- a/src/address.js
+++ b/src/address.js
@@ -4,9 +4,10 @@ var networks = require('./networks')
 var scripts = require('./scripts')
 var types = require('./types')
 
-function fromBase58Check (string) {
-  var payload = base58check.decode(string)
-  if (payload.length !== 21) throw new TypeError('Invalid address length')
+function fromBase58Check (address) {
+  var payload = base58check.decode(address)
+  if (payload.length < 21) throw new TypeError(address + ' is too short')
+  if (payload.length > 21) throw new TypeError(address + ' is too long')
 
   var version = payload.readUInt8(0)
   var hash = payload.slice(1)
@@ -37,7 +38,8 @@ function toOutputScript (address, network) {
   network = network || networks.bitcoin
 
   var payload = base58check.decode(address)
-  if (payload.length !== 21) throw new TypeError('Invalid hash length')
+  if (payload.length < 21) throw new TypeError(address + ' is too short')
+  if (payload.length > 21) throw new TypeError(address + ' is too long')
 
   var version = payload.readUInt8(0)
   var hash = payload.slice(1)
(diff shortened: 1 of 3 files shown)


test/address.js
```diff
@@ -20,17 +20,17 @@ describe('Address', function () {
...
+          Address.fromBase58Check(f.address)
+        }, new RegExp(f.address + ' ' + f.exception))
       })
     })
   })
 
   describe('fromOutputScript', function () {
     fixtures.valid.forEach(function (f) {
-      it('parses ' + f.script + ' (' + f.network + ')', function () {
+      it('parses ' + f.script.slice(0, 30) + '... (' + f.network + ')', function () {
```
dcousens: I also made the test descriptions shorter for readability

dcousens:
@weilu please ACK post merge.

src/address.js
```diff
@@ -4,9 +4,10 @@ var networks = require('./networks')
 var scripts = require('./scripts')
 var types = require('./types')
 
-function fromBase58Check (string) {
-  var payload = base58check.decode(string)
-  if (payload.length !== 21) throw new TypeError('Invalid address length')
+function fromBase58Check (address) {
+  var payload = base58check.decode(address)
+  if (payload.length < 21) throw new TypeError(address + ' is too short')
+  if (payload.length > 21) throw new TypeError(address + ' is too long')
```
dcousens: @weilu what are your thoughts on multiple exceptions in this case?  I know I've found it useful when dealing with the strict `DER` signature format.

dcousens:
Fixed failing tests in https://github.com/bitcoinjs/bitcoinjs-lib/commit/eb752f83165e06d7dae2856f23c567758441cadd

src/address.js
weilu: Nobody ever complains about the error message is too detailed. 

Nit: You can assign the error message based on the length check, and throw error based on the error message check.

weilu:
GLTM

src/address.js
dcousens: @weilu as in:

``` javascript
if (payload.length !== 21) {
  if (payload.length > 21) throw ...
  throw ...
}
```

weilu: I was talking about
```
var errorMessage;
if (payload.length < 21)  errorMessage = 'too short' 
else if (payload.length > 21)  errorMessage = 'too long' 

if (errorMessage !== undefined) throw new TypeError(errorMessage)
```

Meh, it was shorter in my mind :P Never mind then.

PR #442, opened 2015-08-19
scripts: rename compile to compileToBuffer (bikeshed)

dcousens:
Still not sure about this,  so I figured I'd put it on its own.
`decompileToChunks` seems very context specific. 

I feel like these names could be resolved by just basic documentation?

### DIFF
--- a/src/address.js
+++ b/src/address.js
@@ -18,7 +18,7 @@ function fromBase58Check (address) {
 function fromOutputScript (script, network) {
   network = network || networks.bitcoin
 
-  var chunks = scripts.decompile(script)
+  var chunks = scripts.decompileToChunks(script)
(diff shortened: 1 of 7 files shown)


weilu:
I still prefer more specific names. We don't have documentation, yet.

dcousens:
I think we'll resolve this with documentation in `2.0.0`.
The names while effective at inferring the types,  seem out of place when you are using the ambidextrous nature of the functions internally.
They also become cumbersome:

`scripts.decompileToChunks(scripts.compileToBuffer(chunks))`

vs

`scripts.decompile(scripts.compile(chunks))`

While useful,  its also hard to decide whether the name should always include the original type,  such as: `scripts.decompileBufferToChunks`.
I'm not sure if we would want to do that verbosity of function name across *all* our free functions...

dcousens:
Open to discussion,   but,  again,  lets see a convention first, then go with it.

PR #445, opened 2015-08-20
HDNode: avoid creating multiple buffers

dcousens:


src/hdnode.js
```diff
@@ -194,29 +192,23 @@ HDNode.prototype.toBase58 = function (__isPrivate) {
...
     if (!this.keyPair.d) throw new TypeError('Could not derive hardened child key')
 
     // data = 0x00 || ser256(kpar) || ser32(index)
-    data = Buffer.concat([
-      this.keyPair.d.toBuffer(33),
-      indexBuffer
-    ])
+    data[0] = 0x00
+    this.keyPair.d.toBuffer(32).copy(data, 1)
+    data.writeUInt32BE(index, 33)
```
dcousens: Fresh eyes can be quite useful sometimes.

src/hdnode.js
```diff
@@ -194,29 +192,23 @@ HDNode.prototype.toBase58 = function (__isPrivate) {
...
 
   // Normal child
   } else {
     // data = serP(point(kpar)) || ser32(index)
     //      = serP(Kpar) || ser32(index)
-    data = Buffer.concat([
-      this.keyPair.getPublicKeyBuffer(),
-      indexBuffer
-    ])
+    this.keyPair.getPublicKeyBuffer().copy(data, 0)
```
dcousens: The serialized public key is always 32 bytes,  with an added compression flag (total 33 bytes).  If that flag isn't there, it would be only 32... that flag *should* be *always* there, but maybe someone is doing something really odd?

dcousens: The compressed public key is always 33 bytes, unless someone forces it to be `uncompressed`.  Do we need to allow that?
Right now, it will silently pass,  but crop the public key to just the X coordinate.

weilu: HDNode constructor takes a keypair which could very well be an uncompressed one, so I think we should allow that.

src/hdnode.js
weilu: wow why isn't this tested?

src/hdnode.js
weilu: @dcousens pointed out that key here needs to be compressed for HDNode as in bip32 spec. If user passes in a compressed keypair to pass our validation, then changed it to be uncompressed, they are really screwing with themselves. I don't think we need to worry about that case.

src/hdnode.js
dcousens: No change of functionality here, only an optimization.

src/hdnode.js
dcousens: Exactly,  we even validate that in https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/src/hdnode.js#L17.
They have to be deliberately screwing it to force a change.
Perhaps `.compression` shouldn't be public... `getCompression()`?

weilu: > Perhaps .compression shouldn't be public... isCompressed()?
I like that better.

dcousens: Currently extracting the ECDSA code as part of the final parts of https://github.com/bitcoinjs/bitcoinjs-lib/issues/407 and https://github.com/bitcoinjs/bitcoinjs-lib/issues/394.

Will add it.

PR #446, opened 2015-08-20
ECPair: extract WIF to 3rd party module

dcousens:
I needed this aspect of the code, by itself, in a solo module for a project.

It seems fitting that it is probably a good idea to extract this specific code anyway,  makes `ECPair` even cleaner.

### DIFF
--- a/src/ecpair.js
+++ b/src/ecpair.js
@@ -68,8 +68,6 @@ ECPair.fromWIF = function (string, network) {
     network = network.filter(function (network) {
       return version === network.wif
     }).pop() || {}
-
-    if (version !== network.wif) throw new Error('Invalid network version')
   }
 
   var decoded = wif.decodeRaw(network.wif, buffer)
--- a/src/ecpair.js
+++ b/src/ecpair.js
@@ -5,6 +5,7 @@ var ecurve = require('ecurve')
 var randomBytes = require('randombytes')
 var typeforce = require('typeforce')
 var types = require('./types')
+var wif = require('wif')
 
 var NETWORKS = require('./networks')
 var BigInteger = require('bigi')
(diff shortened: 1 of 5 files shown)


test/fixtures/ecpair.json
```diff
@@ -86,16 +86,20 @@
...
-        "exception": "Invalid WIF payload length",
-        "string": "7ZEtRQLhCsDQrd6ZKfmcESdXgas8ggZPN24ByEi5ey6VJW"
+        "exception": "Invalid WIF length",
+        "WIF": "3tq8Vmhh9SN5XhjTGSWgx8iKk59XbKG6UH4oqpejRuJhfYD"
       },
       {
-        "exception": "Invalid WIF payload length",
-        "string": "5qibUKwsnMo1qDiNp3prGaQkD2JfVJa8F8Na87H2CkMHvuVg6uKhw67Rh"
+        "exception": "Invalid WIF length",
+        "WIF": "38uMpGARR2BJy5p4dNFKYg9UsWNoBtkpbdrXDjmfvz8krCtw3T1W92ZDSR"
```
dcousens: These fixtures were changed because the encoded version was `0x05` in each case, which meant `Invalid network version` was being thrown each time as it was being checked first, not last in the sub-module.

weilu:
@dcousens please make https://github.com/dcousens/wif public. Currently it's returning 404

src/ecpair.js
```diff
@@ -57,46 +58,25 @@ ECPair.fromPublicKeyBuffer = function (buffer, network) {
...
-  // otherwise, assume a network object (or default to bitcoin)
-  } else {
-    network = networks || NETWORKS.bitcoin
+    if (version !== network.wif) throw new Error('Invalid network version')
   }
 
-  if (version !== network.wif) throw new Error('Invalid network')
-
-  var d = BigInteger.fromBuffer(payload)
+  var decoded = wif.decodeRaw(network.wif, buffer)
```
weilu: Can we please not swap the order of version and buffer here? I'd prefer it being consistent for bitcoinjs-lib and wif.

src/ecpair.js
```diff
@@ -57,46 +58,25 @@ ECPair.fromPublicKeyBuffer = function (buffer, network) {
...
-  if (Array.isArray(networks)) {
-    network = networks.filter(function (network) {
+    network = network.filter(function (network) {
       return version === network.wif
     }).pop() || {}
 
-  // otherwise, assume a network object (or default to bitcoin)
-  } else {
-    network = networks || NETWORKS.bitcoin
+    if (version !== network.wif) throw new Error('Invalid network version')
```
weilu: Isn't this double-checking what's already being checked in `decodeRaw`? Can we get away with not having this check here?

dcousens:
> @dcousens please make https://github.com/dcousens/wif public. Currently it's returning 404

I moved it to https://github.com/bitcoinjs/wif

src/ecpair.js
dcousens: Opened an issue in https://github.com/bitcoinjs/wif/issues/1.
Would require a sem-ver bump,  doesn't affect the validity of this PR, so,  will move discussion there.

dcousens:
Addressed nits.  @weilu please merge if ACK

PR #447, opened 2015-08-21
ECPair: fix modulo bias in makeRandom

dcousens:
See https://stackoverflow.com/questions/10984974/why-do-people-say-there-is-modulo-bias-when-using-a-random-number-generator for a good explanation.

`secp256k1.n` is approximately `2^256 - 2^128`, so the bias is negligible,  but it does exist.
No security notice needed.

Will back port.

### DIFF
--- a/src/ecpair.js
+++ b/src/ecpair.js
@@ -105,11 +105,14 @@ ECPair.makeRandom = function (options) {
   options = options || {}
 
   var rng = options.rng || randomBytes
-  var buffer = rng(32)
-  typeforce(types.Buffer256bit, buffer)
 
-  var d = BigInteger.fromBuffer(buffer)
-  d = d.mod(secp256k1.n)
+  var d
+  do {
+    var buffer = rng(32)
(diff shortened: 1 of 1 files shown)


dcousens:
ping @jprichardson, @weilu, @rubensayshi

dcousens:
Merged as 252336a,  changed `(d.compareTo(secp256k1.n) > 0)` to `(d.compareTo(secp256k1.n) >= 0)`

dcousens:
Applied to `1.5.x` as https://github.com/bitcoinjs/bitcoinjs-lib/commit/b89c82b9a651f11fe69a5d9a4f1e40bc72af2941

weilu:
@dcousens you still need to fix `>=` in master.

dcousens:
@weilu catch.  We need to put that under test.
Applied fix in https://github.com/bitcoinjs/bitcoinjs-lib/commit/96c04d0787d65bf846d9a4b70d194386cbb9be11.

PR #448, opened 2015-08-21
Extract BIP66 module

dcousens:
**Only breaking change in terms of exact exception messages**.

The messages are now consistent in the https://github.com/bitcoinjs/bip66 module,  and adhering to SEMVER.

The fact I can now test for all of them using:
``` javascript
/Expected DER (integer|sequence)|(R|S) value (excessively padded|is negative)|(R|S|DER sequence) length is (zero|too short|too long|invalid)/
```

Is a huge bonus compared to the huge regex that was necessary prior.

### DIFF
--- a/src/script.js
+++ b/src/script.js
@@ -134,8 +134,16 @@ function isCanonicalPubKey (buffer) {
 
 function isCanonicalSignature (buffer) {
   if (!Buffer.isBuffer(buffer)) return false
+  if (!isDefinedHashType(buffer[buffer.length - 1])) return false
 
   return bip66.check(buffer.slice(0, -1))
+}
+
+function isDefinedHashType (hashType) {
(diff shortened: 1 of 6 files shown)


src/script.js
```diff
@@ -118,34 +118,31 @@ function decompile (buffer) {
...
+  switch (buffer[0]) {
+    case 0x02:
+    case 0x03:
+      return buffer.length === 33
+    case 0x04:
+      return buffer.length === 65
   }
 
-  return true
+  return false
```
dcousens: Again,  huge performance improvement here.   Added tests to cover it,  overall script tests running in half the time,  isolated benchmark gives >200% improvement.

dcousens:
@weilu please post-ACK

weilu:
LGTM

PR #450, opened 2015-08-25
types: restore quacking behaviour

dcousens:
@weilu right as always.
Found this is still an issue while doing some local testing.

Can't wait to be rid of these types.


dcousens:
Hopefully only a temporary resolution until native types are used solely.

weilu:
ACK

PR #451, opened 2015-08-25
ecdsa: rm unnecessary mod n

dcousens:


dcousens:
Nevermind, we need more tests.

PR #453, opened 2015-09-05
standard: remove padded Blocks

dcousens:
Reference: https://github.com/feross/standard/issues/170

This was re-enabled in `5.0.0`,  its just that eslint wasn't picking it up properly.


weilu:
ACK

PR #454, opened 2015-09-05
use blanket over coveralls

dcousens:
Coveralls has just been a massive pain lately, always throwing false positives.
Using `blanket` means we don't allow our test coverage to drop below 99%,  **ever**.

You can still generate great coverage reports using `coverage-local`.


dcousens:
Please POST-ACK.  This project needs a release.

weilu:
ACK

PR #455, opened 2015-09-05
Cleanup and stop exposing HDNode constants

dcousens:
https://github.com/bitcoinjs/bitcoinjs-lib/commit/4b825bf2a84a7cfad979251c4cf6572e5c6c50b1 is a breaking change,  but,  it had no merits to begin with IMHO.
If it doesn't have consensus, the constants can easily be re-added in `2.x.y`.

### DIFF
--- a/src/address.js
+++ b/src/address.js
@@ -9,7 +9,7 @@ function fromBase58Check (address) {
   if (payload.length < 21) throw new TypeError(address + ' is too short')
   if (payload.length > 21) throw new TypeError(address + ' is too long')
 
-  var version = payload.readUInt8(0)
+  var version = payload[0]
   var hash = payload.slice(1)
 
   return { hash: hash, version: version }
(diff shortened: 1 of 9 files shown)


src/address.js
```diff
@@ -9,7 +9,7 @@ function fromBase58Check (address) {
   if (payload.length < 21) throw new TypeError(address + ' is too short')
   if (payload.length > 21) throw new TypeError(address + ' is too long')
 
-  var version = payload.readUInt8(0)
```
dcousens: These functions aren't necessary if we create `payload` our selves.

src/ecdsa.js
```diff
@@ -243,5 +242,8 @@ module.exports = {
   deterministicGenerateK: deterministicGenerateK,
   recoverPubKey: recoverPubKey,
   sign: sign,
-  verify: verify
+  verify: verify,
+
+  // TODO: remove
+  __curve: secp256k1
```
dcousens: Necessary to avoid strange issues with `infinity` due to no de-duplication occurring by npm.

src/hdnode.js
```diff
@@ -23,17 +26,13 @@ function HDNode (keyPair, chainCode) {
   this.parentFingerprint = 0x00000000
 }
 
-HDNode.MASTER_SECRET = new Buffer('Bitcoin seed')
-HDNode.HIGHEST_BIT = 0x80000000
-HDNode.LENGTH = 78
```
dcousens: There is no reason to expose these.  Breaking change though.

src/transaction.js
```diff
@@ -52,23 +52,12 @@ Transaction.fromBuffer = function (buffer, __noStrict) {
 
   var vinLen = readVarInt()
   for (var i = 0; i < vinLen; ++i) {
-    var hash = readSlice(32)
-
-    if (Transaction.isCoinbaseHash(hash)) {
```
dcousens: Since `script` is now a buffer,  we don't need to treat it differently for generation scripts.

dcousens:
Please POST-ACK.  This project needs a release so I'm pushing a bit more aggressively.

dcousens:
This change was reverted in https://github.com/bitcoinjs/bitcoinjs-lib/pull/463, removing the breaking change tag.

PR #456, opened 2015-09-08
Remove message module

dcousens:
As discussed @weilu.

Can now be found (in isolation) [here](https://github.com/bitcoinjs/bitcoinjs-message).

PR #459, opened 2015-09-08
Remove explicit DER/Compact decoding

dcousens:
None of `ECSignature`, except `parseScriptSignature`, is relevant to bitcoin or this libraries intended use.

**Ideally** https://github.com/bitcoinjs/bitcoinjs-lib/issues/394 might allow us to expose these functions for general purpose in the underlying ECC library.
Since we don't expose raw `ecdsa`, that would be the best place for it.

It is basically a pointless abstraction for this library.

IMHO.

Chained on https://github.com/bitcoinjs/bitcoinjs-lib/pull/456


dcousens:
Rebased and addressed https://github.com/bitcoinjs/bitcoinjs-lib/pull/459#commitcomment-14945676

jprichardson:
Funny, I just updated https://github.com/cryptocoinjs/ecdsa to track bitcoinjs-lib/ecdsa. I suppose these methods can rest in there for the time being.

dcousens:
Rebased,  obviously still `3.0.0` though

src/transaction_builder.js
```diff
@@ -308,18 +307,18 @@ function buildFromInputData (input, scriptType, parentType, redeemScript, allowI
...
       scriptSig = bscript.pubKeyInput(pkSignature)
       break
 
     case 'multisig':
       var msSignatures = input.signatures.map(function (signature) {
-        return signature && signature.toScriptSignature(input.hashType)
+        return signature && bscript.signature.encode(signature, input.hashType)
       })
 
       // fill in blanks with OP_0
```
dcousens: This file outlines the real differences that users will feel.

IMHO it is an improvement,  but @jprichardson I'd be up to hear your opinion.
Less magic,  clear parameters and a functional style.

The only thing I don't like is that `.decode(signature).signature` is a thing.
Perhaps
``` js
{
 r: "...",
 s: "...",
 hashType: 0
}
```

Would be better

jprichardson:
Given that signatures need to be serialized, parsed, and validated, I'm not so sure I'd be quick to call the abstraction useless - yes, maybe this implementation, but an implementation, probably not. This seems like would be very low-priority. I'd be in favor of ditching it for 3.0 - but I think 3.0 should just focus on the move to secp2561k and other pending breaking changes.

dcousens:
> This seems like would be very low-priority. 

Indeed my thoughts.   However,  this could make the movement to `secp256k1` easier due to less specialised / non-generic interfaces?

dcousens:
This is ultimately part of https://github.com/bitcoinjs/bitcoinjs-lib/issues/407.
With https://github.com/bitcoinjs/bitcoinjs-lib/pull/456 merged,  and `bip66` being a separate module... there is literally no reason to have `toDER` and `toCompact`.

This finalises that move.
This is hopefully the last PR before `3.0.0`.

Ping @jprichardson @rubensayshi @afk11

dcousens:
Known breaks for code in the wild (based on a GitHub search):

https://github.com/sabakaio/mnemoauth-client/blob/a46202351a976e4346277ddf9827661a29ca30be/index.js#L19

```
var signature = authKey.sign(messageHash).toDER();
```

would need to become
```
var signature = authKey.sign(messageHash)
var encoded = bip66.encode(signature.r, signature.s)
```

Manual unpacking... yay? :confused: 

https://github.com/darkwallet/darkwallet/blob/4c51dacacb8541305f1442ef7fa1628eb7c19a79/src/js/dwutil/multisig.js#L202

--- Very weird usage there.

https://github.com/BitGo/BitGoJS/blob/5a06433c4c424e5f7bebe1ef2895b857622751b6/src/bitcoin.js#L114

Their using `secp256k1` for performance,  and then double serialising? Interesting... would be a `bip66.decode(...)`

dcousens:
If no consensus is reached within 24 hours,  I'm going to bump this to `4.0.0` and start the `3.0.0` release process.

rubensayshi:
:rocket: 

the BitGo case would actually just become `return secp256k1.signatureExport(sig)` right, they overload the `ECPair.sign` and then go from Buffer -> Signature, but now they can just return that buffer.

the others are odd use cases for which the `bip66` package is perfect.

dcousens:
@rubensayshi for PR to be complete in that respect,  just realised it needs to serialize `r` in `ecdsa.sign` and deserialize in `ecdsa.verify`

Maybe we should just return a `bip66` signature there instead?
Save everyone of that hassle?

dcousens:
Bumped to `4.0.0`...

dcousens:
Rebased

src/script_signature.js
```diff
@@ -0,0 +1,51 @@
...
+  return Buffer.concat([
+    bip66.encode(r, s),
+    hashTypeBuffer
+  ])
+}
+
+module.exports = {
+  fromRSBuffer,
+  decode: decode,
+  encode: encode
```
dcousens: For `4.0.0`,  the `signature` parameter will probably become a 64-byte RS `Buffer`,  for direct compatability with `secp256k1` libraries.

This implementation accepts the `{ r, s }` tuple for now as it is a simpler transition path,  even if it never gets published.

PR #460, opened 2015-09-08
Modulo tests

dcousens:
Adds a [failing] test for checking the RNG output for `ECPair.makeRandom`, see https://github.com/bitcoinjs/bitcoinjs-lib/pull/447#issuecomment-133856729

The test is failing only because, prior to this PR, `makeRandom` was accepting of a RNG value of `0`.
The `ECPair` constructor threw in this case,  (`Private key must be >0`),  so it is not an issue,  but may be unexpected.

Given the astronomically small possibility of this case,  it is almost certainly an error,  so I'm not even sure if looping is the right action.

However, to be consistent, I've fixed it for now.
Thoughts @weilu ?

### DIFF
--- a/test/ecpair.js
+++ b/test/ecpair.js
@@ -169,6 +169,16 @@ describe('ECPair', function () {
 
       assert.strictEqual(keyPair.toWIF(), exWIF)
     })
+
+    it('loops until d is within interval [1, n - 1]', sinon.test(function () {
+      var rng = this.mock()
+      rng.exactly(3)
+      rng.onCall(0).returns(new BigInteger('0').toBuffer(32)) // < 1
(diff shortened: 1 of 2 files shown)


test/ecpair.js
```diff
@@ -160,18 +160,6 @@ describe('ECPair', function () {
         var keyPair = ProxiedECPair.makeRandom()
         assert.strictEqual(keyPair.toWIF(), exWIF)
       })
-
-      it('passes the options param', sinon.test(function () {
```
dcousens: Removed this test,  as https://github.com/cjohansen/Sinon.JS/issues/613 doesn't look like it'll be fixed any time soon.
We already check that the parameters themselves are being used,  it would just have been an added precaution to verify the call behaviour.

PR #462, opened 2015-09-08
travis: no Node 0.11

dcousens:
I'm not sure if this travis target is worth it,  it delays status checks and seems mostly unnecessary.


dcousens:
Node `0.11` is considered unstable for all intents and purposes.  If anyone is using this in production,  please speak up here and I'll revert this change.

jprichardson:
ACK, this makes sense.

PR #466, opened 2015-09-12
ECPair (tests) move throwing constructor tests to fixtures

dcousens:


test/ecpair.js
```diff
@@ -44,31 +44,6 @@ describe('ECPair', function () {
...
-      assert.throws(function () {
-        new ECPair(BigInteger.ONE, Q)
-      }, /Unexpected publicKey parameter/)
-    })
-
-    it('throws if network is missing pubKeyHash constants', function () {
-      assert.throws(function () {
-        new ECPair(null, null, {
-          network: {}
-        }, /Unknown pubKeyHash constants for network/)
```
dcousens: Funny enough,  this wasn't even being tested.  The regex is in the constructor call by mistake.

test/ecpair.js
```diff
@@ -44,31 +44,6 @@ describe('ECPair', function () {
       assert.strictEqual(keyPair.network, NETWORKS.testnet)
     })
 
-    it('throws if compressed option is not a bool', function () {
-      assert.throws(function () {
-        new ECPair(null, null, {
-          compressed: 2
-        }, /Expected Boolean, got 2/)
```
dcousens: Funny enough,  this wasn't even being tested.  The regex is in the constructor call by mistake.

test/fixtures/ecpair.json
```diff
@@ -82,7 +82,27 @@
       {
         "exception": "Private key must be less than the curve order",
         "d": "115792089237316195423570985008687907853269984665640564039457584007913129639935"
+      },
+      {
+        "exception": "Expected property \"compressed\" of type \\?Boolean, got Number 2",
```
dcousens: Because this was previously not under test,  turns out the `Error` message was completely wrong.
This is what was being emitted.

test/fixtures/ecpair.json
```diff
@@ -82,7 +82,27 @@
...
+          "compressed": 2
+        }
+      },
+      {
+        "exception": "Unexpected publicKey parameter",
+        "d": "1",
+        "Q": "0279be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798"
+      },
+      {
+        "exception": "Expected property \"network.messagePrefix\" of type Buffer|String, got undefined",
```
dcousens: I wonder if we should simplify the `Network` type per its use case each time.
That is,  only require say `pubKeyHash` and `wif` constants for `ECPair` etc.

PR #473, opened 2015-09-20
travis: add 4.0.0 and io.js targets

dcousens:


dcousens:
Actually, I think it might be `4.0` as mentioned by @fanatid,  will see how travis goes anyway.

dcousens:
Does SEMVER apply to these version numbers?

dcousens:
Also, standard probably doesn't need to be run on all 4 versions.

PR #474, opened 2015-09-21
HDNode: add sign/verify

dcousens:
This should stop all usage of `hdNode.keyPair` in API usage, meaning we should be able to isolate `HDNode` for `3.0.0`.

Non breaking changes.
Bumps minor.

Note: adds the ability to do `txBuilder.sign(vout, hdNode)`


dcousens:
Will be merging within 24 hours if no issues are raised.

PR #476, opened 2015-09-25
use buffer-equals / buffer-reverse

dcousens:
Cons: 
* Added dependencies (`buffer-reverse`, `buffer-equals`)

Pros
* Actually tested (in the case of `buffer-equals`)
* Falls back to `Buffer.prototype.equals`,  inherent performance gain
* Less code to maintain
* Type checked

### DIFF
--- a/src/bufferutils.js
+++ b/src/bufferutils.js
@@ -167,19 +167,13 @@ function varIntBuffer (i) {
   return buffer
 }
 
-function reverse (buffer) {
-  var buffer2 = new Buffer(buffer)
-  Array.prototype.reverse.call(buffer2)
-  return buffer2
-}
-
 module.exports = {
   equal: require('buffer-equals'),
   pushDataSize: pushDataSize,
   readPushDataInt: readPushDataInt,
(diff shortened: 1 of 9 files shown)


src/bufferutils.js
```diff
@@ -184,7 +174,7 @@ function reverse (buffer) {
 }
 
 module.exports = {
-  equal: equal,
+  equal: require('buffer-equals'),
```
dcousens: Kept only to avoid a breaking change

dcousens:
POST-ACKs welcome

jprichardson:
ACK on `buffer-equals`. When we drop support for Node v0.10 we could use `Buffer.compare` though.

Not sure of the point of `buffer-reverse` though. Why not just `[].reverse.call(buffer)`?

dcousens:
For the same reason we had a separate utility function for it in the first place.
We wanted to do a copy and reverse (immutability),  in *some* cases.

dcousens:
Upon further review,  I don't think its necessary any more.

PR #478, opened 2015-09-27
remove use of buffer-reverse

dcousens:


test/integration/crypto.js
```diff
@@ -135,7 +134,7 @@ describe('bitcoinjs-lib (crypto)', function () {
 
         assert(bitcoin.script.isPubKeyHashInput(scriptChunks), 'Expected pubKeyHash script')
 
-        var prevOutTxId = bufferReverse(transaction.ins[input.vout].hash).toString('hex')
+        var prevOutTxId = Array.prototype.reverse.call(new Buffer(transaction.ins[input.vout].hash)).toString('hex')
```
dcousens: Only real case where it gets messy, and it was in tests anyway.

dcousens:
I guess [at the time](https://github.com/bitcoinjs/bitcoinjs-lib/commit/71d4c78b884212edf9c356c0b736b19d2fde9c2e) it seemed like a good idea?

Issue #480, opened 2015-09-29
Minimal OP_PUSHDATA in scripts


fanatid:
script.isScriptHashOutput has next bug:
```js
bjs = require('./src')
buf = new Buffer('a94c14c286a1af0947f58d1ad787385b1c2c4a976f9e7187', 'hex')
console.log(bjs.script.isScriptHashOutput(buf)) // true
```
bitcoind (decodescript):
```json
{
"asm" : "OP_HASH160 c286a1af0947f58d1ad787385b1c2c4a976f9e71 OP_EQUAL",
"type" : "nonstandard",
"p2sh" : "2ND6G91xAxiq3hWHUokremQJBuhyQKTVxWj"
}
```

function should be like something next:
```js
function isScriptHashOutput (buffer) {
  typeforce(types.Buffer, buffer)

  return buffer.length === 23 &&
    buffer[0] === OPS.OP_HASH160 &&
    buffer[1] === 0x14 &&
    buffer[22] === OPS.OP_EQUAL
}
```

small notes in [bitcoin code](https://github.com/bitcoin/bitcoin/blob/2b0567b00291205731cbb910fb3cf33359a046ee/src/script/standard.cpp#L61)

dcousens:
Not technically a bug,  its just that the core client has become more strict about excessive `OP_PUSHDATA1`.

I might make `isScriptHashOutput` have a `nonStrict`/`strict` parameter to make sure the script adheres to standard network rules...

dcousens:
Actually,  since https://github.com/bitcoin/bips/blob/master/bip-0062.mediawiki was soft-forked,  I guess it is a bug now.
Will release a patch version rather than minor.

dcousens:
Nevermind,  this is still just a matter of `isStandard` as can be seen in [your reference](https://github.com/bitcoin/bitcoin/blob/2b0567b00291205731cbb910fb3cf33359a046ee/src/script/standard.cpp#L61).

Not technically a bug,  not sure what to do with this one.

dcousens:
Please move the discussion [here](https://github.com/bitcoinjs/bitcoinjs-lib/pull/481).
**TL;DR** this is just a discussion of whether `bitcoinjs-lib` should enforce `IsStandard` rules on the `is*` script functions...

PR #481, opened 2015-09-29
isStandard compliance [for pubKeyHash / scriptHash outputs]

dcousens:
As pointed out [here](https://github.com/bitcoinjs/bitcoinjs-lib/issues/480),  this is an interesting issue.

`script.decompile` will allow for non-minimal `OP_PUSHDATA`s (that is,  you can specify a larger `OP_PUSHDATAx` than needed),  meaning extra data can be added into the script.

In some cases (as pointed out in https://github.com/bitcoinjs/bitcoinjs-lib/issues/480),  this can be confusing because [bitcoin/bitcoin](https://github.com/bitcoin/bitcoin/blob/2b0567b00291205731cbb910fb3cf33359a046ee/src/script/standard.cpp#L61) doesn't consider these transactions standard.

So far, bitcoinjs-lib hasn't cared either way, and has allowed the construction of non-standard transactions.

Perhaps an `IsStandard` flag that is enabled by default on these scripts?
Thoughts?

**Or,  better yet,  perhaps our `is*` functions are assumed to be adhering to `IsStandard`, so naturally all we have to do is amend the functions as per this PR?**

@fanatid curious on your opinion here.

### DIFF
--- a/src/address.js
+++ b/src/address.js
@@ -18,11 +18,10 @@ function fromBase58Check (address) {
 function fromOutputScript (scriptPubKey, network) {
   network = network || networks.bitcoin
 
-  var chunks = bscript.decompile(scriptPubKey)
-  if (bscript.isPubKeyHashOutput(chunks)) return toBase58Check(chunks[2], network.pubKeyHash)
-  if (bscript.isScriptHashOutput(chunks)) return toBase58Check(chunks[1], network.scriptHash)
+  if (bscript.isPubKeyHashOutput(scriptPubKey)) return toBase58Check(scriptPubKey.slice(3, 23), network.pubKeyHash)
+  if (bscript.isScriptHashOutput(scriptPubKey)) return toBase58Check(scriptPubKey.slice(2, 22), network.scriptHash)
 
-  throw new Error(bscript.toASM(chunks) + ' has no matching Address')
+  throw new Error(bscript.toASM(scriptPubKey) + ' has no matching Address')
 }
 
 function toBase58Check (hash, version) {
--- a/src/script.js
+++ b/src/script.js
@@ -156,15 +156,14 @@ function isPubKeyHashInput (script) {
 }
 
 function isPubKeyHashOutput (script) {
-  var chunks = decompile(script)
+  var buffer = compile(script)
 
-  return chunks.length === 5 &&
-    chunks[0] === OPS.OP_DUP &&
-    chunks[1] === OPS.OP_HASH160 &&
-    Buffer.isBuffer(chunks[2]) &&
-    chunks[2].length === 20 &&
-    chunks[3] === OPS.OP_EQUALVERIFY &&
-    chunks[4] === OPS.OP_CHECKSIG
+  return buffer.length === 25 &&
+    buffer[0] === OPS.OP_DUP &&
+    buffer[1] === OPS.OP_HASH160 &&
+    buffer[2] === 0x14 &&
+    buffer[23] === OPS.OP_EQUALVERIFY &&
+    buffer[24] === OPS.OP_CHECKSIG
 }
 
 function isPubKeyInput (script) {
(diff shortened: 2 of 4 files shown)


dcousens:
But then again,  that would mean `isScriptHashInput` is inconsistent, because right now it doesn't care about standard rules and checks generically.

fanatid:
Not sure about this, but I think better way will be add `isStandard` (or `strict`) argument to `is*` functions and to `docompile` (when `isStandard` is `true` `decompile` following to minimal OP_PUSHDATA compliance).
Example for `isScriptHashOutput`:
```js
function isScriptHashOutput (script, isStandard) {
  var chunks = decompile(script, isStandard)
  return chunks.length === 3 &&
    chunks[0] === OPS.OP_HASH160 &&
    Buffer.isBuffer(chunks[1]) &&
    chunks[1].length === 20 &&
    chunks[2] === OPS.OP_EQUAL
}
```

In this case it's allowed to check scripts in strict mode and have backward compatibility with old code (`isStandard` should be `false` by default).

dcousens:
Thought about this for a few hours,  and,  it makes sense to make the `is*` functions mostly compliant with `IsStandard` rules simply because the entire concept of a `pay-to-pubkeyhash` and `pay-to-scripthash` is about following the community set standards for those functions.

dcousens:
**TL;DR**:
Considering the `is*` functions are used to match scripts against community set script templates,  adhering to the bitcoin core `IsStandard` rules makes sense IMHO.

To match otherwise, you should probably use a script interpreter or write your own function.

fanatid:
I think that `isCanonicalPubKey` also must need changes... I write some closely code in [script2addresses](https://github.com/fanatid/script2addresses/blob/master/src/publickey.js)

dcousens:
@fanatid what do you mean?  The code for `isCanonicalPubKey` mirrors the code from https://github.com/bitcoin/bitcoin/blob/master/src/script/interpreter.cpp#L67-L87

fanatid:
hm, may be I am wrong.
but in different places you can see different rules:
[size is 33 <= x <= 65 (first byte doesn't important)](https://github.com/bitcoin/bitcoin/blob/0fbfc5106cd9866325b4a1ab3b9db8e91e54f070/src/script/standard.cpp#L129)
[size is 33 (first byte 2 or 3) and 65 (first byte is 4, 6 or 7)](https://github.com/bitcoin/bitcoin/blob/0fbfc5106cd9866325b4a1ab3b9db8e91e54f070/src/pubkey.h#L48) ([called from TransactionSignatureChecker::CheckSig](https://github.com/bitcoin/bitcoin/blob/0fbfc5106cd9866325b4a1ab3b9db8e91e54f070/src/script/interpreter.cpp#L1111))
[and finally in secp256k1 with additional checks](https://github.com/bitcoin/bitcoin/blob/0fbfc5106cd9866325b4a1ab3b9db8e91e54f070/src/secp256k1/src/eckey_impl.h#L17)
I don't think, that is right to take example from `interpreter.cpp` for `isCanonicalPubKey`.

dcousens:
@fanatid remember this isn't a script interpreter, but just a parser.
If you want to verify the public key is a valid ECDSA public key, then you will need to do that yourself (and naturally handle each error case that could consist of).

dcousens:
Going to merge this.
I think the desired result is that the script templates we are checking against are [and were] intended for verifying your own scripts against the `isStandard` equivalents.
In this sense, a `isStandard` flag isn't necessary.

If you want to be *more* liberal in what you check/verify against, it is perfectly simple to write your own.
Therefore, I'm going to merge this on the assumption that it is a patch for previously broken behaviour, not a breaking change.

dcousens:
@fanatid `SCRIPT_VERIFY_STRICTENC` is by default on in bitcoin core, which is what we are matching in terms of these script templates.
Therefore, I think comparing against [interpreter.cpp](https://github.com/bitcoin/bitcoin/blob/master/src/script/interpreter.cpp#L67-L87) is fine.

PR #482, opened 2015-09-30
fix isMultisigOutput allowing m data

dcousens:
Hopefully increases readability for these calculations.

### DIFF
--- a/src/script.js
+++ b/src/script.js
@@ -225,10 +225,12 @@ function isMultisigOutput (script) {
   if (chunks[chunks.length - 1] !== OPS.OP_CHECKMULTISIG) return false
 
   var mOp = chunks[0]
+  if (!types.Number(mOp)) return false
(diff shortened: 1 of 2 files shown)


dcousens:
After reviewing https://github.com/bitcoinjs/bitcoinjs-lib/pull/484,  I noticed this bug.
This PR now addresses that bug.

src/script.js
```diff
@@ -4,11 +4,16 @@ var typeforce = require('typeforce')
...
-  REVERSE_OPS[code] = op
-}
+var REVERSE_OPS = (function () {
+  var result = {}
+  for (var op in OPS) {
+    var code = OPS[op]
+    result[code] = op
+  }
+  return result
+})()
```
dcousens: I did this because `op` was annoyingly in global scope.

weilu:
ACK

Issue #483, opened 2015-10-01
Extra code in isMultisigOutput


fanatid:
Useless checks: [first](https://github.com/bitcoinjs/bitcoinjs-lib/blob/85a399d46b6e1e4db83bac4f7105d51e6e789c46/src/script.js#L228) and [second](https://github.com/bitcoinjs/bitcoinjs-lib/blob/85a399d46b6e1e4db83bac4f7105d51e6e789c46/src/script.js#L233) (both in isMultisigOutput), because next two lines duplicate this conditions.

PR #484, opened 2015-10-01
fix isMultisigOutput

fanatid:
Solve #483 
Change `n < pubKeys.length` to `n !== pubKeys.length` like in [bitcoin](https://github.com/bitcoin/bitcoin/blob/cd78c2a421ae5aa1773c8fb84dadc2c61a01e5e8/src/script/standard.cpp#L102)


dcousens:
I'm not sure what my original reasoning for that was,  but you're right `nOp` should always match the number of chunks until `mOp`.

Thanks @fanatid

PR #492, opened 2015-11-03
travis, build matrix

fanatid:
I added [build matrix](http://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix) to `.travis.yml` but real problem that `npm` run `prepublish` for every build..

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,19 @@
 language: node_js
 before_install:
-  - "npm install npm -g"
+  - npm install npm -g
 node_js:
-  - "4.0.0"
-  - "io.js"
-  - "0.12"
-  - "0.10"
+  - 5
+  - 4
+  - io.js
+  - 0.12
+  - 0.10
+matrix:
+  include:
+    - node_js: 4
+      env: TEST_SUITE=standard
 env:
   - TEST_SUITE=coverage
   - TEST_SUITE=integration
-  - TEST_SUITE=standard
   - TEST_SUITE=unit
-script: "npm run-script $TEST_SUITE"
+script: npm run-script $TEST_SUITE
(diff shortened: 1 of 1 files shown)


fanatid:
faucet error, please restart travis tests

dcousens:
@fanatid what is a build matrix?

fanatid:
possible combinations of builds
before this commit you have 16 builds (4 versions of node_js * 4 value of TEST_SUITE)
with matrix.include and removed 2 values from env it will be 10 builds (4 versions of node_js * 2 value of TEST_SUITE + 2 matrix.include)

dcousens:
I guess it is a reasonable assumption then the `standard` should never change across node versions,  as with the coverage.

dcousens:
@fanatid could you reduce the `build_matrix` to only `standard`?  I think `coverage` is still worthwhile running on all versions.

dcousens:
After that, this is good to merge.

dcousens:
Also,  can you add `5.0.0` to the TRAVIS list of versions to test.

fanatid:
I do this, but why you want specific version 4.0.0 and 5.0.0 not 4 and 5?

dcousens:
@fanatid good question,  are they equivalent?

dcousens:
If `4` is the equivalent of `^4.0.0`, then I'd prefer that

fanatid:
Travis use nvm, so 4 mean last 4.x.x version, I also prefer this

dcousens:
Sounds good :+1:

dcousens:
Looks good, waiting on tests to pass now.

dcousens:
Thanks @fanatid

PR #496, opened 2015-11-05
Bump typeforce to 1.5.5

dcousens:
This version of typeforce uses ES5 solely, and should force any users to update their dependencies if they installed the miscompiled version.


jprichardson:
utACK. Failure looks like an API faucet issue that can be ignored: https://travis-ci.org/bitcoinjs/bitcoinjs-lib/jobs/89523681

dcousens:
rebased and added f840151 to ensure a faucet gives a valid unspent.  If not, the test fails from the timeout.

PR #502, opened 2015-11-19
Remove `const` and replace with `var`

alex-seville:
To keep the project consistenly ES5.

Source: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const

### DIFF
--- a/src/script.js
+++ b/src/script.js
@@ -13,7 +13,7 @@ var REVERSE_OPS = (function () {
   return result
 })()
 
-const OP_INT_BASE = OPS.OP_RESERVED // OP_1 - 1
+var OP_INT_BASE = OPS.OP_RESERVED // OP_1 - 1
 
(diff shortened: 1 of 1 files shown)


dcousens:
Thanks @alex-seville,  beat me to it :+1:

alex-seville:
Checks failed?

dcousens:
Unrelated, restarting them now.  The faucet we use on testnet seems to periodically fail.

alex-seville:
ok, I'll leave it in your capable hands. thanks for using blanket!

dcousens:
@jprichardson I'm temporarily disabling status checks required to merge, as the issue stopping this from passing is external to the library, and I can't actually continue to resolve that without this fix without wasting a lot of time.

jprichardson:
@dcousens seems reasonable enough.

PR #506, opened 2015-11-26
Basic cleanup

dcousens:
Improves inline documentation in `TransactionBuilder` as to why the invalidation logic is why it is.

### DIFF
--- a/src/transaction_builder.js
+++ b/src/transaction_builder.js
@@ -208,13 +208,16 @@ TransactionBuilder.prototype.addInput = function (txHash, vout, sequence, prevOu
     input.prevOutType = prevOutType
   }
 
-  var valid = this.inputs.every(function (input2) {
-    if (input2.hashType === undefined) return true
+  // if signatures exist, adding inputs is only acceptable if SIGHASH_ANYONECANPAY is used
+  // throw if any signatures *didn't* use SIGHASH_ANYONECANPAY
+  if (!this.inputs.every(function (otherInput) {
+    // no signature
+    if (otherInput.hashType === undefined) return true
 
-    return input2.hashType & Transaction.SIGHASH_ANYONECANPAY
-  })
-
-  if (!valid) throw new Error('No, this would invalidate signatures')
+    return otherInput.hashType & Transaction.SIGHASH_ANYONECANPAY
+  })) {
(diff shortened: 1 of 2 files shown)


fanatid:
ACK

src/transaction_builder.js
```diff
@@ -208,13 +208,16 @@ TransactionBuilder.prototype.addInput = function (txHash, vout, sequence, prevOu
...
-  var valid = this.inputs.every(function (input2) {
-    if (input2.hashType === undefined) return true
-
-    return input2.hashType & Transaction.SIGHASH_ANYONECANPAY
-  })
-
-  if (!valid) throw new Error('No, this would invalidate signatures')
+  // if signatures exist, adding inputs is only acceptable if SIGHASH_ANYONECANPAY is used
+  // throw if any signatures *didn't* use SIGHASH_ANYONECANPAY
+  if (!this.inputs.every(function (otherInput) {
```
weilu: I find this syntax awfully hard to read. I personally prefer the previous one with the `valid` var.

src/transaction_builder.js
```diff
@@ -227,23 +230,32 @@ TransactionBuilder.prototype.addInput = function (txHash, vout, sequence, prevOu
 }
 
 TransactionBuilder.prototype.addOutput = function (scriptPubKey, value) {
-  var tx = this.tx
-  var valid = this.inputs.every(function (input, index) {
+  var nOutputs = this.tx.outs.length
+
+  // if signatures exist, adding outputs is only acceptable if SIGHASH_NONE or SIGHASH_SINGLE is used
+  // throws if any signatures didn't use SIGHASH_NONE|SIGHASH_SINGLE
+  if (!this.inputs.every(function (input, index) {
```
weilu: ditto

src/transaction_builder.js
dcousens: Perhaps we can break it out in another way then,  the stateful `var` just seems odd given its 1 time use?

weilu: How is one time use of a var odd?

dcousens: Definition of `var` is variable, and one-time usage isn't variable.
But, this is subjective at best.
I'm not bothered either way.

Issue #437, opened 2015-08-13
TransactionBuilder nLockTime API


dcousens:

dcousens:
Closing in favour of #507

PR #507, opened 2015-11-26
txbuilder: add setLockTime

dcousens:
WIP, needs tests.

Open to feedback, not the biggest fan, but, its simple enough.

Maybe I should use a property instead?

### DIFF
--- a/src/transaction_builder.js
+++ b/src/transaction_builder.js
@@ -4,6 +4,8 @@ var bscript = require('./script')
 var bufferEquals = require('buffer-equals')
 var networks = require('./networks')
 var ops = require('./opcodes')
+var typeforce = require('typeforce')
+var types = require('./types')
 
 var ECPair = require('./ecpair')
 var ECSignature = require('./ecsignature')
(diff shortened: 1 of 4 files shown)


dcousens:
Resolves https://github.com/bitcoinjs/bitcoinjs-lib/issues/437

dcousens:
ping @jprichardson thoughts on API?

jprichardson:
`setLockTime()` is fine. 

An aside, builders are commonly used in conjunction with a fluent interface. Since it doesn't, what's the rationalization for splitting up `Transaction` / `TransactionBuilder`? (Really doesn't have to be answered here)

dcousens:
`Transaction` is just the data structure,  no state,  capable of being manually mutated and it will still serialize correctly.
To which, yes, it probably does make sense to remove `addOutput` on it and just have `transaction.outputs.push`  etc.

dcousens:
> An aside, builders are commonly used in conjunction with a fluent interface. 

Any ideas on how we could improve this interface?

src/transaction_builder.js
```diff
@@ -135,6 +137,21 @@ function TransactionBuilder (network) {
...
+TransactionBuilder.setLockTime = function (locktime) {
+  typeforce(types.Uint32, locktime)
+
+  // if any signatures exist, throw
+  if (this.inputs.some(function (input) {
+    if (!input.signatures) return false
+
+    return input.signatures.some(function (s) { return s })
+  })) {
+    throw new Error('No, this would invalidate signatures')
```
weilu: I still find this error message funny because it answers "no" without any question asked :P

weilu:
The API looks ok to me.

src/transaction_builder.js
dcousens: Haha, I know.  But, it works.
If you can think of something for `3.0.0`, go for it :)

dcousens:
@jprichardson not going to merge this without status checks as tests have been added.
Please verify manually and merge when ready. :+1:

dcousens:
Needed https://github.com/bitcoinjs/bitcoinjs-lib/pull/524,  so merged manually after local verification.
@jprichardson please review, verify and merge.

jprichardson:
:+1:

PR #509, opened 2015-12-08
Add Block proof-of-work validation function

dcousens:
Might be a way to do this nicely without resorting to using `bigi`, but, for now, here it is.
Should probably get some more test fixtures from core if possible.

* Adds `Block.calculateTarget` (reference: https://en.bitcoin.it/wiki/Target)
* Adds `Block.prototype.verifyPow`


test/block.js
```diff
@@ -87,4 +87,18 @@ describe('Block', function () {
...
+  describe('verifyPow', function () {
+    fixtures.valid.forEach(function (f) {
+      var block
+
+      beforeEach(function () {
+        block = Block.fromHex(f.hex)
+      })
+
+      it('returns ' + f.valid + ' for ' + f.id, function () {
+        assert.strictEqual(block.verifyPow(), f.valid)
```
dcousens: @jprichardson @weilu any thoughts on `verifyPow` as a name?

src/block.js
```diff
@@ -115,4 +116,14 @@ Block.prototype.toHex = function (headersOnly) {
   return this.toBuffer(headersOnly).toString('hex')
 }
 
+Block.prototype.verifyPow = function () {
+  var hash = BigInteger.fromBuffer([].reverse.call(this.getHash()))
+  var mov = ((this.bits >>> 24) - 3) << 3
+  var target = new BigInteger()
+  target.fromInt(this.bits & 0xffffff)
```
dcousens: `0xffffff000` might be more clear, seeing as it is a 32 bit number?

dcousens:
Removed the need for `bigi` by just using a lexicographical comparison.
This could be further optimized by avoiding the allocation of a `Buffer` or storing `difficulty` after creation.

dcousens:
Needs more tests, not happy with this.  I think its wrong.

dcousens:
Yeah, totally wrong, will re-work it (ref https://en.bitcoin.it/wiki/Target)

dcousens:
Mostly good to go now :+1: , just need a few more fixtures to verify that `calculateTarget` is implemented correctly.

src/block.js
```diff
@@ -115,4 +116,25 @@ Block.prototype.toHex = function (headersOnly) {
...
+  var target = new Buffer(32)
+  target.fill(0)
+
+  var exponent = ((bits & 0xff000000) >> 24) - 3
+  var value = bits & 0x00ffffff
+  var i = 31 - exponent
+  if (value & 0xff) target[i--] = (value & 0xff)
+  if (value >>> 8) target[i--] = (value >>> 8)
+  if (value >>> 16) target[i--] = (value >>> 16)
+  if (value >>> 24) target[i--] = (value >>> 24)
```
dcousens: This zero trimming is the only part I'm still not sure why it happens.  Need to verify it with more fixtures.

src/block.js
```diff
@@ -115,4 +116,25 @@ Block.prototype.toHex = function (headersOnly) {
   return this.toBuffer(headersOnly).toString('hex')
 }
 
+Block.calculateTarget = function (bits) {
+  var target = new Buffer(32)
+  target.fill(0)
+
+  var exponent = ((bits & 0xff000000) >> 24) - 3
+  var value = bits & 0x00ffffff
```
dcousens: Its unclear if this should be `bits & 0x007fffff` or `0x00ffffff`.
I'm happy with this as is for now, will wait to enforce that it verifies for the mainnet blockchain before merge.

src/block.js
```diff
@@ -115,4 +116,21 @@ Block.prototype.toHex = function (headersOnly) {
...
+var ZEROS = '0000000000000000000000000000000000000000000000000000000000000000'
+
+Block.calculateTarget = function (bits) {
+  var exponent = ((bits & 0xff000000) >> 24) - 3
+  var value = bits & 0x00ffffff
+  var target = value * Math.pow(2, 8 * exponent)
+  var targetHex = target.toString(16)
+
+  return new Buffer(ZEROS.slice(0, 64 - targetHex.length) + targetHex, 'hex')
+}
```
dcousens: This is way clearer now

dcousens:
Matched against every possible target in the `uint32_t` space in terms of bitcoin/bitcoin consensus.
So,  good to go :+1:

dcousens:
@jprichardson guessing you were OK with this API?  If not,  please comment within 24 hours :)

jprichardson:
> @jprichardson guessing you were OK with this API? If not, please comment within 24 hours :)

Yep :) ACK.

PR #511, opened 2015-12-17
add OP_CHECKLOCKTIMEVERIFY and integration test example

dcousens:


dcousens:
@jprichardson please review/merge :+1:

test/integration/advanced.js
```diff
@@ -51,11 +51,65 @@ describe('bitcoinjs-lib (advanced)', function () {
...
+      var tx = new bitcoin.TransactionBuilder(network)
+
+      // now + 1 month
+      var hodlDate = Math.floor((Date.now() + new Date(0).setMonth(1)) / 1000)
+      var hodlLockTimeBuffer = new Buffer(4)
+      hodlLockTimeBuffer.writeInt32LE(hodlDate | 0, 0)
+
+      // {signature} {signature} or
+      // OP_0 {signature} after 1 month
+      var hodlScript = bitcoin.script.compile([
```
dcousens: Its an arbitrary example, sure, 1 signature or 2 signatures, but, it works, and its clear how to extend it IMHO.

Until we have something better? :)

src/opcodes.json
```diff
@@ -113,6 +113,9 @@
 
   "OP_NOP1": 176,
   "OP_NOP2": 177,
+  "OP_CHECKLOCKTIMEVERIFY": 177,
+  "OP_HODL": 177,
```
dcousens: @weilu / @jprichardson its a common synonym, haha.  Thoughts?

dcousens:
Might make the lock time shorter such that I can verify this script as an example before merging.

dcousens:
In which case, might also have to change the test to make the private key more deterministic.

src/opcodes.json
jprichardson: Haha. I'd kill it though. Just one more thing that we'd have to maintain.

jprichardson:
utACK. I'd remove `OP_HODL`. Also,

> Might make the lock time shorter such that I can verify this script as an example before merging.

This would be awesome.

src/opcodes.json
dcousens: Good point.  Will remove.

dcousens:
@jprichardson how about a convenience wrapper that does the equivalent of:

``` javascript
function cltvAlternatorScriptPubKey (scriptA, lockTime, scriptB, scriptFinal) {
  var lockTimeBuffer = new Buffer(4)
  lockTimeBuffer.writeInt32LE(lockTime | 0, 0)

  let aChunks = decompile(scriptA)
  let bChunks = decompile(scriptB)
  let finalChunks = scriptFinal ? decompile(scriptFinal) : []

  return compile([
    opcodes.OP_IF,
    lockTimeBuffer,
    opcodes.OP_CHECKLOCKTIMEVERIFY
  ].concat(
    aChunks,
    opcodes.OP_ELSE,
    bChunks,
    opcodes.OP_ENDIF,
    finalChunks
  )
])
```

dcousens:
Then the script would just be:

``` javascript
var hodlScript = cltvAlternatorScriptPubKey([bitcoin.opcodes.OP_DROP], hodlDate, [keyPair.getPublicKeyBuffer(), bitcoin.opcodes.OP_CHECKSIGVERIFY]), [bitcoin.opcodes.OP_CHECKSIG])`

**Comment by dcousens** [comment]
A little confusing, but, this is the default structure for a CLTV script in most cases, and would mean you could do: `cltvAlternatorScriptPubKey(pay2pk(Alice), oneDayLater, pay2pk(Bob), pay2pk(Charles)` which would require a pay2pk signature for Alice and Charles the first day, otherwise Bob and Charles for every day thereafter.

**Comment by dcousens** [comment]
Perhaps we just want `ifElseEndIfSpk(cltv(oneDayLater, pay2pk(Alice)), pay2pk(Bob), pay2pk(Charles))`

That way its more composable, and clearer how to reverse the condition such that the CLTV check is higher in priority.

File: test/integration/advanced.js [hunk 3]
```diff
@@ -46,16 +46,70 @@ describe('bitcoinjs-lib (advanced)', function () {
...
+        bitcoin.opcodes.OP_DROP,
+        bitcoin.opcodes.OP_ELSE,
+        keyPair.getPublicKeyBuffer(),
+        bitcoin.opcodes.OP_CHECKSIGVERIFY,
+        bitcoin.opcodes.OP_ENDIF,
+        keyPair.getPublicKeyBuffer(),
+        bitcoin.opcodes.OP_CHECKSIG
+      ])
+
+      tx.addInput(unspent.txId, unspent.vout)
```
dcousens: This is technically wrong, as the sequence number must not be `0xffffffff`, which it is currently.  Therefore the `OP_CHECKLOCKTIMEVERIFY` is invalid, and the entire script unspendable.

jprichardson:
I like where you're going with this. What if we created a new module, `bip65` that contains these utility functions? 

> Perhaps we just want
>
> `ifElseEndIfSpk(cltv(oneDayLater, pay2pk(Alice)), pay2pk(Bob), pay2pk(Charles))`

> That way its more composable, and possible to reverse the condition such that the CLTV check is higher in priority.

I'm usually in favor of the more functional approach.

dcousens:
> I like where you're going with this. What if we created a new module, bip65 that contains these utility functions?

Sounds like a plan, also means we could add all the examples in the BIP itself too as full-fledged templates.
Will need `bitcoinsjs-lib/opcodes` though. Perhaps for now we'll hard-code it it in and then find a way to extract it nicely?

dcousens:
Merging as is for now since this can't actually be verified as a spend yet thanks to no block explorers supporting that kind of lookup (and we can't track `txId`s or `P2SH` hashes between tests).

dcousens:
Happy  to revisit this if we can come up with something.

dcousens:
Wait, we should be able to query it via `P2SH`, never mind, I'll look into this now.

dcousens:
This example lock time is wrong,  I didn't realise we were completely missing a `CScriptNum` equivalent.

PR #516, opened 2016-01-04
script_number: add impl/tests

dcousens:
@jprichardson in working on the spend example for #511 (which I now regret merging without it), I've realised we didn't have a `CScriptNum` equivalent for script numbers encoded in a stack.

### DIFF
--- a/src/script.js
+++ b/src/script.js
@@ -372,6 +372,8 @@ module.exports = {
   fromASM: fromASM,
   toASM: toASM,
 
+  number: require('./script_number'),
+
   isCanonicalPubKey: isCanonicalPubKey,
   isCanonicalSignature: isCanonicalSignature,
   isDefinedHashType: isDefinedHashType,
--- a/src/script_number.js
+++ b/src/script_number.js
@@ -20,23 +20,11 @@ function decode (buffer, maxLength, minimal) {
 
     if (b & 0x80) return -((b & ~0x80) * 0x100000000 + a)
     return b * 0x100000000 + a
+  }
 
-  // 32-bit
-  } else if (length === 4) {
-    result = buffer.readUInt32LE(0)
-
-  // 24-bit
-  } else if (length === 3) {
-    result = buffer.readUInt16LE(0)
-    result |= buffer.readUInt8(2) << 16
-
-  // 16-bit
-  } else if (length === 2) {
-    result = buffer.readUInt16LE(0)
-
-  // 8-bit
-  } else {
-    result = buffer.readUInt8(0)
+  // 32-bit / 24-bit / 16-bit / 8-bit
+  for (var i = 0; i < length; ++i) {
+    result |= buffer[i] << (8 * i)
   }
 
   if (buffer[length - 1] & 0x80) return -(result & ~(0x80 << (8 * (length - 1))))
--- a/src/script_number.js
+++ b/src/script_number.js
@@ -1,0 +1,78 @@
+function decode (buffer, maxLength, minimal) {
+  maxLength = maxLength || 4
+  minimal = minimal === undefined ? true : minimal
+
+  var length = buffer.length
+  if (length === 0) return 0
+  if (length > maxLength) throw new TypeError('Script number overflow')
+  if (minimal) {
+    if ((buffer[length - 1] & 0x7f) === 0) {
+      if (length <= 1 || (buffer[length - 2] & 0x80) === 0) throw new Error('Non-minimally encoded script number')
+    }
+  }
+
+  // 32-bit?
+  if (length < 5) {
+    var result = 0
+
+    for (var i = 0; i < length; ++i) {
+      result += buffer[i] << (8 * i)
+    }
+
+    if (buffer[length - 1] & 0x80) return -(result & ~(0x80 << (8 * (length - 1))))
+    return result
+  }
+
+  // 40-bit
+  var a = buffer.readUInt32LE(0)
+  var b = buffer.readUInt8(4)
+
+  // TODO: refactor
+  var neg = false
+  if (b & 0x80) {
+    b &= ~0x80
+    neg = true
+  }
+
+  var r = b * 0x100000000 + a
+
+  if (neg) {
+    r = -r
+  }
+
+  return r
+}
+
+function scriptNumSize (i) {
+  return i > 0x7fffffff ? 5
+  : i > 0x7fffff ? 4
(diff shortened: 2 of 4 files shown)


test/fixtures/script_number.json
```diff
@@ -0,0 +1,225 @@
+[
```
dcousens: These fixtures are [generated](https://gist.github.com/dcousens/5352f9488577631afd96) from bitcoin core.
@jprichardson any ideas on how we could increase transparency over this?

A link to a gist that will likely disappear soon isn't great.

src/script_number.js
```diff
@@ -0,0 +1,78 @@
+function decode (buffer, maxLength, minimal) {
+  maxLength = maxLength || 4
+  minimal = minimal === undefined ? true : minimal
+
+  var length = buffer.length
+  if (length === 0) return 0
+  if (length > maxLength) throw new TypeError('Script number overflow')
+  if (minimal) {
+    if ((buffer[length - 1] & 0x7f) === 0) {
+      if (length <= 1 || (buffer[length - 2] & 0x80) === 0) throw new Error('Non-minimally encoded script number')
```
dcousens: Needs tests for this exception

src/script_number.js
```diff
@@ -0,0 +1,78 @@
+function decode (buffer, maxLength, minimal) {
+  maxLength = maxLength || 4
+  minimal = minimal === undefined ? true : minimal
+
+  var length = buffer.length
+  if (length === 0) return 0
+  if (length > maxLength) throw new TypeError('Script number overflow')
```
dcousens: Needs tests for this exception

dcousens:
@jprichardson thoughts on `bitcoin.script.number.[encode/decode]`?  See 29a1a83

src/script_number.js
```diff
@@ -0,0 +1,78 @@
...
+  if (minimal) {
+    if ((buffer[length - 1] & 0x7f) === 0) {
+      if (length <= 1 || (buffer[length - 2] & 0x80) === 0) throw new Error('Non-minimally encoded script number')
+    }
+  }
+
+  var result
+
```
dcousens: I couldn't go with the basic `while` loop construction here simply because we only have 32-bit integers.

src/script_number.js
```diff
@@ -0,0 +1,78 @@
...
+  // 40-bit
+  if (length === 5) {
+    var a = buffer.readUInt32LE(0)
+    var b = buffer.readUInt8(4)
+
+    if (b & 0x80) return -((b & ~0x80) * 0x100000000 + a)
+    return b * 0x100000000 + a
+
+  // 32-bit
+  } else if (length === 4) {
```
dcousens: I could go with the `for` loop construction here, but, it seemed clearer to just use the `read*` functions here.
YMMV.  Thoughts @jprichardson?

src/script_number.js
fanatid: Not sure on 100%, but should work
```js
var result
for (var i = 0; i < length; ++i) {
  result += buffers[i] * Math.pow(2, 8 * i)
}
```

dcousens: Suffers rounding errors, I tried that initially

dcousens:
I'll merge this tomorrow after I add the missing tests for exceptional cases and run a full 1:1 value matching over the `2**32` value space against [`bitcoin/bitcoin`](https://github.com/bitcoin/bitcoin).
Its small enough, so no reason why not.  Hopefully the fixtures already cover all the interesting cases anyway.

src/script_number.js
fanatid: How it's possible? Number have 53 bit precision, there only 40 bits.
Can you paste example that give rounding errors?

dcousens: @fanatid never mind, works fine, will have a look at it again tomorrow.
To avoid the 40-bit special case, I'd need a non-integer equivalent of the negative handling code (aka, `-result` on `0x80`).

dcousens: @fanatid I'm thinking when I was testing it,  there was other issues that gave me the wrong impression.  You are certainly correct that there **should not** have been issues.

src/script_number.js
```diff
@@ -0,0 +1,66 @@
...
+    if (b & 0x80) return -((b & ~0x80) * 0x100000000 + a)
+    return b * 0x100000000 + a
+  }
+
+  // 32-bit / 24-bit / 16-bit / 8-bit
+  for (var i = 0; i < length; ++i) {
+    result |= buffer[i] << (8 * i)
+  }
+
+  if (buffer[length - 1] & 0x80) return -(result & ~(0x80 << (8 * (length - 1))))
```
dcousens: @fanatid if you have any ideas on how to apply this to the 40-bit case,  then I'll go ahead and generalize the entire thing (yay).

I can add you as a contributor if you want to push to this branch :+1:

fanatid: I think it should be fine:
```js
if (buffer[length - 1] & 0x80) return -((result / Math.pow(2, 8 * (length - 1) - 1)) | 0)
```

dcousens: `| 0` will force it to 32-bits again, won't work with 40-bit.
Also, this doesn't work at all,  I'll look over it in depth in a bit, but,  was it meant to be just a drop-in replacement?

dcousens: @fanatid got the last statement with:

``` javascript
  var b = buffer[length - 1]
  if (b & 0x80) {
    result -= b * Math.pow(2, 8 * (length - 1))

    return -((b & ~0x80) * Math.pow(2, 8 * (length - 1)) + result)
  }
```

Not really worth it though, the readability is marginal and the performance is almost 50% worse.
With the current code, the difference between when bit-wise operators are OK and the 40-bit special case is explicit.

src/script.js
```diff
@@ -372,6 +372,8 @@ module.exports = {
   fromASM: fromASM,
   toASM: toASM,
 
+  number: require('./script_number'),
```
dcousens: @jprichardson not going to release this publically till I get an ACK from you on this

src/script_number.js
fanatid: Agree, bitwise is more readable.

jprichardson:
ACK. Nicely done! Any reasons why you want this in the core repo and not a separate package?

dcousens:
@jprichardson not specifically, however it does coincide with the idea of being a core-primitive.
I'm not opposed to simply exposing it in this repository however?

PR #517, opened 2016-01-05
tests/integration: amend lockTime examples

dcousens:
Adds redemption examples too.

I set the `nLockTime` to `now - 3 hours` so as to be able to test.
I'll probably add a test to ensure it fails if time is `now + 3 hours`, later, maybe not in this PR.


dcousens:
@jprichardson sorry for the screw around :fearful:

PR #519, opened 2016-01-05
tests: rm unused fixtures

dcousens:


dcousens:
The changes RE the faucet didn't decrease the amount of issues it has, but, it did make the issues clear to see.

PR #520, opened 2016-01-15
Segregated Witness Support - WIP

rubensayshi:
# WORK IN PROGRESS

getting shit to work before making it pretty; almost done with getting shit to work!

depends on: https://github.com/bitcoinjs/bitcoinjs-lib/pull/540

first time spending a p2wpkh (6a539927494f3d6f07078dd1c70a28e0c4910662593072305529eaa66257f59d) and p2wpkh as p2sh (4704de796b8ea819f262eb3399fad85b838327497639da2ceafbd04a08a06be3) with bitcoinjs-lib :D
### ToDo
- [x] Output script creation for P2WPKH
- [x] Output script creation for P2WSH
- [x] Transaction encoding / decode
- [x] SignatureHash v2
- [x] Transaction signing of P2WPKH
- [x] Transaction signing of P2WPKH as P2SH
- [x] Transaction signing of P2WSH
- [x] `toBuffer` should be able to return both segwit and non-segwit TX
  - [x] fix `getTxId` using the segwit TX atm
- [x] BIP142 addresses for P2WPKH
- [ ] Add `value` arg to `addInput` and use it in `TransactionBuilder.sign` instead of `value` arg there?
- [x] Refactoring of changes made so far
- [ ] More tests
  - [x] Add tests for mix of non witness and witness (already manually tested and works)
  - [ ] Integration tests (and an API to already run them against segnet)
  - [ ] Check bitcoin core projects for any vectors that we can copy
- [ ] Review and discuss changes for BIP142 addresses
- [ ] Review and discuss changes to `TransactionBuilder.sign`
- [ ] Review and discuss changes to `TransactionBuilder.extractInput` / `script.classifyInput`
- [ ] Cache `hashPrevouts`, `hashSequence` and `hashOutputs` for all signatures in a transaction (since that was one of the main reasons for changing the signatureHash xD)
### Usage

**Signing Native P2WPKH**

``` js
transactionBuilder.sign(0, keyPair, null, null, /* segWit = */ true, /* txIn.value = */ parseInt(0.9998 * 1e8)))
```

**Signing P2WPKH nested in P2SH**

``` js
var pubKeyHash = bcrypto.hash160(keyPair.getPublicKeyBuffer())
var redeemScript = bscript.segWitPubKeyHashOutput(pubKeyHash)

transactionBuilder.sign(0, keyPair, redeemScript, null, /* segWit = */ true, /* txIn.value = */ parseInt(0.9998 * 1e8)))
```
### References
- BIP141 (general segwit) https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki
- BIP142 (segwit address) https://github.com/bitcoin/bips/blob/master/bip-0142.mediawiki
- BIP143 (signature verification with new signature hash) https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki
- BIP144 (peer services) https://github.com/bitcoin/bips/blob/master/bip-0144.mediawiki
- Staging branch https://github.com/sipa/bitcoin/tree/segwit

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 bitcoin.js
 coverage
 node_modules
+
+_dev.*.js
--- a/src/address.js
+++ b/src/address.js
@@ -6,13 +6,24 @@ var types = require('./types')
 
 function fromBase58Check (address) {
   var payload = bs58check.decode(address)
-  if (payload.length < 21) throw new TypeError(address + ' is too short')
-  if (payload.length > 21) throw new TypeError(address + ' is too long')
 
   var version = payload[0]
   var hash = payload.slice(1)
+  var segWitVersion
+  var segWitPadding
 
-  return { hash: hash, version: version }
+  if (hash.length === 22 || hash.length === 34) {
+    segWitVersion = hash.readUInt8(0)
+    segWitPadding = hash.readUInt8(1)
+    hash = hash.slice(2)
+
+    if (segWitVersion > 16 || segWitPadding !== 0) {
+      throw new Error(address + ' has the length of a segWit address, but is not a valid segWit address')
+    }
+  } else if (hash.length > 20) throw new Error(address + ' is too long')
+  else if (hash.length < 20) throw new Error(address + ' is too short')
+
+  return { hash: hash, version: version, segWitVersion: segWitVersion }
 }
 
 function fromOutputScript (scriptPubKey, network) {
@@ -20,16 +31,29 @@ function fromOutputScript (scriptPubKey, network) {
 
   if (bscript.isPubKeyHashOutput(scriptPubKey)) return toBase58Check(bscript.compile(scriptPubKey).slice(3, 23), network.pubKeyHash)
   if (bscript.isScriptHashOutput(scriptPubKey)) return toBase58Check(bscript.compile(scriptPubKey).slice(2, 22), network.scriptHash)
+  if (bscript.isSegWitPubKeyHashOutput(scriptPubKey)) return toBase58Check(bscript.compile(scriptPubKey).slice(2, 22), network.segWitPubKeyHash, 0)
+  if (bscript.isSegWitScriptHashOutput(scriptPubKey)) return toBase58Check(bscript.compile(scriptPubKey).slice(2, 34), network.segWitScriptHash, 0)
 
   throw new Error(bscript.toASM(scriptPubKey) + ' has no matching Address')
 }
 
-function toBase58Check (hash, version) {
-  typeforce(types.tuple(types.Hash160bit, types.UInt8), arguments)
+function toBase58Check (hash, version, segWitVersion) {
+  var payload
+  var isSegWit = typeof segWitVersion !== 'undefined'
 
-  var payload = new Buffer(21)
-  payload.writeUInt8(version, 0)
-  hash.copy(payload, 1)
+  typeforce(types.tuple((isSegWit ? types.oneOf(types.Hash160bit, types.Hash256bit) : types.Hash160bit), types.UInt8, types.maybe(types.UInt8)), arguments)
+
+  if (isSegWit) {
+    payload = new Buffer(3 + hash.length) // dynamic size based on hash because of difference between P2WPKH and P2WSH
+    payload.writeUInt8(version, 0)
+    payload.writeUInt8(segWitVersion, 1)
+    payload.writeUInt8(0, 2) // padding byte to make pretty prefixes
+    hash.copy(payload, 3)
+  } else {
+    payload = new Buffer(21)
+    payload.writeUInt8(version, 0)
+    hash.copy(payload, 1)
+  }
 
   return bs58check.encode(payload)
 }
@@ -38,8 +62,23 @@ function toOutputScript (address, network) {
   network = network || networks.bitcoin
 
   var decode = fromBase58Check(address)
+
   if (decode.version === network.pubKeyHash) return bscript.pubKeyHashOutput(decode.hash)
   if (decode.version === network.scriptHash) return bscript.scriptHashOutput(decode.hash)
+  if (decode.version === network.segWitPubKeyHash) {
+    if (decode.segWitVersion === 0) {
+      if (decode.hash.length === 20) {
+        return bscript.segWitPubKeyHashOutput(decode.hash)
+      }
+    }
+  }
+  if (decode.version === network.segWitScriptHash) {
+    if (decode.segWitVersion === 0) {
+      if (decode.hash.length === 32) {
+        return bscript.segWitScriptHashOutput(decode.hash)
+      }
+    }
+  }
 
   throw new Error(address + ' has no matching Script')
 }
--- a/src/bufferwriter.js
+++ b/src/bufferwriter.js
@@ -1,0 +1,55 @@
+var bufferutils = require('./bufferutils')
+var typeforce = require('typeforce')
+var types = require('./types')
+
+var BufferWriter = function (length) {
+  typeforce(types.tuple(types.Number), arguments)
+  this.buffer = new Buffer(length)
+  this.offset = 0
+}
+
+BufferWriter.prototype.writeSlice = function (slice) {
+  slice.copy(this.buffer, this.offset)
+  this.offset += slice.length
+
+  return this
+}
+
+BufferWriter.prototype.writeSliceWithVarInt = function (script) {
+  this.writeVarInt(script.length)
+  this.writeSlice(script)
+
+  return this
+}
+
+BufferWriter.prototype.writeScript = BufferWriter.prototype.writeSliceWithVarInt
+
+BufferWriter.prototype.writeInt = function (i) {
+  this.buffer.writeUInt8(i, this.offset)
+  this.offset += 1
+
+  return this
+}
+
+BufferWriter.prototype.writeUInt64 = function (i) {
+  bufferutils.writeUInt64LE(this.buffer, i, this.offset)
+  this.offset += 8
+
+  return this
+}
+
+BufferWriter.prototype.writeUInt32 = function (i) {
+  this.buffer.writeUInt32LE(i, this.offset)
+  this.offset += 4
+
+  return this
+}
+
+BufferWriter.prototype.writeVarInt = function (i) {
+  var n = bufferutils.writeVarInt(this.buffer, i, this.offset)
+  this.offset += n
+
+  return this
+}
+
+module.exports = BufferWriter
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -4,9 +4,15 @@ var bufferutils = require('./bufferutils')
 var opcodes = require('./opcodes')
 var typeforce = require('typeforce')
 var types = require('./types')
+var BufferWriter = require('./bufferwriter')
+
+var ADVANCED_TRANSACTION_MARKER = 0
+var ADVANCED_TRANSACTION_FLAG = 1
 
 function Transaction () {
   this.version = 1
+  this.marker = null
+  this.flag = null
   this.locktime = 0
   this.ins = []
   this.outs = []
@@ -37,6 +43,13 @@ Transaction.fromBuffer = function (buffer, __noStrict) {
     return i
   }
 
+  function readInt () {
+    var i = buffer.readUInt8(offset)
+    offset += 1
+
+    return i
+  }
+
   function readVarInt () {
     var vi = bufferutils.readVarInt(buffer, offset)
     offset += vi.size
@@ -47,8 +60,23 @@ Transaction.fromBuffer = function (buffer, __noStrict) {
     return readSlice(readVarInt())
   }
 
+  // console.log('fromBuffer')
+
   var tx = new Transaction()
   tx.version = readUInt32()
+
+  tx.marker = readInt()
+  tx.flag = readInt()
+
+  // check if transaction is advanced (segwit) format
+  if (tx.marker === ADVANCED_TRANSACTION_MARKER && tx.flag === ADVANCED_TRANSACTION_FLAG) {
+    // -
+  } else {
+    // undo the reading of the marker and flag byte
+    offset -= 2
+    tx.marker = null
+    tx.flag = null
+  }
 
   var vinLen = readVarInt()
   for (var i = 0; i < vinLen; ++i) {
@@ -56,7 +84,8 @@ Transaction.fromBuffer = function (buffer, __noStrict) {
       hash: readSlice(32),
       index: readUInt32(),
       script: readScript(),
-      sequence: readUInt32()
+      sequence: readUInt32(),
+      witness: undefined
     })
   }
 
@@ -66,6 +95,16 @@ Transaction.fromBuffer = function (buffer, __noStrict) {
       value: readUInt64(),
       script: readScript()
     })
+  }
+
+  if (tx.flag === ADVANCED_TRANSACTION_FLAG) {
+    for (i = 0; i < vinLen; ++i) {
+      tx.ins[i].witness = []
+      var witnessLen = readVarInt()
+      for (var x = 0; x < witnessLen; ++x) {
+        tx.ins[i].witness.push(readScript())
+      }
+    }
   }
 
   tx.locktime = readUInt32()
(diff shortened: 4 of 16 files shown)


src/transaction.js
```diff
@@ -170,9 +223,100 @@ var VALUE_UINT64_MAX = new Buffer('ffffffffffffffff', 'hex')
...
+
+  if (segWit) {
+    return this.hashForSignatureV2(inIndex, prevOutScript, amount, hashType)
+  } else {
+    return this.hashForSignatureV1(inIndex, prevOutScript, hashType)
+  }
+}
+
+Transaction.prototype.hashForSignatureV2 = function (inIndex, prevOutScript, amount, hashType) {
+  var hashPrevouts = new Buffer("00" * 32, 'hex')
```
fanatid: @rubensayshi it isn't python :laughing:

rubensayshi: hehe yea noticed that yday, been switching between projects too much :P

dcousens:
@rubensayshi when is the next time you can be available on slack?   It'd be good to talk over how this [and future changes] will integrate with `TransactionBuilder` and `Transaction`.

rubensayshi:
@dcousens I've got slack on my phone so you can always poke me but I usually prefer keeping my weekends code-free, though this weekend is an exception and working on this PR atm :D
poke me whenever you have time this week, I'm available all week (GMT+1)

src/types.js
```diff
@@ -13,6 +13,7 @@ function Buffer256bit (value) { return nBuffer(value, 32) }
 var UINT53_MAX = Math.pow(2, 53) - 1
 function UInt2 (value) { return (value & 3) === value }
 function UInt8 (value) { return (value & 0xff) === value }
+function UInt16 (value) { return (value >> 0) === value } // @TODO: is this correct? copy paste of UInt32
```
rubensayshi: is this okay to do ?

src/address.js
```diff
@@ -38,8 +63,15 @@ function toOutputScript (address, network) {
   network = network || networks.bitcoin
 
   var decode = fromBase58Check(address)
+
   if (decode.version === network.pubKeyHash) return bscript.pubKeyHashOutput(decode.hash)
   if (decode.version === network.scriptHash) return bscript.scriptHashOutput(decode.hash)
+  if (decode.version === network.segWitPubKeyHash) {
+    if (decode.segWitVersion !== 0 && decode.segWitPadding !== 0) {
+      throw new TypeError(address + ' is too long for normal address and not a valid segWit address')
```
rubensayshi: not sure if this should go here or in `fromBase58Check`, considering the length checks are in `fromBase58Check`, this code has gotten a bit akward now ...

consider one of the test vectors for `fromBase58Check` we had of an address with hash length of 22, it no longer throws a 'is too long' exception and because we're unaware of `network` in `fromBase58Check` we can't really check if it's sane either ...

what was the original reason for seperating `fromBase58Check` and putting the exceptions there?

src/transaction_builder.js
```diff
@@ -401,6 +447,18 @@ TransactionBuilder.prototype.sign = function (index, keyPair, redeemScript, hash
 
           break
 
+        case 'segwitpubkeyhash':
```
rubensayshi: for P2WPKH as P2SH we place the `redeemScript` (short version) in the `scriptSig` but we'll use the long version when generating the signatureHash (hence we added `signatureScript`)

src/transaction.js
```diff
@@ -250,8 +399,22 @@ Transaction.prototype.getId = function () {
   return [].reverse.call(this.getHash()).toString('hex')
 }
 
-Transaction.prototype.toBuffer = function () {
-  var buffer = new Buffer(this.byteLength())
+Transaction.prototype.toBuffer = function (withWitness) {
```
rubensayshi: I think it would make more sense that `withWitness` would default to TRUE, because the only reason for wanting the non-witness is to `getId` or if you want to submit it to an old node or some old piece of software

and in the tests that would mean I'll add a `txHexWithoutWitness` instead of the `txHexWithWitness` that I've added so far in a few places ... 

/discuss

rubensayshi:
should I update other places where the `BufferWriter` can be used too?

src/transaction.js
```diff
@@ -250,8 +399,22 @@ Transaction.prototype.getId = function () {
...
+  // console.log('tx.Buffer')
+
+  // @TODO: it would be nicer if the marker/flag would be set when witness is set on one of the txIn...
```
rubensayshi: /discuss

dcousens:
Not yet,  lets keep the changes as atomic as necessary.  We can refactor later.

dcousens:
@greenaddress lots of WIP here,  be weary of using this until its ready. :+1:

greenaddress:
@dcousens ack thanks!

bityogi:
ACK. Thanks for working on this.

rubensayshi:
depends on https://github.com/bitcoinjs/bitcoinjs-lib/pull/524

.gitignore
```diff
@@ -1,3 +1,5 @@
 bitcoin.js
 coverage
 node_modules
+
+_dev.*.js
```
rubensayshi: hope you don't mind, but this makes it easy to keep debug files around

rubensayshi:
had to do a refactor of `extractInputs` and `__build` to be able to support P2SH[ P2WSH[ multsig / p2pkh / p2wpkh ] ]

I'm gonna see if I can seperate those changes into a seperate PR to make it easier to review (doing that right now)

created an overview of different scripts that are supported and what is what; https://gist.github.com/rubensayshi/fbe39657e90ef416d411

dcousens:
@rubensayshi could we start writing up the CHANGELOG for this such that the changes can be quicker to review?

dcousens:
Might move this to `2.4.0`

bpdavenport:
Curious if anyone is still working on this. @dcousens is there any other work on segwit happening besides this pull request?

dcousens:
@bpdavenport yes,  but it isn't public at this time.  I intend to have SegWit support before activation :+1:

rubensayshi:
@dcousens I was about to ask you about this, because I noticed the refactoring of a few things ...

let me know if there's anything I can do!

btcdrak:
It would be great to get this finished and merged so it can be used on testnet. Segwit is merged to bitcoin/master and active on Bitcoin testnet. It means people can start working on integration in earnest.

dcousens:
@btcdrak time permitting, I'll see if I can't work towards a 10th July PR date.

src/transaction.js
dcousens: I don't know if this is the right way to do it,  I've been playing with `toWitnessBuffer`,  or `WitnessTransaction` inheriting from `Transaction` where `toBuffer` is overwritten... still not sure.
I think we might just be better off detecting if a witness is needed and not bothering with a flag.

src/transaction.js
```diff
@@ -170,9 +229,100 @@ var VALUE_UINT64_MAX = new Buffer('ffffffffffffffff', 'hex')
...
+  typeforce(types.tuple(types.UInt32, types.Buffer, /* types.UInt8 */ types.Number, types.maybe(types.Boolean)), arguments)
+
+  if (segWit) {
+    return this.hashForSignatureV2(inIndex, prevOutScript, amount, hashType)
+  } else {
+    return this.hashForSignatureV1(inIndex, prevOutScript, hashType)
+  }
+}
+
+Transaction.prototype.hashForSignatureV2 = function (inIndex, prevOutScript, amount, hashType) {
```
dcousens: Why is this necessary?

src/transaction.js
```diff
@@ -4,9 +4,15 @@ var bufferutils = require('./bufferutils')
...
 var types = require('./types')
+var BufferWriter = require('./bufferwriter')
+
+var ADVANCED_TRANSACTION_MARKER = 0
+var ADVANCED_TRANSACTION_FLAG = 1
 
 function Transaction () {
   this.version = 1
+  this.marker = null
+  this.flag = null
```
dcousens: Are these ever going to change?
My understanding was the `0` marker was to allow for easy detection in serialization?

dcousens:
Lots of debugging `console.log` lefts in here

.gitignore
dcousens: ? What

src/bufferwriter.js
```diff
@@ -0,0 +1,55 @@
+var bufferutils = require('./bufferutils')
```
dcousens: I'd rather opt for something like `varstruct` than go this route if possible

src/address.js
```diff
@@ -6,30 +6,54 @@ var types = require('./types')
...
   if (bscript.isScriptHashOutput(scriptPubKey)) return toBase58Check(bscript.compile(scriptPubKey).slice(2, 22), network.scriptHash)
+  if (bscript.isSegWitPubKeyHashOutput(scriptPubKey)) return toBase58Check(bscript.compile(scriptPubKey).slice(2, 22), network.segWitPubKeyHash, 0)
+  if (bscript.isSegWitScriptHashOutput(scriptPubKey)) return toBase58Check(bscript.compile(scriptPubKey).slice(2, 34), network.segWitScriptHash, 0)
 
   throw new Error(bscript.toASM(scriptPubKey) + ' has no matching Address')
 }
 
-function toBase58Check (hash, version) {
-  typeforce(types.tuple(types.Hash160bit, types.UInt8), arguments)
+function toBase58Check (hash, version, segWitVersion) {
```
dcousens: It'd be great if we can split out all the address based changes to another PR

dcousens:
Ideally it'd be great to do this in 3 PRs.

`scripts` (https://github.com/bitcoinjs/bitcoinjs-lib/pull/602)
`Transaction`
`TransactionBuilder`
`address`

dcousens:
> Cache hashPrevouts, hashSequence and hashOutputs for all signatures in a transaction (since that was one of the main reasons for changing the signatureHash xD)

Why are we doing such a change in this PR?

btcdrak:
*bump*

btcdrak:
Note for OP, BIP142 was deferred

btcdrak:
The witness program size was extended from 32 to max 40 bytes https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki#Witness_program

dcousens:
@btcdrak support for SegWit scripts was already added a small while ago.  
This PR is just serving as a reminder to implement the transaction serialisation logic. 

See https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/src/script.js#L209-L223 and https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/src/script.js#L343-L354 and https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/src/script.js#L382

btcdrak:
Also need to disable uncompressed keys which are disallowed under segwit.

dcousens:
> Also need to disable uncompressed keys which are disallowed under segwit.

Is that in the BIP?

btcdrak:
@dcousens See BIP143 https://github.com/bitcoin/bips/pull/459/files#diff-d6a595a37a5d316054e79e53e9c2382fR128

src/transaction.js
afk11: Not sure if this was cleared up - segwit has a different signature hashing scheme which commits the input amount in the signature, amongst other changes. It has a reusable midstate for all inputs in a transaction which may be cached, so avoids the O(n^2) hashing issue.

afk11:
I might find some time for this coming up, the release including segwit activation is coming up pretty close. There are some breaks in the API, probably passing the sigVersion in places, and the input amount now needs to be known during signing. 

Mostly, sigVersion can be determined in situ, and by including the amount, the full TxOut is now a requirement for signing (previously it was just the scriptPubKey). To address the upcoming BC break, I'd suggest accepting a full txout in the signing API (eg, `hashForSignatureV2`) mostly to avoid cluttering things with extra parameters.

src/transaction.js
afk11: The zero marker makes it obvious to old clients that something is wrong (zero-inputs), and I don't think it will change. Flags may change - it's interpreted as a bit-field where the first indicates segwit support, but setting anything else will intentionally result in an error. https://github.com/bitcoin/bitcoin/blob/2f71490d21796594ca6f55e375558944de9db5a0/src/primitives/transaction.h#L320

dcousens:
>  and the input amount now needs to be known during signing.

That's the first I've seen of that,  but again,  my attention has been diverted?  Any references?

dcousens:
@afk11 do you have a medium with which we can collaborate on that effort?

afk11:
> That's the first I've seen of that, but again, my attention has been diverted? Any references?

I think the 'motivation' is scenarios like a Trezor being fed the wrong UTXO amount, since it blindly signs. If we commit the amount in the sig-hash, full nodes would reject a signature where the wrong amount was used. 

https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki#specification 

I'm starting to think for a signing API, the requirements (bar keypairs) are the `TxOut`, `redeemScript`, `witnessScript`, and `hashType`. Looking at it further, I'd love to remove the prevOutScript from addInput, and consolidate it into a txout parameter that replaces `amount` in `TransactionBuilder.sign` as seen in this PR. 

### Inferences

I notice for P2SH the library infers prevOutScript based on the redeemScript. It could probably do this based with segwit scripts too, although I prefer requiring the full TxOut from users always, and only optionally taking a redeemScript / witnessScript. 

Core doesn't infer things, it works from the scriptPubKey and decides what it needs next. (Solver in script/standard.cpp) While checking p2sh?/witness?/bare-scriptPubKey? it determines "is it a standard type, and what sighash sigVersion do I use?". P2SH / P2WSH are treated specially, and involve checking elsewhere to see that the hashes match and another call to Solver. 

Probably not a major issue, but I like there being a chain of hashes indicating which `witnessScript` to use (the reverse of infering the scriptPubKey) because then your wallet works based strictly on what scriptPubKey's do you have the required data for. 

> @afk11 do you have a medium with which we can collaborate on that effort?

Core's slack could work, otherwise happy to use email / XMPP at me$(at)$(IRL name)$(dot)io

dcousens:
> I'd love to remove the prevOutScript from addInput, and consolidate it into a txout parameter that replaces amount in TransactionBuilder.sign as seen in this PR.

Interestingly, I'd prefer to move more to `addInput` and keep `sign` as minimal as possible

afk11:
Ah ok :) I suppose the main reason would be to keep signing atomic - given a transaction, signing only takes so much `(nIn, PrivateKeyInterface key, TransactionOutputInterface txOut, ScriptInterface redeemScript = null, ScriptInterface witnessScript = null, sigHashType = SigHashInterface::ALL)`

Whether we pass all the arguments at signing, or all the arguments at addInput probably doesn't matter, but at least keep it pure, since I don't see any other method to set the prevOutScript (except for directly modifying internal values). I guess my disagreement stems from (i) it being split across two functions and, (ii) we're saving state anticipating it will only be used in that later function, so why not provide it only when necessary?

Also believe the scriptPubKey should be the thing you check a redeemScript against to test suitability when signing, rather than inferring a scriptPubKey from an arbitrary redeemScript and signing with it. I really like how Core work out what to sign with given only a scriptPubKey and it's key/script store: https://github.com/bitcoin/bitcoin/blob/master/src/script/sign.cpp#L139
</opinion>

dcousens:
As discussed on slack, we should set up an example somewhere that idiomatically uses 
P2PKH, P2SH, P2WPKH and P2WSH in succession, with all the necessary pieces of information.
Obviously the `TransactionBuilder` API isn't ready for the task,  but from that point we could play with the API prospectively, seeing what fits nicely and what doesn't.

dcousens:
Updated milestone to `3.1.0`

src/transaction.js
afk11: It was mentioned yesterday, but worth documenting: a param or separate serialization methods are required because you *should* be able to convert a segwit tx into the older version. The other way around (old to new) doesn't make sense, but is still technically valid to witness-enabled software.

src/address.js
afk11: They shelved that proposal until later - pending feedback/proposals from wallet developers IIRC

dcousens:
Closing in favour of the newer PRs.
Captured some of the checklist with https://github.com/bitcoinjs/bitcoinjs-lib/issues/686 and https://github.com/bitcoinjs/bitcoinjs-lib/issues/685

PR #522, opened 2016-01-22
Added Nubits to supported networks.

jgeewax:
I have no idea whether you guys are interested in adding more alt-coins, so if that's out of scope for your project, feel free to kill this PR. If not, here's some stuff to make Nubits work.

I manually tested that the addresses are valid against the latest Nu client (as well as the BIP32 stuff), and there's a simple test akin to Litecoin included.

### DIFF
--- a/src/networks.js
+++ b/src/networks.js
@@ -45,5 +45,16 @@ module.exports = {
     scriptHash: 0x16,
     wif: 0x9e,
     dustThreshold: 0 // https://github.com/dogecoin/dogecoin/blob/v1.7.1/src/core.h#L155-L160
+  },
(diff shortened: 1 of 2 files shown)


jprichardson:
I'm not 100% sure of @dcousens position on this, but the idea is to remove these from `bitcoinjs-lib`. I could be wrong on this though. There's another library that maintains these constants: https://github.com/cryptocoinjs/coininfo

@dcousens as an aside, I'm going to modify `coininfo` to return a `bitcoinjs-lib` compatible data structure for easy interoperation.

dcousens:
@jprichardson :+1:

PR #524, opened 2016-01-26
locktime and version were set too late in test scenario 

rubensayshi:
locktime and version were set too late in test scenario and not included when signing, seems the `txHex` to compare against was generated with this code so it needed to be updated too


dcousens:
True,  I see what you mean,  the signatures don't take into account the changed lock time.

ACK.

Can't merge yet due to the stupid integration tests failing.

dcousens:
Merging without status checks.
Tests manually verified.

Opened https://github.com/bitcoinjs/bitcoinjs-lib/issues/527 for integration test issues.

PR #526, opened 2016-01-27
README: use https for badges

dcousens:


dcousens:
Merging without status checks.
Tests manually verified.

Opened https://github.com/bitcoinjs/bitcoinjs-lib/issues/527 for integration test issues.

PR #529, opened 2016-01-27
Fixed tests

rubensayshi:
- there was another transactionbuilder test which had a fixture of a tx signed before locktime was set (#524)
 - fixed integration test that was fed a UTXO with a value different from what it expected


dcousens:
Tests still failing.

test/integration/_blockchain.js
```diff
@@ -24,7 +24,7 @@ testnet.faucet = function faucet (address, amount, done) {
         if (err) return callback(err)
 
         var unspent = result.filter(function (unspent) {
-          return unspent.value > 1e3
+          return unspent.value > amount
```
dcousens: This isn't right.  You might want `>=` if you're looking to filter non-amount values, which is probably the better way to do it.
I was previously just filtering all dust-like values `<= 1e3`.

test/integration/multisig.js
```diff
@@ -40,6 +40,9 @@ describe('bitcoinjs-lib (multisig)', function () {
     blockchain.t.faucet(address, 2e4, function (err, unspent) {
       if (err) return done(err)
 
+      var fee = 1e4
+      var send = unspent.value - fee
```
dcousens: `s/send/targetValue`?

rubensayshi:
didn't fix it either, will make sure it get's fixed tomorrow!

test/integration/multisig.js
```diff
@@ -60,16 +63,19 @@ describe('bitcoinjs-lib (multisig)', function () {
       blockchain.t.transactions.propagate(tx.toHex(), function (err) {
         if (err) return done(err)
 
-        // check that the above transaction included the intended address
-        blockchain.t.addresses.unspents(targetAddress, function (err, unspents) {
-          if (err) return done(err)
+        // allow for TX to be processed
+        setTimeout(function () {
```
dcousens: should this really be necessary?

rubensayshi: hmm, it's been working fine without it (issues seem unrelated) but wouldn't hurt to have a short timeout.

propagate (and faucet) broadcast TX through bitcoind which then relays it to our custom node, so there is a short delay in getting the TX processed (which is slightly more during stresstests or during block processing).  
normally the overhead of the http requests covers the time it takes to process it, but a short timeout helps, I'll lower it to 200ms.

rubensayshi:
I think I've found this real issue, which is - like you properly concluded already - in our faucet, spamming `mocha test/integration` to see if it no longer pops up

dcousens:
@jprichardson status checks re-enabled.
Yay.

Thanks @rubensayshi :+1:

dcousens:
> that shouldn't have happened and shouldn't happen again, in case it does it needs fixing and not async.retry spamming.

Well,  we did what we could :)
To be fair,  I did ping you in https://github.com/bitcoinjs/bitcoinjs-lib/pull/469 with no reply.

PR #530, opened 2016-01-28
address: fix compiled assumption for fromOutputScript

dcousens:
All of the `script.*` functions allow you to provide a compiled or uncompiled script,  and it will decide based on the input type whether to compile/decompile it.

This is actually a major optimization in many cases since you may be doing this many times over and the functions vary as to what state of script they prefer.

`fromOutputScript` actually already supports this,  but if a valid script is found,  then it fails with `toBase58Check` complaining about a non-buffer being given (confusing!).

### DIFF
--- a/src/address.js
+++ b/src/address.js
@@ -18,8 +18,8 @@ function fromBase58Check (address) {
 function fromOutputScript (scriptPubKey, network) {
   network = network || networks.bitcoin
 
-  if (bscript.isPubKeyHashOutput(scriptPubKey)) return toBase58Check(scriptPubKey.slice(3, 23), network.pubKeyHash)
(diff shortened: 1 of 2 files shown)


jprichardson:
As an aside... asking out curiosity...

why:

```js
var bscript = require('./script')
```

Is it named `bscript` and not `script`? Not suggesting it needs to be changed at all, just curious.

dcousens:
@jprichardson I prefixed a few modules with `b` internally because already had local variables named `script`,  and typical behaviour in production environments is to use it with the namespace `bitcoin`.

dcousens:
Internally:

``` javascript
var bscript = require('./script')
```

Externally:
``` javascript
var bitcoin = require('bitcoinjs-lib')

bitcoin.script.compile(...)
```

Issue #527, opened 2016-01-27
Fix integration tests from periodical failure


dcousens:
It seems to periodically fail due on BlockTrail's end, @rubensayshi ?

dcousens:
@rubensayshi this is still an issue.

See https://travis-ci.org/bitcoinjs/bitcoinjs-lib/jobs/109219660.

fanatid:
@dcousens as I understand it issue with faucet?
We can use [bitcoin-faucet-api](https://github.com/fanatid/bitcoin-faucet-api), only need install somewhere.

dcousens:
This isn't an issue with faucet anymore.  Its a matter of the BlockTrail common-blockchain API failing occasionally.

dcousens:
Seems to have stopped happening for now.
If it happens again I'll be changing it over to a different API that I know is solid

PR #531, opened 2016-01-28
fixed another test where TX was signed before locktime was set

dcousens:
From https://github.com/bitcoinjs/bitcoinjs-lib/pull/529,  doesn't fix the integration tests,  but it does resolve an issue that came from https://github.com/bitcoinjs/bitcoinjs-lib/pull/507.

Can't wait for #527 to be resolved.

### DIFF
--- a/test/fixtures/transaction_builder.json
+++ b/test/fixtures/transaction_builder.json
@@ -266,7 +266,7 @@
       },
       {
         "description": "Transaction w/ nLockTime, pubKeyHash -> pubKeyHash",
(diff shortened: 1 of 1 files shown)


dcousens:
@jprichardson I'm thinking of releasing `2.2.0` after this,  will submit a PR before publish.

jprichardson:
utACK

dcousens:
https://travis-ci.org/bitcoinjs/bitcoinjs-lib/builds/105340182 showed that the only failing tests were the integration tests, to be resolved by #527

PR #533, opened 2016-02-05
networks: add genesis block ID

dcousens:
Ready to bike shed on the variable name, but,  I think adding this is definitely worth while.
@jprichardson thoughts?


jprichardson:
Haha, variable name is fine or `genesisBlockHash` (maybe too verbose). You forgot a comma though :stuck_out_tongue:

dcousens:
`genesisBlockHash` would mean (by our standards) that it is a `Buffer` type.   I don't know if that is more or less reasonable.

jprichardson:
ACK on https://github.com/bitcoinjs/bitcoinjs-lib/commit/ef7739c0376c97a7bc24186c4e27c7cb64dab7e6

src/networks.js
```diff
@@ -14,6 +15,7 @@ module.exports = {
     dustThreshold: 546 // https://github.com/bitcoin/bitcoin/blob/v0.9.2/src/core.h#L151-L162
   },
   testnet: {
```
dcousens: Perhaps for `3.0.0`, this should be `.testnet3`?  ping @fanatid @jprichardson

jprichardson: Isn't it implied that it's testnet3? Would naming it testnet3 be confusing? Don't a significant majority use testnet3 as opposed to any previous? I don't have a strong preference on the naming.

dcousens:
`types.js` needs updating to the new format

src/networks.js
dcousens: @jprichardson I suppose,  just thinking about avoiding changes in some unforseeable future change

dcousens: Actually, I think for this change to make sense,  it should be named `testnet3`.
If `testnet` is ever to change,  then the `genesisBlockId` will also change.

I see nothing wrong with aliasing `testnet` to the "latest" network however.
But `testnet3` should be independent IMHO...

dcousens:
:shrug:

PR #536, opened 2016-02-05
Adding function to decide if HDNode is private

runn1ng:
It's a very simple function that probably isn't that important, but can help someone.

If you like this idea, I will add some tests for that

### DIFF
--- a/src/hdnode.js
+++ b/src/hdnode.js
@@ -170,7 +170,7 @@ HDNode.prototype.toBase58 = function (__isPrivate) {
 
   // Version
   var network = this.keyPair.network
-  var version = this.keyPair.d ? network.bip32.private : network.bip32.public
+  var version = (!this.isNeutered()) ? network.bip32.private : network.bip32.public
   var buffer = new Buffer(78)
 
   // 4 bytes: version bytes
@@ -190,7 +190,7 @@ HDNode.prototype.toBase58 = function (__isPrivate) {
   this.chainCode.copy(buffer, 13)
 
   // 33 bytes: the public key or private key data
-  if (this.keyPair.d) {
+  if (!this.isNeutered()) {
     // 0x00 + k for private keys
     buffer.writeUInt8(0, 45)
     this.keyPair.d.toBuffer(32).copy(buffer, 46)
@@ -211,7 +211,7 @@ HDNode.prototype.derive = function (index) {
 
   // Hardened child
   if (isHardened) {
-    if (!this.keyPair.d) throw new TypeError('Could not derive hardened child key')
+    if (this.isNeutered()) throw new TypeError('Could not derive hardened child key')
 
     // data = 0x00 || ser256(kpar) || ser32(index)
     data[0] = 0x00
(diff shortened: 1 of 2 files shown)


dcousens:
Nice!

Maybe `isNeutered`?
That way it is consistent with the `neutered()` interface?

dcousens:
Also, are you up to add some tests? :+1:

runn1ng:
Hm. On one hand, it makes sense, because it's consistent, on the other hand I just wouldn't guess from the name what it means.

I will look more into the travis failures (seems like style issue... that's what I get for editing code in github GUI :))

src/hdnode.js
```diff
@@ -281,6 +281,10 @@ HDNode.prototype.deriveHardened = function (index) {
   return this.derive(index + HDNode.HIGHEST_BIT)
 }
 
+HDNode.prototype.isPrivate = function () {
+  return !!(this.keyPair.d);
```
fanatid: extra semicolon

dcousens:
why not both?

I wouldn't be opposed to having both then I guess?  Its just more to maintain and think about.
I'm just wanting to be consistent, especially since `isPrivate` is equally vague,  at least the `isNeutered` introduces less context.

@fanatid thoughts?

dcousens:
In any case @runn1ng,  please add tests in lieu of the bike shed.

fanatid:
Since `HDNode` already has [neutered](https://github.com/bitcoinjs/bitcoinjs-lib/blob/daafb9794e60a7329c7d5825fceab71e17f6f0a7/src/hdnode.js#L147), `isNeutered` is better.

runn1ng:
I have renamed the function and removed the semicolon. Now I am adding tests

runn1ng:
I have added some tests. Is it OK like this?

runn1ng:
...stupid semicolons! :D ok now?

runn1ng:
....the test failure is timeout failure and not related to my changes (I think).

test/hdnode.js
```diff
@@ -240,7 +242,7 @@ describe('HDNode', function () {
     var f = fixtures.valid[0]
 
     it('strips all private information', function () {
-      var hd = HDNode.fromBase58(f.master.base58, NETWORKS_LIST)
+      var hd = HDNode.fromBase58(f.master.base58Priv, NETWORKS_LIST)
```
dcousens: ha.  Good catch.

src/hdnode.js
```diff
@@ -239,7 +239,7 @@ HDNode.prototype.derive = function (index) {
 
   // Private parent key -> private child key
   var derivedKeyPair
-  if (this.keyPair.d) {
+  if (!this.isNeutered()) {
```
dcousens: I'm not sure if `isNeutered` is that great after all.
Its hard to derive the intended meaning... `isPrivate` would be much clearer here.

@runn1ng could you add both?
We'll know by `3.0.0` as to whether we should clean it up or not...

dcousens:
Sorry for the screw around @runn1ng... :+1:

dcousens:
Bugger it actually, thanks @runn1ng :+1: 

If `.isNeuetered` ends up being a bit too weird or confusing, we'll use `isPrivate`.

dcousens:
Won't publish for a little while yet,  will be in `2.3.0`

Issue #508, opened 2015-11-30
API for 4.0.0


dcousens:
Just a bike shed for ~~`3.0.0`~~ `4.0.0`, if its necessary, and the currently open PRs:

https://github.com/bitcoinjs/bitcoinjs-lib/pull/459 - Extraction of DER encoding
https://github.com/bitcoinjs/bitcoinjs-lib/pull/456 - Remove message module (`bitcoinjs-message`)
https://github.com/bitcoinjs/bitcoinjs-lib/pull/440 - Var naming conventions
https://github.com/bitcoinjs/bitcoinjs-lib/pull/350 - secp256k1
https://github.com/bitcoinjs/bitcoinjs-lib/pull/428 - Altcoin support

@jprichardson I was also thinking about `TransactionBuilder`, and what it represents.

It is basically a convenience wrapper around `Transaction` (the primitive) that makes everyone's life easier when building complex transactions, but, to be honest, it could easily be extracted from the core library.

Thoughts?
In general, I'm thinking we work on making this library primarily all the primitives for bitcoin operations such that other libraries can easily compose using them.
I'd like to see other crypto primitives like merkle trees, bloom filters and rolling bloom filters in the ecosystem too.

dcousens:
Naturally, keeping that list of related modules is super helpful (`BIPxx` etc), as it helps keep things related and the ecosystem linked together.

dcousens:
I think these decisions are becoming more and more important as more complicated structures and algorithms are becoming the norm (see opt-in RBF, CLTV, payment channels, etc).

jprichardson:
> It is basically a convenience wrapper around Transaction (the primitive) that makes everyone's life easier when building complex transactions, but, to be honest, it could easily be extracted from the core library.

Yes, you know I'd be in favor of breaking up the library into smaller modules =) I'd be willing to contribute any of my existing modules / names if you think they'd be of use. We can start fresh too.

> In general, I'm thinking we work on making this library primarily all the primitives for bitcoin operations such that other libraries can easily compose using them.
I'd like to see other crypto primitives like merkle trees, bloom filters and rolling bloom filters in the ecosystem too.

Yes, definitely.

dcousens:
Other thoughts:

`block.timestamp -> block.time` for consistency with bitcoin/bitcoin RPC.

jprichardson:
To add to the discussion, here's what I'd like:

[...]

PR #538, opened 2016-02-06
Adding path derivation

runn1ng:
Now this might be a bit controversial, but I found something like that to be really useful.

What will be controversial, I think:

* the usage of strings as indexes 
   * yes, it's not ideal, but adding an apostrophe is a way to easily say "this is hardened" and this notation is common, and you cannot do that otherwise
* the usage of descriptions as paths in the tests
   * again, not ideal, but I didn't want to repeat the description as path again if it's already there


dcousens:
ping @jprichardson,  merge here,  finalize API,  then extract to `BIP32`?

src/hdnode.js
```diff
@@ -281,6 +281,24 @@ HDNode.prototype.deriveHardened = function (index) {
   return this.derive(index + HDNode.HIGHEST_BIT)
 }
 
+// number + ' (e.g. 234') as a string -> hardened
+HDNode.prototype.derivePath = function (path) {
```
dcousens: Any reason why this just doesn't take a `string` as input?

dcousens:
IMHO,  I think  `node.derivePath('m/0'/1/0')` would be a better API,  aka a `String` argument.

I don't think there is much value in adding an `Array` based path API.  You can trivially do that using `derive` and `forEach` already.

dcousens:
We also intend to break out `BIP32` to its own module,  probably encompassing `HDNode` standalone with other utilities for making life easier when dealing with them.

See https://github.com/bitcoinjs/bitcoinjs-lib/issues/508#issuecomment-176188789

runn1ng:
ad array based API:

My original thinking for using Array was "What if someone puts there an array that is *not* a path, or that is wrongly formatted? Do I have to trim spaces around the slashes? Does it have to start with 'm/' or not? What if the node is not a master and is a sub-node, the 'm/' doesn't make sense", and that by using Array I just go around all these problems.

Then I also thought about the function accepting both array and string, but that would be really ugly.

jprichardson:
> ping @jprichardson, merge here, then extract to BIP32?

@dcousens Yep, I don't see any harm in adding it at the moment. Agreed on not needing the array API.

runn1ng:
how exactly would you do the string-based API? Should it start with 'm/', or should it be "relative"? So for example,` master.derivePath("0'/1/0'/1") `(which would equal to `master.derivePath("0'/1").derivePath("0'/1")`)

dcousens:
I think `0'/1/0` is a suitable enough API.
We could easily regex,  optionally skipping any prefixes until the first number is reached?

runn1ng:
By "we", you mean the code in the library, or the user of the API?

dcousens:
We, the developers of this library.
My suggestion was that we could make the `m/` optional,  but,  ambiguous API's suck,  so,  not sure.

runn1ng:
>ambiguous API's suck

Exactly. But then, I don't know how to make it unambiguous and non-sucky.

And I don't also want to totally abandon it either. 😓

dcousens:
@runn1ng I think if an `m/` is present,  we check that there is no parent (aka, this is a `master` node).
If it isn't present,  proceed on with derivation.

Thoughts?

runn1ng:
Sorry for the delay. OK, that sounds reasonable

runn1ng:
Done :)

src/hdnode.js
```diff
@@ -291,6 +291,30 @@ HDNode.prototype.isNeutered = function () {
...
+  typeforce(types.String, path)
+
+  var splitPath = path.split('/')
+  if (splitPath[0] === 'm') {
+    if (!this.parentFingerprint) {
+      splitPath = splitPath.slice(1)
+    } else {
+      throw new Error('Not a master node')
+    }
```
dcousens:
``` javascript
if (this.parentFingerprint) throw new Error('Not a master node')
```

runn1ng: I am not sure standard allow if block without curly braces :) but yeah I got you. Ok, will do tomorrow

runn1ng:
I have done some cosmetic changes and rebased. The failing check has nothing to do with my code.

src/hdnode.js
runn1ng: done

dcousens:
Sounds good, will review asap :)

fanatid:
@runn1ng why unary plus instead `praseInt`?
with unary plus string `/` equals `0/0` because `+''` gives `0`, with `parseInt` string `/` throw error.
```js
return splitPath.reduce(function (prevHd, indexStr) {
  if (indexStr.slice(-1) !== "'") return prevHd.derive(parseInt(indexStr, 10))
  return prevHd.deriveHardened(parseInt(indexStr.slice(0, -1), 10))
}, this)
```

runn1ng:
Good catch.

runn1ng:
Just thinking.... if you put in path with decimal points ("m/0.5/1") by some mistake, using "parseInt" will silently truncate it instead of throwing errors.

But I am not sure that is worth caring about.

fanatid:
As option, you can check that result of `parseInt` is equal original string, something llike this:
```js
return splitPath.reduce(function (prevHd, indexStr) {
  var derive = prevHd.derive
  if (indexStr.slice(-1) === "'") {
    derive = prevHd.deriveHardened
    indexStr = indexStr.slice(0, -1)
  }

  var index = parseInt(indexStr, 10)
  if (index.toString() !== indexStr) throw new Error(...)

  return derive.call(prevHd, index)
}, this)
```

runn1ng:
I am also thinking if things like `m/   2   /   0'  /` should be allowed. `parseInt` will munch it just fine.

I will probably do what you suggest.

Alternatively, I can just test the index with regexp if it's only numbers.

dcousens:
`/([m]\/)?([0-9]+[']?\/)*([0-9]+[']?)/`

runn1ng:
Well, I wanted to test indices one by one after splitting, which is more readable and easily debuggable, but it's true that testing the parameter at the start of a function is more in line with the rest of the code.

runn1ng:
I have added path validation to typeforce types.

src/types.js
```diff
@@ -26,6 +26,11 @@ function UInt53 (value) {
     Math.floor(value) === value
 }
 
+function Path (value) {
```
dcousens: `BIP32Path`?  *shrug*

It's private anyway,  so I'm not worried too much.

test/fixtures/hdnode.json
```diff
@@ -134,7 +134,7 @@
           "address": "1BxrAr2pHpeBheusmd6fHDP2tSLAUa3qsW"
         },
         {
-          "description": "m/0/2147483647'/1/2147483646",
+          "description": "m/0/2147483647'/1/2147483646'",
```
dcousens: Hah.  Good catch.

test/hdnode.js
```diff
@@ -285,6 +286,42 @@ describe('HDNode', function () {
           verifyVector(hd, c, i + 1)
         })
       })
+
+      // testing deriving path from master
+      f.children.forEach(function (c) {
+        it(c.description + ' from ' + f.master.fingerprint + ' by path', function () {
+          var path = c.description
+          var child = master.derivePath(path)
```
dcousens: Nice :)

test/hdnode.js
```diff
@@ -368,6 +405,30 @@ describe('HDNode', function () {
...
+      }, /Expected Path/)
+      assert.throws(function () {
+        master.derivePath([2, 3, 4])
+      }, /Expected Path/)
+      assert.throws(function () {
+        master.derivePath('/')
+      }, /Expected Path/)
+      assert.throws(function () {
+        master.derivePath('m/m/123')
+      }, /Expected Path/)
```
dcousens: Its not great,   but,  its not terrible either.  ACK for now with room for improvement haha

fanatid:
```js
[
  'foo',
  '',
  2,
  ...
].forEach(function (path) {
  assert.throws(function () { master.derivePath(path) },  /Expected Path/)
})
```

src/types.js
fanatid: +1 for `BIP32Path`

test/hdnode.js
dcousens: And put those negative fixtures in the json fixtures under `invalid: { paths: [...] }`

dcousens: See https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/test/fixtures/hdnode.json#L195 for examples

dcousens: @runn1ng if you're time is running short on this (not that there has been any hint of that),  I'll see if I can submit a PR to this branch (on your repo) to help out :)

runn1ng: All right, I can do that

src/types.js
runn1ng: Ok

runn1ng:
OK, feedback was put in

fanatid:
ACK

dcousens:
ACK :+1:

runn1ng:
:sunglasses: great

test/fixtures/hdnode.json
```diff
@@ -109,7 +109,7 @@
           "address": "19EuDJdgfRkwCmRzbzVBHZWQG9QNWhftbZ"
         },
         {
-          "description": "m/0/2147483647",
+          "description": "m/0/2147483647'",
```
dcousens: I might rename this to `path`

runn1ng: That would be great

src/hdnode.js
```diff
@@ -291,6 +291,30 @@ HDNode.prototype.isNeutered = function () {
   return !(this.keyPair.d)
 }
 
+HDNode.prototype.derivePath = function (path) {
+  typeforce(types.Bip32Path, path)
+
+  var splitPath = path.split('/')
+  if (splitPath[0] === 'm') {
+    if (this.parentFingerprint) {
+      throw new Error('Not a master node')
```
dcousens: Note,  I might change this error message in near future versions.  Consider unstable

src/hdnode.js
```diff
@@ -291,6 +291,30 @@ HDNode.prototype.isNeutered = function () {
...
+  return splitPath.reduce(function (prevHd, indexStr) {
+    var index
+    if (indexStr.slice(-1) === "'") {
+      index = parseInt(indexStr.slice(0, -1), 10)
+      return prevHd.deriveHardened(index)
+    } else {
+      index = parseInt(indexStr, 10)
+      return prevHd.derive(index)
+    }
+  }, this)
```
dcousens: nit: `this` unnecessary here

runn1ng: Not sure, what would happen with empty paths

runn1ng:
Thanks for merging, I kind of forgot about this PR :)

src/hdnode.js
dcousens: Oh woops, of course.  Ignore my comment.

PR #539, opened 2016-02-06
Adding some checks on deriving indexes

runn1ng:
I have added some checks on hardening indexes.

You don't want to accidentally put in twice hardened index, or a negative number to harden (which produces non-hardened index), etc.

### DIFF
--- a/src/hdnode.js
+++ b/src/hdnode.js
@@ -206,6 +206,8 @@ HDNode.prototype.toBase58 = function (__isPrivate) {
 
 // https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#child-key-derivation-ckd-functions
 HDNode.prototype.derive = function (index) {
+  typeforce(types.UInt32, index)
+
   var isHardened = index >= HDNode.HIGHEST_BIT
   var data = new Buffer(37)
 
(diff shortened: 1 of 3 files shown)


src/hdnode.js
```diff
@@ -206,6 +206,15 @@ HDNode.prototype.toBase58 = function (__isPrivate) {
...
 HDNode.prototype.derive = function (index) {
+  typeforce('Number', index)
+  var highestIndex = (HDNode.HIGHEST_BIT * 2) // should be 0x100000000
+  if (index < 0) {
+    throw new Error('Could not derive with negative index')
+  }
+  if (index >= highestIndex) {
+    throw new Error('Could not derive with too high index')
+  }
+
```
dcousens: @runn1ng it might be easiest to make a custom [typeforce](https://github.com/dcousens/typeforce) type for this :)

runn1ng: I have to say I don't understand typeforce that much, but it seems it's actually pretty easy to do that in there. How do you want to name the types... UnsignedInt31 and UnsignedInt32? :) (can typeforce right now to check if the number is integer by the way?)

runn1ng: also would you want these types to be added into typeforce source directly, or rather somewhere in bitcoin.js?

runn1ng: ...oh, it seems you actually *do* have the types already in bitcoinjs (well not the 31 bit one, but that can be added). OK, will use that instead

runn1ng:
yep you were right, it's easier with typeforce types

dcousens:
Nice @runn1ng :+1:

PR #541, opened 2016-02-14
Adding getPubKeyHash to HDNode and keypair

runn1ng:
I found out this can get quite useful.

If you think this is good to have, I will add some tests.

### DIFF
--- a/src/ecpair.js
+++ b/src/ecpair.js
@@ -112,6 +112,10 @@ ECPair.prototype.getPublicKeyBuffer = function () {
   return this.Q.getEncoded(this.compressed)
 }
 
+ECPair.prototype.getPubKeyHash = function () {
+  return bcrypto.hash160(this.getPublicKeyBuffer())
+}
+
 ECPair.prototype.sign = function (hash) {
   if (!this.d) throw new Error('Missing private key')
 
--- a/src/hdnode.js
+++ b/src/hdnode.js
@@ -1,5 +1,4 @@
 var base58check = require('bs58check')
-var bcrypto = require('./crypto')
 var createHmac = require('create-hmac')
 var typeforce = require('typeforce')
 var types = require('./types')
@@ -131,7 +130,11 @@ HDNode.prototype.getAddress = function () {
 }
 
 HDNode.prototype.getIdentifier = function () {
-  return bcrypto.hash160(this.keyPair.getPublicKeyBuffer())
+  return this.getPubKeyHash()
+}
+
+HDNode.prototype.getPubKeyHash = function () {
+  return this.keyPair.getPubKeyHash()
 }
 
 HDNode.prototype.getFingerprint = function () {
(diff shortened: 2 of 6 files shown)


dcousens:
I could never decide if this was worth it.   @fanatid @jprichardson ?
I'm neutral either way.

runn1ng:
I will work on the tests if it's worth the time spent on them :)

I would definitely use it; although I am actually using it from HDNode, so `node.getIdentifier` is enough anyway - but it's less clear what it actually does.

dcousens:
Needs rebase

runn1ng:
Rebased + added some tests. Again I don't really care that much if it won't be accepted, I can use it by `getIdentifier`

dcousens:
@fanatid penny for your thoughts?

fanatid:
@dcousens idk, at least if you want merge this, I recommend rename `getPubKeyHash` to `getPublicKeyHash`
@runn1ng OOI for what purpose you want use `getPubKeyHash` from `HDNode`?

dcousens:
IMHO I almost always just do `bitcoin.crypto.hash160(keyPair.getPublicKey())`

runn1ng:
@fanatid 
Well, I wanted to use it (in Trezor client) to check if transaction with output to my own node has correct script. Trezor tells me the final outputs and I want to check if it created the script correctly; when Trezor creates the output to its own node (most often change node), I send it only the path of the node, not Base58check address (then the user doesn't need to confirm on device; it's his own address!). To check if the script is correct, I derive the HDNode on the client and compare the script in the transaction with the expected script, for which I need the pubkeyhash.

Sure, I can first derive a b58c address and then decode it and get the hash from there, but that's one extra step. :)

dcousens:
Needs rebase.

I'm going to ACK this change,  as it would probably be useful for many.
Anyone against?

jprichardson:
> Anyone against?

Nope. I'm not keen on the naming scheme... but no need for me to bikeshed on 2.x. We can save that fun for 3.x =)

dcousens:
@runn1ng if you rebase I'll merge.

dcousens:
ping @runn1ng

dcousens:
ping @runn1ng

runn1ng:
Going to rebase now, sorry about the huge delay

runn1ng:
Hm. The integration tests are failing, I am not sure why.

dcousens:
@runn1ng the integration tests failing are due to Blocktrail failing to give us an unspent to test with (ping @rubensayshi).
Not your fault.

dcousens:
Still not sure how I feel. 
`pubKeyHash` is only going to become less useful now that SegWit is soon upon us.
It's almost identical,  but without the fee subsidy.

jprichardson:
> pubKeyHash is only going to become less useful now that SegWit is soon upon us.
P2WSH w/ a 160-bit hash will be identical in security, and will have a fee subsidy.

👍

PR #542, opened 2016-02-15
whitelist of files for npm package

fanatid:
Only src, because (from docs):

>
Certain files are always included, regardless of settings:
  * package.json
  * README (and its variants)
  * CHANGELOG (and its variants)
  * LICENSE / LICENCE

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,19 @@
+sudo: false
 language: node_js
 before_install:
   - npm install npm -g
 node_js:
-  - 5
-  - 4
-  - io.js
-  - 0.12
-  - 0.10
+  - "0.10"
+  - "0.12"
+  - "io.js"
+  - "4"
+  - "5"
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,19 +1,21 @@
+sudo: false
+os:
+  - linux
```
dcousens: why `linux`?
I mean,  I guess I always assumed this was the case,  but,  wouldn't it be default?

fanatid: Good question, `linux` is os by default, but I can't find this in the travis docs. Should I remove this line?

dcousens: I think its overly verbose and doesn't serve any meaningful purpose.  Personally.

fanatid:
updated
@dcousens please check slack

dcousens:
Cheers @fanatid,  great work :)

PR #546, opened 2016-02-18
Adding name to network object

runn1ng:
It would be quite useful to us to have the ability to read network name back from the network object.

We are migrating from our custom structures to bitcoin.js structures, but info about network name is something we are missing there.

(This might break some tests; I will edit them if you are not against this on principle.)

### DIFF
--- a/src/networks.js
+++ b/src/networks.js
@@ -3,6 +3,7 @@
 
 module.exports = {
   bitcoin: {
+    name: 'bitcoin',
     messagePrefix: '\x18Bitcoin Signed Message:\n',
     bip32: {
       public: 0x0488b21e,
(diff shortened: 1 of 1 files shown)


fanatid:
just for you information, [coininfo](http://github.com/cryptocoinjs/coininfo/) can do this
but as I remember `coininfo` not in plans for [v3.0.0](https://github.com/bitcoinjs/bitcoinjs-lib/milestones/3.0.0)
ping @dcousens

dcousens:
@fanatid it could be

dcousens:
ping @jprichardson how drop in can `coininfo` be?

jprichardson:
@dcousens drop-in ftw https://github.com/cryptocoinjs/coininfo#want-to-use-with-bitcore-or-bitconjs-lib

src/networks.js
```diff
@@ -3,6 +3,7 @@
 
 module.exports = {
   bitcoin: {
+    name: 'bitcoin',
```
dcousens: perhaps this should be `Bitcoin`?

dcousens:
IMHO if this is added,  it should be their formal names (capitalized?).  But I'm honestly thinking we avoid this?

@runn1ng what was your use case?

runn1ng:
It's not that useful in the end; I am closing if it's all right with you

dcousens:
No worries.
@runn1ng for others,  could you explain why you felt it was no longer useful?

PR #549, opened 2016-02-24
Update wif to 2.0.1

fanatid:


src/ecpair.js
```diff
@@ -66,11 +66,12 @@ ECPair.fromWIF = function (string, network) {
 
     network = network.filter(function (network) {
       return version === network.wif
-    }).pop() || {}
+    }).pop()
+    if (!network) throw new Error('Invalid network version')
```
dcousens: Can we add a test for this?

dcousens: Oh wait, it is tested.  Why do this then?

dcousens:
LGTM, ready to merge after nit addressed

src/ecpair.js
fanatid: I made this because wif version is now optional: https://github.com/bitcoinjs/wif/blob/8355b2bba749de38742ef88f6d8d46992afe7c5e/index.js#L5
We have discussion about this in this issue: https://github.com/bitcoinjs/wif/issues/3

dcousens: Will re-check this in the next 24 hours,  suspect its all OK just want to double check.

dcousens: This isn't tested.
I'll make a new PR if relevant.

Thanks for the PR @fanatid :)

dcousens: Oh!

I thought this was `HDNode`... never mind.

PR #550, opened 2016-02-25
Add network list tests for ECPair/HDNode

dcousens:
This actually fixes a bug `bip32 is undefined` when a networks list was passed and no valid version was found in `HDNode`.

### DIFF
--- a/src/hdnode.js
+++ b/src/hdnode.js
@@ -66,7 +66,7 @@ HDNode.fromBase58 = function (string, networks) {
              version === network.bip32.public
     }).pop()
 
-    if (!network) throw new Error('Invalid network version')
+    if (!network) throw new Error('Unknown network version')
 
   // otherwise, assume a network object (or default to bitcoin)
   } else {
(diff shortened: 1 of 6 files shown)


dcousens:
@fanatid please review and merge :+1:.

Thoughts on aae5db6?  On the 1 hand,  easier to verify the tests are actually following the path they should be (not ambiguous with the other other case).
On the other hand,  its two cases a user has to be aware of,  despite the fact only 1 should occur depending on whether they pass in a list or singular network parameter.

fanatid:
@dcousens I think throwing 2 different errors is OK, first says that not one of network from `networks` don't match, second says that version for given|selected network is invalid.
LGTM

src/hdnode.js
```diff
@@ -72,7 +74,7 @@ HDNode.fromBase58 = function (string, networks) {
   }
 
   if (version !== network.bip32.private &&
-    version !== network.bip32.public) throw new Error('Invalid network')
+    version !== network.bip32.public) throw new Error('Invalid network version')
```
fanatid: just curious, you don't use align in `if` blocks?
```js
if (version !== network.bip32.private &&
    version !== network.bip32.public) throw new Error('Invalid network version')
```

dcousens: Eh,  alignment isn't that big a deal IMHO.
Its also very hard to enforce semantically so I don't bother.

PR #540, opened 2016-02-08
Refactoring `extractInput` and `__build` to enable nested scripts

rubensayshi:
This is preperation for the segwit support which will have `P2SH[ P2WSH[ P2PKH / multsig ] ]` scripts.  
It's only refactoring code, shouldn't contain any function changes and should make the https://github.com/bitcoinjs/bitcoinjs-lib/pull/520 PR easier to diff once merged.

Considering the following possible scripts we want to support: https://gist.github.com/rubensayshi/fbe39657e90ef416d411

PR #551, opened 2016-02-25
Cleanup for #540

dcousens:
This should still be in the spirit of what you had in mind @rubensayshi.
I've just pulled out the in-line function declarations and named the functions different such that they don't all have the name `processScript`.

~~Still WIP. ~~


dcousens:
Needs rebase

PR #552, opened 2016-02-25
README: `1.0.0` is now more popular

dcousens:
This is somewhat hard to quantify,  so citations would be great.
But my general impression is that most people are using at least `1.0.0` (if not `2.0.0`) now,  and in fact even blockchain.info are using `2.0.0` AFAIK.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 
 
 The pure JavaScript Bitcoin library for node.js and browsers.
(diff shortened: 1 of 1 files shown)


dcousens:
ping @kyledrake

kyledrake:
This is what they're using I think: https://blockchain.info/Resources/wallet/bitcoinjs.min.js

That said, I think this is pretty much the canonical library for JS at this point, so I'm comfortable with the assertion. +1, go ahead and merge it.

Issue #513, opened 2015-12-17
Generalizable serialization format


dcousens:
Related to https://github.com/bitcoinjs/bitcoinjs-lib/pull/428,  would make patching in alternate crypto currencies much simpler.
Could potentially allow for stricter and more deterministic type checking.

I don't want to over-generalize though, and if this was to be done, it'd definitely be implemented/tested through an external module, not in `bitcoinjs-lib`.

**edit**:  Should also support encoding of `valueBuffer` natively
* Would also rename internal variables more consistently (aka #622)

jprichardson:
I'm definitely in favor of this and more than happy to take the lead on this. This is something that I definitely need.

dcousens:
@jprichardson taking your time on this one? :laughing: 
Going to un-assign for now

dcousens:
Wants https://github.com/varstruct/varstruct/pull/32

dcousens:
Bump to `5.0.0`?
I don't see anything easy to do here.

dcousens:
Moved to `5.0.0`

junderw:
Moving to 6.0.0 ...

brandoncurtis:
The workflow of using bitcoinjs-lib for multiple coins with network parameters provided by something like https://github.com/cryptocoinjs/coininfo is quite a nice one.  This improvement would enable that workflow for coins such as Decred, Zcash (https://github.com/bitcoinjs/bitcoinjs-lib/issues/865), and many more.

jprichardson:
yes, I'd love this. Exodus would be willing to fund the work here for this effort.

junderw:
tbh, I'm sort of at a loss on how to move forward on this.

Would be nice if we could get a laundry list going.

maxchernin:
Hi everyone, whats the status on this one?

junderw:
I'm closing this.

PR #554, opened 2016-02-27
varstruct - WIP

fanatid:
[varstruct](https://github.com/dominictarr/varstruct/) for encode/decode blocks/txs?

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,6 +73,8 @@
     "ecurve": "^1.0.0",
     "randombytes": "^2.0.1",
     "typeforce": "^1.6.2",
+    "varstruct": "^5.0.1",
+    "varuint-bitcoin": "^1.0.1",
     "wif": "^2.0.1"
   },
   "devDependencies": {
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -5,6 +5,23 @@ var opcodes = require('./opcodes')
 var typeforce = require('typeforce')
 var types = require('./types')
 
+var VStruct = require('varstruct')
+var VarUIntBitcoin = require('varuint-bitcoin')
+var TxStruct = VStruct([
+  { name: 'version', type: VStruct.UInt32LE },
+  { name: 'ins', type: VStruct.VarArray(VarUIntBitcoin, VStruct([
+    { name: 'hash', type: VStruct.Buffer(32) },
+    { name: 'index', type: VStruct.UInt32LE },
+    { name: 'script', type: VStruct.VarBuffer(VarUIntBitcoin) },
+    { name: 'sequence', type: VStruct.UInt32LE }
+  ])) },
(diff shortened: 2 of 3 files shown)


dcousens:
Will review soon :)

fanatid:
@dcousens please check it, I want to know you opinion about this.
I fixed `varstruct` and now `fromBuffer` works fine. `toBuffer` doesn't work because `txOut` can have `valueBuffer` instead `value` (I don't know how handle it, may be create custom codec...).

@mappum since same structures are used in [bitcoin-protocol](https://github.com/mappum/bitcoin-protocol) maybe make sense create package `bitcoin-struct-tx-block` or two `bitcoin-tx-struct` and `bitcoin-block-struct`?

mappum:
Maybe we should just make one `bitcoin-struct` package with all the structs in it?

fanatid:
Sorry, what do you mean under *all* ? `tx` and `block` ?

fanatid:
ping

mappum:
I was thinking there might be other structs we might want to include, such as merkleblocks and full blocks (although those are not used by bitcoinjs).

fanatid:
Looks like not actual right now, hope [varstruct](https://github.com/dominictarr/varstruct/) will be used in https://github.com/bitcoinjs/bitcoinjs-lib/issues/508

dcousens:
@fanatid this looks really good :+1:

dcousens:
`bitcoin-structs` might work,  assuming we just put all the basic POD types in there.

`Transaction`, `Block`?  Any others?

fanatid:
or move messages from `bitcoin-protocol` to `bitcoin-protocol-messages` and use tx, block from there
https://github.com/mappum/bitcoin-protocol/blob/10d67127c89ed1928094be84375f70dfef0a43b3/src/messages.js

dcousens:
@fanatid looks good to me. 
Though that would make https://github.com/bitcoinjs/bitcoinjs-lib/pull/440 either impossible, or something we'd need to do in lockstep with @mappum's `bitcoin-protocol`.

dcousens:
Maybe @mappum we could move `bitcoin-protocol` to the bitcoinjs-org,  you'd still be lead contributor, but then it would be clear that the repository is linked (and hence in lockstep)?

fanatid:
@dcousens if you wanted to move `bitcoin-protocol` you probably should do same with [bitcoin-net](https://github.com/mappum/bitcoin-net)

dcousens:
Well,  that is up to @mappum

dcousens:
I started playing with this @fanatid,  and ran into issues with `valueBuffer` edge cases.
I'd have to make a custom `encode`/`decode` that handled both cases (`Buffer` and `Number`).

dcousens:
Closing for now

dcousens:
Still covered by https://github.com/bitcoinjs/bitcoinjs-lib/issues/513

PR #556, opened 2016-03-08
Adding flow exports

runn1ng:
We are using [Facebook Flow](https://github.com/facebook/flow) for static typechecking of our code.

It's super useful, and we wrote our interface file for Flow, for the things we use the most. It isn't complete at all - it really just have things that we use.

Anyway, I thought it will be useful to put it upstream, even when it's not complete. (Also, it would be good for us, so we don't have to copy the file all the time :) etc.)

(Note: the semicolons are necessary here, since they are in the class property syntax.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 coverage
 node_modules
+test/flow/result
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ sudo: false
 language: node_js
 before_install:
   - npm install npm -g
+  - npm install flow-bin
 node_js:
   - "0.10"
   - "0.12"
@@ -16,4 +17,5 @@ env:
   - TEST_SUITE=coverage
   - TEST_SUITE=integration
   - TEST_SUITE=unit
+  - TEST_SUITE=flowtest
 script: npm run-script $TEST_SUITE
--- a/package.json
+++ b/package.json
@@ -38,7 +38,8 @@
     "prepublish": "npm run test",
     "standard": "standard",
     "test": "npm run standard && npm run unit",
-    "unit": "mocha"
+    "unit": "mocha",
+    "flowtest": "bash ./test/flow/run-test.sh"
   },
   "repository": {
     "type": "git",
--- a/src/index.js.flow
+++ b/src/index.js.flow
@@ -1,0 +1,85 @@
+/* @flow */
+export type Network = {
+  messagePrefix: string;
+  bip32: {
+    public: number;
+    private: number;
+  };
+  pubKeyHash: number;
+  scriptHash: number;
+  wif: number;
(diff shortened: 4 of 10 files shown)


runn1ng:
When the file is in `index.js.flow`, flow will read it automatically and use it for typechecking.

runn1ng:
If you think it's a good idea, I will think about how to properly test this so it's up-to-date.

jprichardson:
I do really like the idea of Flow. My concern about these files is maintenance and becoming stale. But given that it's for 2.x, it's probably not a big deal since the interfaces wouldn't change (other than additions). @dcousens thoughts?

runn1ng:
Yep, I was thinking about that too. 

My idea was taking some example code and checking the errors, but then someone has to keep on checking that, etc. 

The code in current test cannot be used for that, since the fixtures are untyped (and there are purposeful errors anyway). 

Basically you would need some code, which you would need to:

* check that it's working in current bitcoin.js
* and produces some expected flow output (similar to how flow itself has tests)

dcousens:
I'm for this.
Does `flow` know how to find these files? (as in, is this the complete change required?)

runn1ng:
> is this the complete change required?
With regards to Flow and this repo, yes.

I will think about how to test this and prevent it from becoming stale, as was mentioned here

runn1ng:
I am experimenting with using the mocha tests also for testing if the `flow` definition is correct, it actually doesn't end up that badly, I will update this PR once I have something.

runn1ng:
OK, I have added some idea how to test that.

I have enabled typechecking for one of the tests, then added a travis test that checks if the error output of flow matches the expected output.

I wanted to touch the test code as least as possible, but I had to change the module imports somewhat.*

Tell me what you think and if this is a good way to go.

----

\* that happens because I have the `.js.flow`  in `src` only with `index`. I tried to make one for each file, but then I probably hit some bug within Flow with the re-exporting of modules in `.js.flow` files. So I left it like this.

runn1ng:
I don't want to push this too hard; so I instead made own repo that I will add to my projects that are using bitcoin.js

you are welcome to add PRs

https://github.com/runn1ng/bitcoinjs-flow

runn1ng:
Hm, I just now discovered facebook has made a repo of flow type definitions here.... I will add it there instead.

https://github.com/flowtype/flow-typed

dcousens:
I think both of those are great additions,  and would probably do better than us maintaining those definitions here.

runn1ng:
I described, hopefully, the whole library, and put the flow file here

https://github.com/flowtype/flow-typed/pull/211

you can check the definition file if it's correct if you wish (I am not sure about some parts)

dcousens:
@runn1ng,  once merged,  you should add a link to these in the README :+1:

PR #557, opened 2016-03-09
Use equal and reverse for buffer from bufferutils

fanatid:


dcousens:
The point was that we are always trying to remove/deprecate `bufferutils`.

dcousens:
@fanatid I'd rather you just inline `bufferReverse` operations,  not use `bufferutils`.
Make that change and we'll merge :+1: (didn't realise the performance difference was so significant).

fanatid:
@dcousens I didn't know that you trying remove `bufferutils`.
Why don't you add this point to https://github.com/bitcoinjs/bitcoinjs-lib/issues/508 ?

dcousens:
@fanatid it was a general push,  and not something that may be possible for `3.0.0`.

PR #575, opened 2016-04-27
ECPair cleanup

dcousens:


src/ecpair.js
```diff
@@ -58,20 +58,24 @@ ECPair.fromPublicKeyBuffer = function (buffer, network) {
...
-    var version = buffer[0]
-
     network = network.filter(function (network) {
       return version === network.wif
     }).pop()
 
     if (!network) throw new Error('Unknown network version')
+
+  // network
+  } else if (network) {
```
fanatid: `network` always will be truthly here?
why not `} else if (version !== network.wif) throw new Error('Invalid network version')` ?

dcousens: Oh true,  it gets defaulted.  Will change

dcousens: Fixed in ea4ce4e]

fanatid:
LGTM

PR #578, opened 2016-05-03
Add isCoinbase

dcousens:
@fanatid please merge after review [and ACK].

Reference from the [consensus rules](https://github.com/bitcoin/bitcoin/blob/86b800c6a299455580fe76e5fb43218f0222e179/src/main.cpp#L3262-L3267) and [core implementation](https://github.com/bitcoin/bitcoin/blob/a9c8b744e8875785ff9a3768df3517b2d9ed7a36/src/primitives/transaction.h#L269-L272).


dcousens:
I don't think `2.3.0` will be released for a small while yet,  so should give us time to double check the name etc.
Seems solid,  and its a good reference to the core implementation too.

fanatid:
LGTM
@dcousens do not you want to change `isCoinbaseHash`?
```js
Transaction.isCoinbaseHash = function (buffer) {
  typeforce(types.Buffer, buffer)
  if (buffer.length !== 32) return false
  for (var i = 0; i < 32; ++i) {
    if (buffer[i] !== 0) return false
  }
  return true
}
```
Of course is not readable as `Array.every`, but faster.

dcousens:
@fanatid maybe a separate PR to be honest

dcousens:
Tests all passed,  a few of integration tests failed due to timeouts.   I'm going to merge.

PR #579, opened 2016-05-03
Update Transaction.isCoinbaseHash

fanatid:
- add argument type-checking
- replace Array.every on for-loop


fanatid:
Q: should `isCoinbaseHash` return `false` for `Buffer` with length not equal 32?

dcousens:
> Q: should isCoinbaseHash return false for Buffer with length not equal 32?

That is a really good question.

dcousens:
I think this isn't an exception circumstance,  its simply a lack of an adequate type system.  Therefore, in this case, I think the use of `typeforce` is more than adequate.

PR #580, opened 2016-05-04
Block: add checkMerkleRoot/calculateMerkleRoot

dcousens:


src/block.js
```diff
@@ -133,6 +135,44 @@ Block.calculateTarget = function (bits) {
...
+      var m = Math.min(i - 1, j + 1)
+      var hash1 = hashes[j + k]
+      var hash2 = hashes[k + m]
+
+      var hasher = createHash('sha256')
+      hasher.update(hash1)
+      hasher.update(hash2)
+
+      // TODO: use less memory
+      hashes.push(bcrypto.sha256(hasher.digest()))
```
dcousens: @fanatid thoughts on what we can do here?
AFAIK this is `N!` space complexity right now,  whereas it could probably just be `N`.
Thoughts?

test/block.js
```diff
@@ -97,6 +97,22 @@ describe('Block', function () {
...
+      var block
+
+      beforeEach(function () {
+        block = Block.fromHex(f.hex)
+      })
+
+      if (f.hex.length === 160) return
+
+      it('returns ' + f.valid + ' for ' + f.id, function () {
+        assert.strictEqual(block.checkMerkleRoot(), true)
```
dcousens: Needs negative fixtures

src/block.js
```diff
@@ -1,9 +1,11 @@
+var createHash = require('create-hash')
 var bufferutils = require('./bufferutils')
 var bcrypto = require('./crypto')
 var bufferCompare = require('buffer-compare')
 var bufferReverse = require('buffer-reverse')
 
 var Transaction = require('./transaction')
+var NULL_HASH = new Buffer(32).fill(0)
```
fanatid: `fill` return `undefined` in node 0.10

fanatid:
@dcousens I think this intersects with https://github.com/ethers/bitcoin-proof
maybe we should move `bitcoin-proof` to bitcoinjs org? if @ethers don't mind :)

src/block.js
fanatid: Why `N!`? I see `O(N)` hashes here..

dcousens: Actually... never mind.

dcousens:
@fanatid perhaps `NULL_HASH` should just be `null`?  It doesn't really make sense as a `Buffer` now that I think about it...
A merkle root of nothing is nothing,  not `0000000000000000000000000000...`

dcousens:
@fanatid the implementation in `bitcoin-proof` is very convoluted by comparison,  mixed encodings etc.
It also has completely different data structure expectations.

I don't think it'd be reasonable to use that package without major changes tbh.

fanatid:
> @fanatid perhaps `NULL_HASH` should just be `null`?  It doesn't really make sense as a `Buffer` now that I think about it...
A merkle root of nothing is nothing,  not `0000000000000000000000000000...`

Maybe throw error when `transactions.length` equal to zero? It's impossible situation.

fanatid:
@dcousens we can change `bitcoin-proof` for one style with `bitcoinjs` :) There isn't much changes should be.

jprichardson:
> @dcousens we can change bitcoin-proof for one style with bitcoinjs :) There isn't much changes should be.

If we can avoid NIH, the better IMHO.

dcousens:
> Maybe throw error when transactions.length equal to zero? It's impossible situation.

Agreed,  will change.

@jprichardson I don't consider this a NIH issue,  it was simply a matter of concise code.
`bitcoin-proof` also seems quite an odd name for something that should probably just be just a generic merkle tree package.

dcousens:
>  If we can avoid NIH, the better IMHO.

I don't think this is a NIH issue tbh,  I'm not against using external/pre-written modules?
I'd just rather not have to 'adapt' to them via unnecessary allocations of objects/data structures.

fanatid:
@dcousens I know yet one module for merkle proof: https://github.com/mappum/bitcoin-merkle-proof (supports proof for bip37)

dcousens:
@fanatid but why is it `bitcoin-` at all,  merkle trees are nothing new,  and if we don't use a `bitcoin-` specific package, just one that operates purely on `Buffer`,  we can avoid any hazards in API wrappers/adapters.

src/block.js
```diff
@@ -133,6 +136,43 @@ Block.calculateTarget = function (bits) {
...
+      hasher.update(hash2)
+
+      hashes.push(bcrypto.sha256(hasher.digest()))
+    }
+
+    k += i
+    i = (i + 1) >> 1
+  }
+
+  return hashes[hashes.length - 1]
```
fanatid: Should be better:
```js
var hashes = transactions.map(function (transaction) {
  return transaction.getHash()
})

for (var length = hashes.length; length > 1; hashes = (hashes / 2) | 0) {
  for (var i = 0, j = 1, s = 0; j < length; i += 2, j += 2, ++s) {
    var hasher = createHash('sha256')
    hasher.update(hashes[i])
    hasher.update(hashes[j === length ? i : j])
    hashes[s] = bcrypto.sha256(hasher.digest())
  }
}

return hashes[0]
```

dcousens: `TypeError: Data must be a string or a buffer` for `hasher.update(hashes[i])`,  probably a bad index.

ethers:
@fanatid @dcousens FWIW If relevant, I do not mind moving bitcoin-proof to here, and whatever style, name, and other changes desired.  (It was named that way to help make it more discoverable.)

dcousens:
@ethers nothing against your package mate!  It's great,  I'm just thinking out loud that we should probably generalize the concept and then use that instead.

I'd love to use merkle trees and proofs for other applications too.   Maybe even in `common-blockchain` @jprichardson

src/block.js
fanatid: Sorry, I not tested snippet, just concept. I think this wrong: `; j < length;`, `i` should be here.

dcousens: >  hashes = (hashes / 2) | 0) {

dcousens: Still doesn't work :(,  working through it now

dcousens:
ping @fanatid check 007d2a7,  reworked from yours,  thanks for the hint :)

src/block.js
```diff
@@ -1,9 +1,12 @@
+var createHash = require('create-hash')
 var bufferutils = require('./bufferutils')
 var bcrypto = require('./crypto')
 var bufferCompare = require('buffer-compare')
 var bufferReverse = require('buffer-reverse')
 
 var Transaction = require('./transaction')
+var NULL_HASH = new Buffer(32)
+NULL_HASH.fill(0)
```
fanatid: `NULL_HASH` is unused now

dcousens: fixed

src/block.js
```diff
@@ -133,6 +136,35 @@ Block.calculateTarget = function (bits) {
   return target
 }
 
+Block.calculateMerkleRoot = function (transactions) {
+  if (length === 0) throw TypeError('Cannot compute merkle root for zero transactions')
```
fanatid: this works? Looks very wired..
```js
var length = transactions.length
if (length === 0) ..
```

dcousens:
@fanatid if you're happy with latest, please merge :)
(making a new issue for extracting this logic)

src/block.js
dcousens: Indeed, we don't have a test for that exception yet... good catch!

dcousens: added a test in afb6533

dcousens:
@fanatid will squash if needed

fanatid:
LGTM, please squash and I merge

dcousens:
@fanatid squashed

Issue #587, opened 2016-06-08
Release 2.3.0


dcousens:
Related https://github.com/bitcoinjs/bitcoinjs-lib/issues/583#issuecomment-224569662

PR #589, opened 2016-06-08
Fix for Webpack

jprichardson:
It's been reported to me by @shea256 that `bitcoinjs-lib` doesn't work with Webpack even using the Webpack JSON loader. Apparently it's because the Webpack needs the `.json` extension in `require` statements so that it knows to use the JSON loader.

### DIFF
--- a/src/bufferutils.js
+++ b/src/bufferutils.js
@@ -1,4 +1,4 @@
-var opcodes = require('./opcodes')
+var opcodes = require('./opcodes.json')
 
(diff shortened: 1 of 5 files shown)


shea256:
\o/ Thanks for making this update @jprichardson. It addresses the issues I've been having.

dcousens:
:+1:

dcousens:
Relevant to https://github.com/bitcoinjs/bitcoinjs-lib/pull/587

PR #591, opened 2016-06-14
index: add .network default for bitcoin

dcousens:
Implementation is a bit lame,  but,  thoughts on the concept?  @jprichardson ?
Example that shows my motivation:

``` js
var bitcoin = require('bitcoinjs-lib')

// ...

console.log(bitcoin.network.pubKeyHash)
// or maybe console.log(bitcoin.network.constants.pubKeyHash)
console.log(bitcoin.networks.bitcoin.pubKeyHash)
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -11,6 +11,7 @@ module.exports = {
   crypto: require('./crypto'),
   message: require('./message'),
   networks: require('./networks'),
+  network: require('./networks').bitcoin,
   opcodes: require('./opcodes.json'),
   script: require('./script')
 }


jprichardson:
Even though it's a small change, it's one more public thing to maintain and could break people's code with a move to 3.0. What are your thoughts on 'why' this is needed?

dcousens:
@jprichardson it just felt really clunky to write: `bitcoin.networks.bitcoin.pubKeyHash` (for a **bitcoin** library).
Maybe the problem lies elsewhere though.

Issue #565, opened 2016-04-08
Use `varuint-bitcoin` over inline implementation


dcousens:
Might be worthy of investigation.
Reference: https://github.com/fanatid/varuint-bitcoin

ping @fanatid

dcousens:
Closing in favour of #595

PR #595, opened 2016-06-17
use varuint-bitcoin

dcousens:
Tests are failing due to different error messages.
Not sure if its a breaking change or not,  tagging as such for now until we discuss otherwise.

Closes https://github.com/bitcoinjs/bitcoinjs-lib/issues/565

### DIFF
--- a/src/bufferutils.js
+++ b/src/bufferutils.js
@@ -1,12 +1,5 @@
 var opcodes = require('./opcodes.json')
-
-// https://github.com/feross/buffer/blob/master/index.js#L1127
-function verifuint (value, max) {
-  if (typeof value !== 'number') throw new Error('cannot write a non-number as a number')
-  if (value < 0) throw new Error('specified a negative value for writing an unsigned value')
-  if (value > max) throw new Error('value is larger than maximum value for type')
-  if (Math.floor(value) !== value) throw new Error('value has a fractional component')
-}
+var varuint = require('varuint-bitcoin')
 
 function pushDataSize (i) {
   return i < opcodes.OP_PUSHDATA1 ? 1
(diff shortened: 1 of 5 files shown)


src/bufferutils.js
```diff
@@ -174,8 +119,8 @@ module.exports = {
   readUInt64LE: readUInt64LE,
   readVarInt: readVarInt,
   reverse: require('buffer-reverse'),
-  varIntBuffer: varIntBuffer,
-  varIntSize: varIntSize,
+  varIntBuffer: varuint.encode,
+  varIntSize: varuint.encodingLength,
```
dcousens: @fanatid its worth asking if this should be included.
I'd say these functions are non-generic enough that it makes sense to keep them in this library.

dcousens:
ping @fanatid

jprichardson:
> ping @fanatid 

FYI. I think he's gone until late July.

src/bufferutils.js
fanatid: I think they should be included for backward compatibility (at least while 3.x not be released).

dcousens: This PR is breaking anyway, I meant including them in `3.x.`

dcousens:
Ready to merge

dcousens:
Not a breaking change, I've kept the functions internally

dcousens:
https://travis-ci.org/bitcoinjs/bitcoinjs-lib/jobs/165689761 was the only failed job,  with the reason:

>  1) bitcoinjs-lib (advanced) can create an OP_RETURN transaction:
     Error: timeout of 20000ms exceeded. Ensure the done() callback is being called in this test.

I'll consider extending the timeouts I suppose.

Issue #436, opened 2015-08-13
Add Transaction.hashForSignature tests


dcousens:

dcousens:
Closing favour of #598

PR #598, opened 2016-06-22
Add Transaction.hashForSignature tests (and OP_CODESEPARATOR)

dcousens:
Resolves https://github.com/bitcoinjs/bitcoinjs-lib/issues/436 and finally resolves an inline TODO to add tests for `hashForSignature`.
Granted, it was already under significant integration tests and the Bitcoin core tests,  but we didn't have anything further than that.

I also added 697f2da as a means of cleaning up the code which previously was not-so-straight forward to read.

@jprichardson / @rubensayshi / @fanatid please review :)


jprichardson:
utACK

dcousens:
Rebasing then will merge.

PR #599, opened 2016-06-22
TxBuilder: add setVersion

dcousens:
I wasn't sure about this function at first,  as it seemed so pass-through'ish (aka, perhaps a a sign of bad design).

But with BIP68 about to come into effect,  it may be worth having some sanity checking logic here.  We'll see.

(aka, `setVersion(1)` for a transaction the CSV OP codes is effectively invalid AFAIK)

### DIFF
--- a/src/transaction_builder.js
+++ b/src/transaction_builder.js
@@ -163,6 +163,13 @@ TransactionBuilder.prototype.setLockTime = function (locktime) {
   this.tx.locktime = locktime
 }
 
+TransactionBuilder.prototype.setVersion = function (version) {
+  typeforce(types.UInt32, version)
+
+  // XXX: this might eventually become more complex depending on what the versions represent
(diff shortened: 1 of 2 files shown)


jprichardson:
Would it be more appropriate to version the scripts instead of the tx?

dcousens:
@jprichardson well,  no?  The `Transaction` is what *has* a version,  but,  at least for CSV,  it will be contents of the script which dictate what version of the transaction is required...  so I'd say we just work within those constraints,  if any of the scripts contain CSV,  `setVersion` must have been set to `2`,  and we just maintain that invariant?

jprichardson:
> so I'd say we just work within those constraints, if any of the scripts contain CSV, setVersion must have been set to 2, and we just maintain that invariant?

Yep, that sounds good.

LGTM.

PR #603, opened 2016-07-12
Transaction: allow parameterizable buffer

dcousens:


jprichardson:
utACK. Would you add a test?

dcousens:
Going to put this in https://github.com/bitcoinjs/bitcoinjs-lib/pull/614,  so will add test today

fanatid:
utACK

src/transaction.js
```diff
@@ -304,6 +304,9 @@ Transaction.prototype.toBuffer = function () {
 
   writeUInt32(this.locktime)
 
+  // avoid slicing unless necessary
+  if (initialOffset !== undefined) return buffer.slice(initialOffset, offset)
```
dcousens: this will help in node environments,  but in the browser,  will this still result in a copy anyway?

dcousens:
ping @jprichardson @fanatid ,  see comments

src/transaction.js
fanatid: I don't understand, for what purpose are you using `slice` here?
In browser, `buffer` uses `subarray` for `slice`, i.e. if you change sliced array -- parent array also will be changed.

dcousens: Ok,  then not a concern!

dcousens: > I don't understand, for what purpose are you using slice here?

To encode the amount of data actually encoded into the `Buffer`.

dcousens:
So @fanatid this is good to go then?

fanatid:
I think yes.
ACK

dcousens:
ping @jprichardson

jprichardson:
It LGTM, but I should have originally asked the most important question... why?

dcousens:
@jprichardson performance!  Noticed some fun things with lots of large transactions,  saved myself quite a few `Buffer` allocations using this method.

PR #613, opened 2016-08-04
README: fix brainwallet url

hisener:


dcousens:
To be honest,  since they are still using `v0.1.3`,  which is highly out-dated and missing a lot of improvements (even bug fixes),  I'm not sure if we should continue saying they are actually using *this* library.

Thoughts @jprichardson ?

hisener:
We can delete it or update URL.

I have also checked other projects.

jprichardson:
Agreed @dcousens.

hisener:
@jprichardson deleted.

dcousens:
Thanks @hisener :+1:

PR #617, opened 2016-08-31
bufferutils: fix pushDataInt output, BIP62 compliance

dcousens:
Fixes an edge case where `OP_PUSHDATA2` was used instead of `OP_PUSHDATA1` when the data size was `255`.

_This is a bug in terms of BIP62 minimal push data compliance,  which is a `IsStandard` relay policy._


dcousens:
This makes us tested and compliant against BIP62,  which is currently implemented for MINIMAL DATA PUSH.

dcousens:
Ping @kristovatlas, heh

dcousens:
I wonder if the repo https://github.com/bitcoinjs/bip62 is worth it though.
It only checks minimal data push at the moment...

Brings back into question what this repository should do in most cases... adhere to `IsStandard` strictly only?

We have several cases now where this impacts us:

`isPubKeyHashInput` checks for compliance via `IsCanonicalPubKey`

fanatid:
utACK

dcousens:
Read back to https://github.com/bitcoinjs/bitcoinjs-lib/pull/481,  I think we'll maintain the approach that we should be *atleast* able to decode any content in any block past the BIP66 soft-fork.

@fanatid its likely we should probably support decode any content from *any block* at any point.  But we'd need to re-import ECDSA code at this point,  so I'd rather just move on to the alternative ECDSA library first.

dcousens:
Our in-repository `IsStandard` compliance should be consistent always in creation,  but in terms of import/decode it is not limited to `IsStandard` compliance.

`script.is*` functions are all standards compliant,  by their nature.

dcousens:
Moved `BIP62` repository to https://github.com/bitcoinjs/minimaldata,  as its not really "just" BIP62,  its more related to bitcoin-core policy.

Issue #619, opened 2016-09-12
HDNode with consistent and quick serialization


runn1ng:
Hello,

is there a way to serialize/deserialize HDNode to a JavaScript object in a way, that's efficient?

We are now saving HDNodes in MyTrezor using toBase58 and HDNode.fromBase58; however, we found this to be quite slow, especially if the user has more such HDNodes (one per account).

I am now experimenting with function like this, but it feels too "hacky" and any internal implementation change might break something. However, the speed is immediately obvious. (We *do* know that the HDNode is neutered.)

````javascript
function serializeBitcoinJsHDPubNode(node) {
    return {
        depth: node.depth,
        index: node.index,
        parentFingerprint: node.parentFingerprint,
        chainCode: node.chainCode,
        Q: node.keyPair.Q.getEncoded(false),
    };
}

function deserializeBitcoinJsHDPubNode(node) {
    const depth = node.depth;
    const index = node.index;
    const parentFingerprint = node.parentFingerprint;
    const chainCode = new Buffer(node.chainCode);
    const Q = ecurve.Point.decodeFrom(bcurve, new Buffer(node.Q));
    const res = new HDNode(new ECPair(null, Q), chainCode);
    res.depth = depth;
    res.index = index;
    res.parentFingerprint = parentFingerprint;
    return res;
}
````

runn1ng:
I might have said, the speed is probably more important for the *de*serialization...

dcousens:
While the performance of these functions leaves so much to be desired,  and is easily resolved by https://github.com/bitcoinjs/bitcoinjs-lib/issues/610 (aka, moving to a faster ECDSA library),  there are other tricks you can do.

Lazy evaluation is by far the simplest path through this.
However,  it would be much simpler for you to just save the X/Y `bigi` coordinates of the `ECPoint` rather than re-calculating (`ecurve.Point.decodeFrom`) them each time.
This is the only real operation that takes any time.

dcousens:
As simply,  you could just use the non-compressed versions of the public key.  That would be as performant as just a `new Buffer` operation IIRC.

dcousens:
> I am now experimenting with function like this, but it feels too "hacky" and any internal implementation change might break something.

Understandable,  and I apologise we haven't moved on #610 sooner to make this not worth your time,  that said,  you can assume these are part of the public API and will adhere to SEMVER.

runn1ng:
[...]

PR #620, opened 2016-09-13
HDNode - Skipping validation when requested

runn1ng:
A quick "fix" for  https://github.com/bitcoinjs/bitcoinjs-lib/issues/619

I haven't added any tests, since the functionality should be exactly the same with valid xpubs, so I am not sure what would be tested there.

### DIFF
--- a/src/hdnode.js
+++ b/src/hdnode.js
@@ -51,7 +51,7 @@ HDNode.fromSeedHex = function (hex, network) {
   return HDNode.fromSeedBuffer(new Buffer(hex, 'hex'), network)
 }
 
-HDNode.fromBase58 = function (string, networks) {
+HDNode.fromBase58 = function (string, networks, skipValidation) {
   var buffer = base58check.decode(string)
   if (buffer.length !== 78) throw new Error('Invalid buffer length')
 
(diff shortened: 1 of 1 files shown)


src/hdnode.js
```diff
@@ -51,7 +51,7 @@ HDNode.fromSeedHex = function (hex, network) {
   return HDNode.fromSeedBuffer(new Buffer(hex, 'hex'), network)
 }
 
-HDNode.fromBase58 = function (string, networks) {
+HDNode.fromBase58 = function (string, networks, skipValidation) {
```
dcousens: I don't like adding arguments like this because it can cause weird results if people have bound these functions in a way that might currently ignore the extra arguments (`.forEach(fromBase58)` or something),

But,  I can't currently see another way around this...

src/hdnode.js
```diff
@@ -109,9 +109,12 @@ HDNode.fromBase58 = function (string, networks) {
...
     if (!Q.compressed) throw new Error('Invalid public key')
 
-    // Verify that the X coordinate in the public point corresponds to a point on the curve.
-    // If not, the extended public key is invalid.
-    curve.validate(Q)
+    // Skip validation if requested for efficiency
+    if (!skipValidation) {
+      // Verify that the X coordinate in the public point corresponds to a point on the curve.
+      // If not, the extended public key is invalid.
+      curve.validate(Q)
```
dcousens: `curve.isInfinity(Q)` could still be run,  that is practically instant.

src/hdnode.js
runn1ng: Yeah I understand. I could have added `function (string, options)`, and if `options` is a JS object take `options.network` and `options.skipValidation`; and if it is not, take it as network array.... but that could maybe break if `networks` is in fact just one network... and it sounds hacky.

Yeah I am not sure what to do with it.

src/hdnode.js
runn1ng: ok! I will add that

dcousens:
I think I'd rather see #623 finished.
@runn1ng are you OK with that?  You should be able to use your fork until we fix that more 'upstream' issue...

runn1ng:
OK!

PR #629, opened 2016-09-27
types: add Satoshi

dcousens:


dcousens:
Might be worth making this a separate library.  But I use this type all the time in my own projects...  probably worth changing some of our `UInt53` types over to it tbh.

dcousens:
Failing due to https://github.com/bitcoinjs/bitcoinjs-lib/issues/631
Good to merge.

PR #630, opened 2016-09-27
tests: fix random failures due to non-minimal push

dcousens:


dcousens:
Actually... this should really just be fixed. 
See https://github.com/bitcoinjs/bitcoinjs-lib/issues/631 instead.

Issue #633, opened 2016-09-27
TransactionBuilder.addInput duplicate txId needs tests


dcousens:

dcousens:
Resolved in https://github.com/bitcoinjs/bitcoinjs-lib/pull/637

dcousens:
As shown in 9b62ee4,  this actually wasn't working anyway for `vout:0`!

Issue #631, opened 2016-09-27
Buffer's of length 1, with OP_1 <= data[0] <= OP_16,  should compile to OP_1 to OP_16 instead of Buffer


dcousens:
This is per our minimal push adherence policy for script compilation.

dcousens:
Closing in favour of discussion in #638

PR #634, opened 2016-09-27
TransactionBuilder cleanup

dcousens:
Just a few small things.
Noticed the error message `'Transaction is already an input'` was quite odd,  you can add the same transaction many times,  but not the same transaction out (tx:vout).

Opened https://github.com/bitcoinjs/bitcoinjs-lib/issues/633 to resolve that.


dcousens:
Tests only failing sporadically due to #631... going to merge within 24 hours.

fanatid:
utACK

PR #637, opened 2016-09-27
TransactionBuilder: major internal refactor

dcousens:
This PR restructures `TransactionBuilder`s use of `extractInput` (now named `expandInput`) and `extractFromOutputScript` such they are used consistently,  by value,  and without duplicate handling.

### DIFF
--- a/src/transaction_builder.js
+++ b/src/transaction_builder.js
@@ -225,6 +225,40 @@ function prepareInput (input, kpPubKey, redeemScript, hashType) {
   input.hashType = hashType
 }
 
+function fixMultisigOrder (input, transaction, vin) {
+  var hashScriptType = input.redeemScriptType || input.prevOutType
+  if (hashScriptType !== 'multisig') throw new TypeError('Expected multisig input')
+
+  var hashType = input.hashType || Transaction.SIGHASH_ALL
+  var hashScript = input.redeemScript || input.prevOutScript
+
+  // maintain a local copy of unmatched signatures
+  var unmatched = input.signatures.concat()
+  var signatureHash = transaction.hashForSignature(vin, hashScript, hashType)
+
+  input.signatures = input.pubKeys.map(function (pubKey, y) {
+    var keyPair = ECPair.fromPublicKeyBuffer(pubKey)
+    var match
+
+    // check for a signature
+    unmatched.some(function (signature, i) {
+      // skip if undefined || OP_0
+      if (!signature) return false
+
+      // skip if signature does not match pubKey
+      if (!keyPair.verify(signatureHash, signature)) return false
+
+      // remove matched signature from unmatched
+      unmatched[i] = undefined
+      match = signature
+
+      return true
+    })
+
+    return match || undefined
+  })
+}
+
 function TransactionBuilder (network) {
   this.prevTxMap = {}
   this.network = network || networks.bitcoin
@@ -280,7 +314,7 @@ TransactionBuilder.fromTransaction = function (transaction, network) {
       if (!input.pubKeys || !input.signatures) return
       if (input.pubKeys.length === input.signatures.length) return
 
-      txb.__fixMultisigOrder(transaction, i)
+      fixMultisigOrder(input, transaction, i)
     }
   })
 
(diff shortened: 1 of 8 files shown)


test/fixtures/transaction_builder.json
```diff
@@ -632,7 +632,7 @@
       },
       {
         "description": "Incomplete transaction w/ prevTxScript defined",
-        "exception": "Transaction is missing signatures",
+        "exception": "Not enough signatures provided",
```
dcousens: > Transaction is missing signatures

To be honest,  this is actually an error case where something has gone wrong internally.

`.signatures` is now always defined if the number of signatures required is known.
Therefore,  its a matter of "not enough" or "too many signatures" instead.

src/transaction_builder.js
```diff
@@ -334,43 +293,56 @@ var canBuildTypes = {
   'pubkeyhash': true
 }
 
-function buildFromInputData (input, scriptType, parentType, redeemScript, allowIncomplete) {
+function buildFromInputData (input, scriptType, allowIncomplete) {
```
dcousens: This entire function could probably take more from `CombineSignatures`, https://github.com/bitcoin/bitcoin/blob/d612837814020ae832499d18e6ee5eb919a87907/src/script/sign.cpp#L309

Though we'd also need to modify `TransactionBuilder.sign`

src/transaction_builder.js
```diff
@@ -437,65 +405,101 @@ TransactionBuilder.prototype.sign = function (index, keyPair, redeemScript, hash
...
+      input.redeemScriptType = expanded.scriptType
+      input.signatures = expanded.signatures
+
+    // no redeemScript
     } else {
       // pay-to-scriptHash is not possible without a redeemScript
       if (input.prevOutType === 'scripthash') throw new Error('PrevOutScript is P2SH, missing redeemScript')
 
-      // if we don't have a scriptType, assume pubKeyHash otherwise
-      if (!input.scriptType) {
```
dcousens: `.scriptType` was some weird half-half of `prevOutType` and `redeemScriptType`,  it was entirely unnecessary.

src/transaction_builder.js
```diff
@@ -437,65 +405,101 @@ TransactionBuilder.prototype.sign = function (index, keyPair, redeemScript, hash
...
 
     input.signatures[i] = keyPair.sign(signatureHash)
-
     return true
   })
 
   if (!valid) throw new Error('Key pair cannot sign for this input')
 }
 
+TransactionBuilder.prototype.__fixMultisigOrder = function (transaction, vin) {
```
dcousens: This could be `function fixMultisigOrder (vin) {` if I did a iteration in `fromTransaction`... but maybe another time

dcousens:
The motivation for this PR is mostly to ensure that the internals are consistent for SegWit integration.

dcousens:
@rubensayshi @fanatid I'm not sure how in-depth you guys want to review this.
b3fd50ffd62a8921446e91309c052ecf4c8501cc is the meat of it.

src/transaction_builder.js
```diff
@@ -243,51 +209,44 @@ TransactionBuilder.prototype.addInput = function (txHash, vout, sequence, prevOu
...
+  if (!input.prevOutScript && prevOutScript) {
+    var prevOutScriptChunks = bscript.decompile(prevOutScript)
+    var prevOutType = bscript.classifyOutput(prevOutScriptChunks)
+
+    if (!input.pubKeys && !input.signatures) {
+      var expanded = expandOutput(prevOutScript)
+      if (expanded) {
+        input.pubKeys = expanded.pubKeys
+        input.signatures = expanded.signatures
+      }
```
dcousens: Its a good question as to whether `prevOutScript` should even be considered for expansion if a non-empty `scriptSig` was provided (aka,  with a redeemScript in tow).

dcousens:
Preliminary SegWit support could be as simple as forcing users to specify `SegWit` inputs w/ a `prevOutScript`,  in every other way it'd act like similar to a `scripthash` input.

`build` wouldn't change much,  but it would use `setWitnessScript` instead of `setInputScript` for witness `prevOutTypes`.
The rest is just `Transaction.toBuffer` changes...

dcousens:
Really want to merge this before end of week if possible

dcousens:
@fanatid please POST ACK

PR #638, opened 2016-09-29
script: always compile in a minimaldata compliant way

dcousens:
I can't decide if this is the right thing to do.
On the one hand,  I can't imagine another work flow that isn't a pain in the ass,  but at the same time it means `decompile(compile(x)) !== x` in more cases.

Existing exceptions already exist around non-standard data pushes... so I suppose that makes it more acceptable?


test/script.js
```diff
@@ -33,21 +33,27 @@ describe('script', function () {
...
   })
 
-  describe('compile', function () {
+  describe('compile (via fromASM)', function () {
     fixtures.valid.forEach(function (f) {
       if (f.scriptSig) {
         it('(' + f.type + ') compiles ' + f.scriptSig, function () {
           var scriptSig = bscript.fromASM(f.scriptSig)
 
-          assert.strictEqual(bscript.compile(scriptSig).toString('hex'), f.scriptSigHex)
```
dcousens: the explicit `compile` here was actually a double up... removed

test/script.js
```diff
@@ -59,14 +65,25 @@ describe('script', function () {
...
 
+          assert.strictEqual(bscript.compile(chunks).toString('hex'), f.scriptSigHex)
           assert.strictEqual(bscript.toASM(chunks), f.scriptSig)
+
+          if (f.nonstandard) {
+            var chunksNS = bscript.decompile(new Buffer(f.nonstandard.scriptSigHex, 'hex'))
+
+            assert.strictEqual(bscript.compile(chunksNS).toString('hex'), f.scriptSigHex)
+
+            // toASM converts verbatim, only `compile` transforms the script to a minimalpush compliant script
```
dcousens: and here lies some more confusing aspects of this change

dcousens:
Fixes https://github.com/bitcoinjs/bitcoinjs-lib/issues/631

dcousens:
ping @fanatid, @jprichardson, @rubensayshi, @runn1ng, @Sjors

dcousens:
We probably need more non-standard script tests anyway...

dcousens:
any comments?

dcousens:
This is in line with https://github.com/bitcoinjs/bitcoinjs-lib/issues/640

PR #639, opened 2016-10-05
stealth: more explicit inner variable names

dcousens:


dcousens:
ping @rubensayshi, seems to be an issue with testnet issued unspents again... merging as all tests pass if it weren't for that

PR #645, opened 2016-10-05
Fixing issues with prepareInput

runn1ng:


dcousens:
`.pubKeys` shouldn't be empty,  that is true.  But `.signatures` could be

dcousens:
Oh wait no, you're right... it should match.

dcousens:
That check *shouldn't* be necessary though... it means some other assumption didn't hold from the `expandInput` failed... I'll put more checks in there instead :+1:

src/transaction_builder.js
```diff
@@ -206,13 +206,15 @@ function prepareInput (input, kpPubKey, redeemScript, hashType) {
     input.signatures = expanded.signatures
 
   // maybe we have some prior knowledge
-  } else if (input.prevOutType) {
+  } else if (input.prevOutType && input.prevOutType !== 'pubkeyhash') {
```
dcousens: This seems wrong... `pubkeyhash` should be fine.

dcousens:
Closing in favour of https://github.com/bitcoinjs/bitcoinjs-lib/pull/647

PR #646, opened 2016-10-06
package: update to typeforce 1.8.7

dcousens:
Uses `typeforce.BufferN` and `UInt*` functions over our own,  better tested although the exception message for a 256-bit Buffer is slightly different.
IMHO not a breaking change,  but if someone was directly matching the error message,  it could break their flow.

It also fixes the error messages for function arguments checking which previously omitted the positional information.

### DIFF
--- a/test/fixtures/script.json
+++ b/test/fixtures/script.json
@@ -338,25 +338,25 @@
     ],
     "pubKeyHashOutput": [
       {
-        "exception": "Expected 160-bit Buffer, got 16-bit Buffer",
+        "exception": "Expected Buffer\\(Length: 20\\), got Buffer\\(Length: 2\\)",
         "hash": "ffff"
       }
     ],
     "scriptHashOutput": [
       {
-        "exception": "Expected 160-bit Buffer, got 24-bit Buffer",
+        "exception": "Expected Buffer\\(Length: 20\\), got Buffer\\(Length: 3\\)",
         "hash": "ffffff"
       }
     ],
     "witnessPubKeyHashOutput": [
       {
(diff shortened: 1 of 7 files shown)


test/fixtures/script.json
```diff
@@ -338,25 +338,25 @@
...
     "witnessPubKeyHashOutput": [
       {
-        "exception": "Expected 160-bit Buffer, got 24-bit Buffer",
+        "exception": "Expected Buffer\\(Length: 20\\), got Buffer\\(Length: 3\\)",
         "hash": "ffffff"
       }
     ],
     "witnessScriptHashOutput": [
       {
-        "exception": "Expected 256-bit Buffer, got 24-bit Buffer",
```
dcousens: If anyone considers this a breaking change,  please speak up.   It just means we'd roll our own wrappers until the next major version.

dcousens:
@jprichardson are exception messages part of the public API?  They are and aren't in mind... so I'm not sure.

jprichardson:
> are exception messages part of the public API? They are and aren't in mind... so I'm not sure.

I don't think so. I think that's why node sets a `code` property and if that were to change, that'd break the API. While there is some risk to changing the message, I think it's an acceptable risk. Not sure it's worth it to add codes yet to this library.

dcousens:
@jprichardson ultimately these are more for saving the developer from catastrophic fund-losing errors,  not necessarily application level errors... so I'd agree, a `code` field is wrong here.

dcousens:
All tests passed except a integration test timed out.

Issue #644, opened 2016-10-05
[question] Transaction builder not building


runn1ng:
I am trying to construct a transaction using TransactionBuilder and I am doing something wrong, but I can't tell what.

I have this transaction
https://testnet.blockexplorer.com/tx/0eb90544be26a25942ee61bbaea5e990d0290895cd0ba94127432a0c930b5775

I want to use the first output as an input.

I am doing this:

````javascript
const privkey = 'private key, belonging to address mowxUGLvamMSYyCUMMzChxgBosUQDvsAMk';
const privkeypair = bitcoin.ECPair.fromWIF(privkey, bitcoin.networks.testnet);
const txhex = 'transaction in hex';
const transaction = bitcoin.Transaction.fromHex(txhex);
const builder = new bitcoin.TransactionBuilder(bitcoin.networks.testnet);
builder.addInput(transaction, 0, 0);
builder.addOutput("n4pSwWQZm8Wter1wD6n8RDhEwgCqtQgpcY", 6000);
builder.sign(0, privkeypair);
````

And I get this error (using latest in npm)

````
Error: pubkeyhash not supported
    at TransactionBuilder.sign (.../node_modules/bitcoinjs-lib/src/transaction_builder.js:482:56)
    (my file here)
````

Or this error, using master

````
node_modules/bitcoinjs-lib/src/transaction_builder.js:482
  if (!valid) throw new Error('Key pair cannot sign for this input')
              ^

Error: Key pair cannot sign for this input
    at TransactionBuilder.sign (.../node_modules/bitcoinjs-lib/src/transaction_builder.js:482:21)
    (my file here)
````

What am I doing wrong?

runn1ng:
Well, let's try with this transaction instead, it's already spent, so I can show the private key.

https://testnet.blockexplorer.com/tx/393b227a1cb669b4277a693f0d8ee521e62ddce9ca4cdb97f0cc4fc735585932

`cV6Fs4jspcMmUoPXyQSAVkVDfNVjL2Sn8HsBwpDvmKdhF5A9CkQW` is the private key to `mjfVsKHRFQyFTFSqETGVLZf5NTjkEhVLhT`. The tx is already spent, but builder shouldn't notice. My code is:


````javascript
"use strict";

const bitcoin = require('bitcoin.js');

const privkey = 'cV6Fs4jspcMmUoPXyQSAVkVDfNVjL2Sn8HsBwpDvmKdhF5A9CkQW';
[...]

PR #647, opened 2016-10-06
TransactionBuilder: if prevOutScript is defined, but not signable, try again

dcousens:
This should have fixed https://github.com/bitcoinjs/bitcoinjs-lib/issues/644#issuecomment-251782056,  but I'm still getting issues.

None the less, we need to add some test cases for this,  as it is a valid case.

### DIFF
--- a/src/transaction_builder.js
+++ b/src/transaction_builder.js
@@ -191,15 +191,16 @@ function prepareInput (input, kpPubKey, redeemScript, hashType) {
 
       var prevOutScriptScriptHash = bscript.decompile(input.prevOutScript)[1]
       if (!prevOutScriptScriptHash.equals(redeemScriptHash)) throw new Error('Inconsistent hash160(RedeemScript)')
-
-    // or, we don't have a prevOutScript, so generate a P2SH script
-    } else {
-      input.prevOutScript = bscript.scriptHashOutput(redeemScriptHash)
-      input.prevOutType = 'scripthash'
     }
 
     var expanded = expandOutput(redeemScript, undefined, kpPubKey)
     if (!expanded.pubKeys) throw new Error('RedeemScript not supported "' + bscript.toASM(redeemScript) + '"')
+
+    // if we don't have a prevOutScript, generate a P2SH script
+    if (!input.prevOutType) {
+      input.prevOutScript = bscript.scriptHashOutput(redeemScriptHash)
+      input.prevOutType = 'scripthash'
+    }
 
     input.pubKeys = expanded.pubKeys
     input.redeemScript = redeemScript
--- a/src/transaction_builder.js
+++ b/src/transaction_builder.js
@@ -87,11 +87,13 @@ function expandInput (scriptSig, redeemScript) {
   }
 }
 
-function expandOutput (script, ourPubKey) {
+function expandOutput (script, scriptType, ourPubKey) {
   typeforce(types.Buffer, script)
 
   var scriptChunks = bscript.decompile(script)
-  var scriptType = bscript.classifyOutput(script)
+  if (!scriptType) {
+    scriptType = bscript.classifyOutput(scriptChunks)
+  }
 
   var pubKeys = []
 
@@ -113,7 +115,7 @@ function expandOutput (script, ourPubKey) {
       pubKeys = scriptChunks.slice(1, -2)
       break
 
-    default: return
+    default: return { scriptType: scriptType }
   }
 
   return {
@@ -196,8 +198,8 @@ function prepareInput (input, kpPubKey, redeemScript, hashType) {
       input.prevOutType = 'scripthash'
     }
 
-    var expanded = expandOutput(redeemScript, kpPubKey)
-    if (!expanded) throw new Error('RedeemScript not supported "' + bscript.toASM(redeemScript) + '"')
+    var expanded = expandOutput(redeemScript, undefined, kpPubKey)
+    if (!expanded.pubKeys) throw new Error('RedeemScript not supported "' + bscript.toASM(redeemScript) + '"')
 
     input.pubKeys = expanded.pubKeys
     input.redeemScript = redeemScript
@@ -209,8 +211,13 @@ function prepareInput (input, kpPubKey, redeemScript, hashType) {
     // pay-to-scriptHash is not possible without a redeemScript
     if (input.prevOutType === 'scripthash') throw new Error('PrevOutScript is P2SH, missing redeemScript')
 
-    // throw if we can't sign with it
-    if (!input.pubKeys || !input.signatures) throw new Error(input.prevOutType + ' not supported')
+    // try to derive the missing information about the script now that we
+    // have a kpPubKey
+    expanded = expandOutput(input.prevOutScript, input.prevOutType, kpPubKey)
+    if (!expanded.pubKeys) return
+
+    input.pubKeys = expanded.pubKeys
+    input.signatures = expanded.signatures
 
   // no prior knowledge, assume pubKeyHash
   } else {
@@ -356,19 +363,21 @@ TransactionBuilder.prototype.__addInputUnsafe = function (txHash, vout, sequence
 
   // derive what we can from the previous transactions output script
   if (!input.prevOutScript && prevOutScript) {
-    var prevOutScriptChunks = bscript.decompile(prevOutScript)
-    var prevOutType = bscript.classifyOutput(prevOutScriptChunks)
+    var prevOutType
 
     if (!input.pubKeys && !input.signatures) {
       var expanded = expandOutput(prevOutScript)
-      if (expanded) {
+
+      if (expanded.pubKeys) {
         input.pubKeys = expanded.pubKeys
         input.signatures = expanded.signatures
       }
+
+      prevOutType = expanded.scriptType
     }
 
     input.prevOutScript = prevOutScript
-    input.prevOutType = prevOutType
+    input.prevOutType = prevOutType || bscript.classifyOutput(prevOutScript)
   }
 
   var vin = this.tx.addInput(txHash, vout, sequence, scriptSig)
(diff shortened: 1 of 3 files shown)


src/transaction_builder.js
```diff
@@ -210,8 +212,18 @@ function prepareInput (input, kpPubKey, redeemScript, hashType) {
     // pay-to-scriptHash is not possible without a redeemScript
     if (input.prevOutType === 'scripthash') throw new Error('PrevOutScript is P2SH, missing redeemScript')
 
+    if (!input.pubKeys || input.pubKeys.length === 0) {
+      // maybe we can derive more information about the script
+      // now that we have a kpPubKey?
+      expanded = expandOutput(input.prevOutScript, input.prevOutType, kpPubKey)
+      input.pubKeys = expanded.pubKeys
+      input.signatures = expanded.signatures
```
dcousens: @runn1ng specifically,  this step was missing and is a simple result of you gave `TransactionBuilder` more information than usual (at the start),  but not enough to know if it could sign with `kpPubKey`... until you went to `sign` with it,  but just assumed that if `prevOutScript` was defined,  that it was "complete".
Which isn't true... and also opens up an interesting can of worms around what else could be "added" or "derived"... but this will do for now.

src/transaction_builder.js
```diff
@@ -447,7 +461,8 @@ TransactionBuilder.prototype.sign = function (vin, keyPair, redeemScript, hashTy
     input.prevOutScript !== undefined &&
     input.pubKeys !== undefined &&
     input.signatures !== undefined &&
-    input.signatures.length === input.pubKeys.length
+    input.signatures.length === input.pubKeys.length &&
+    input.pubKeys.length > 0
```
dcousens: if `input.pubKeys` is `0`, we should try to expand whatever we can for more information

test/fixtures/transaction_builder.json
```diff
@@ -285,6 +285,28 @@
...
+        "description": "Transaction w/ 1 pubKeyHash transaction input (Issue #644)",
+        "network": "testnet",
+        "txHex": "",
+        "inputs": [
+          {
+            "txHex": "0100000001f7e6430096cd2790bac115aaab22c0a50fb0a1794305302e1a399e81d8d354f4020000006a47304402205793a862d193264afc32713e2e14541e1ff9ebb647dd7e7e6a0051d0faa87de302205216653741ecbbed573ea2fc053209dd6980616701c27be5b958a159fc97f45a012103e877e7deb32d19250dcfe534ea82c99ad739800295cd5429a7f69e2896c36fcdfeffffff0340420f00000000001976a9145c7b8d623fba952d2387703d051d8e931a6aa0a188ac8bda2702000000001976a9145a0ef60784137d03e7868d063b05424f2f43799f88ac40420f00000000001976a9145c7b8d623fba952d2387703d051d8e931a6aa0a188ac2fcc0e00",
+            "vout": 0,
+            "signs": [
+              {
+                "keyPair": "cV6Fs4jspcMmUoPXyQSAVkVDfNVjL2Sn8HsBwpDvmKdhF5A9CkQW"
```
dcousens: @runn1ng this private key isn't right,  as specified in https://github.com/bitcoinjs/bitcoinjs-lib/issues/644#issuecomment-251851722

src/transaction_builder.js
```diff
@@ -210,8 +213,18 @@ function prepareInput (input, kpPubKey, redeemScript, hashType) {
...
+      // now that we have a kpPubKey?
+      expanded = expandOutput(input.prevOutScript, input.prevOutType, kpPubKey)
+      input.pubKeys = expanded.pubKeys
+      input.signatures = expanded.signatures
+    }
+
     // throw if we can't sign with it
-    if (!input.pubKeys || !input.signatures) throw new Error(input.prevOutType + ' not supported')
+    if (!input.pubKeys || !input.signatures) {
+      throw new Error(input.prevOutType + ' not supported')
```
dcousens: [If `input.pubKeys.length === 0` by this point](https://github.com/bitcoinjs/bitcoinjs-lib/pull/645/files#r82109281... maybe we should throw `not supported`... though,  I think we need more tests cases.
Either way it will fail anyway, so its as safe either way, just a matter of terminology.

test/fixtures/transaction_builder.json
runn1ng: yep, I edited the comment. I confused the private keys.

src/transaction_builder.js
runn1ng: OK, this is waaay too complex for me :))

test/fixtures/transaction_builder.json
dcousens: Thanks @runn1ng! :) Works now

src/transaction_builder.js
```diff
@@ -190,16 +193,17 @@ function prepareInput (input, kpPubKey, redeemScript, hashType) {
...
+    }
+
+    var expanded = expandOutput(redeemScript, undefined, kpPubKey)
+    if (!expanded.pubKeys) throw new Error('RedeemScript not supported "' + bscript.toASM(redeemScript) + '"')
 
-    // or, we don't have a prevOutScript, so generate a P2SH script
-    } else {
+    // if we don't have a prevOutScript, generate a P2SH script
+    if (!input.prevOutType) {
       input.prevOutScript = bscript.scriptHashOutput(redeemScriptHash)
```
dcousens: moved this to after the `throw` to avoid `input` mutation until we're on the home stretch... though, even then its not perfect,  but it is an aim

src/transaction_builder.js
```diff
@@ -210,8 +214,13 @@ function prepareInput (input, kpPubKey, redeemScript, hashType) {
     // pay-to-scriptHash is not possible without a redeemScript
     if (input.prevOutType === 'scripthash') throw new Error('PrevOutScript is P2SH, missing redeemScript')
 
-    // throw if we can't sign with it
-    if (!input.pubKeys || !input.signatures) throw new Error(input.prevOutType + ' not supported')
+    // try to derive the missing information about the script now that we
+    // have a kpPubKey
+    expanded = expandOutput(input.prevOutScript, input.prevOutType, kpPubKey)
+    if (!expanded.pubKeys) return
```
dcousens: I can not throw here because I added a second `canSign` check in `sign` now

src/transaction_builder.js
```diff
@@ -437,29 +448,36 @@ TransactionBuilder.prototype.__build = function (allowIncomplete) {
...
+  if (input.hashType !== undefined) {
     if (input.hashType !== hashType) throw new Error('Inconsistent hashType')
-  } else {
+  }
+
+  var kpPubKey = keyPair.getPublicKeyBuffer()
+  if (!canSign(input)) {
     prepareInput(input, kpPubKey, redeemScript, hashType)
+
+    if (!canSign(input)) throw Error(input.prevOutType + ' not supported')
```
dcousens: @runn1ng this should help avoid this in future :)

src/transaction_builder.js
```diff
@@ -437,29 +448,36 @@ TransactionBuilder.prototype.__build = function (allowIncomplete) {
...
-
-  if (canSign) {
-    // if redeemScript was provided, enforce consistency
-    if (redeemScript) {
-      if (!bufferEquals(input.redeemScript, redeemScript)) throw new Error('Inconsistent redeemScript')
-    }
+  // if redeemScript was provided, enforce consistency
+  if (input.redeemScript !== undefined && redeemScript) {
+    if (!bufferEquals(input.redeemScript, redeemScript)) throw new Error('Inconsistent redeemScript')
+  }
```
dcousens: I flattened this,  because why not

test/fixtures/transaction_builder.json
```diff
@@ -285,6 +285,28 @@
...
+        "network": "testnet",
+        "txHex": "010000000132595835c74fccf097db4ccae9dc2de621e58e0d3f697a27b469b61c7a223b39000000006b483045022100d771395776280955561190c09a6bca731684d09db8995c53496b816b8222019302202a21c9a90d0b5de188800673ad31861183c3f4cb15ea0988b485686aed9fce1d012103f29374a4c2c218a4077db9ba0b9d674cde3719560460af4eb3190d512dd5de92ffffffff0170170000000000001976a914ff99e06c1a4ac394b4e1cb3d3a4b2b47749e339a88ac00000000",
+        "inputs": [
+          {
+            "txHex": "0100000001f7e6430096cd2790bac115aaab22c0a50fb0a1794305302e1a399e81d8d354f4020000006a47304402205793a862d193264afc32713e2e14541e1ff9ebb647dd7e7e6a0051d0faa87de302205216653741ecbbed573ea2fc053209dd6980616701c27be5b958a159fc97f45a012103e877e7deb32d19250dcfe534ea82c99ad739800295cd5429a7f69e2896c36fcdfeffffff0340420f00000000001976a9145c7b8d623fba952d2387703d051d8e931a6aa0a188ac8bda2702000000001976a9145a0ef60784137d03e7868d063b05424f2f43799f88ac40420f00000000001976a9145c7b8d623fba952d2387703d051d8e931a6aa0a188ac2fcc0e00",
+            "vout": 0,
+            "signs": [
+              {
+                "keyPair": "cQ6483mDWwoG8o4tn6nU9Jg52RKMjPUWXSY1vycAyPRXQJ1Pn2Rq"
```
dcousens: @runn1ng I assume you're OK with this becoming a test fixture?

src/transaction_builder.js
```diff
@@ -437,29 +448,36 @@ TransactionBuilder.prototype.__build = function (allowIncomplete) {
   return tx
 }
 
+function canSign (input) {
+  return input.hashType !== undefined &&
```
dcousens: `canSign` is probably the wrong name,  its more like `isInputReady`... but another time

dcousens:
ping @runn1ng to please review :)

src/transaction_builder.js
dcousens: @runn1ng let me know if there is anything I can do to help simplify it further... I really value readability and for people to easily understand the underlying algorithms

runn1ng: I am just confused by all the possible variations of what information you can have and why. Not sure how it can be made cleaner.

runn1ng:
I will try it later today.

Also thanks for the invite! :)

src/transaction_builder.js
runn1ng: Basically I am never 100% sure what info we have at any point; I am also a little confused on differences between prepareInput/expandInput/expandOutput and at which point are they run (on signing? or at adding?)

dcousens: @runn1ng indeed!
I'm hoping to continue improving that,  and have the flow clearly outlined and probably more comments added.

*  `prepareInput` is done only if the input was not previously "prepared" when `addInput` was called.
* `expandInput` is done when a `scriptSig` is already present,  aka from `fromTransaction`.  This script also calls itself recursively if you supplied a `redeemScript`.
* `expandOutput` is done when we have a `prevOutScript` (aka, the previous transactions output) and are attempting to derive information from that script about how we might sign it.

`expandOutput` is basically useless except in the raw multisig or raw pubkey output script cases... which are very rare,  but,  if it is a pubkeyhash output... and you have your public key handy,  you can check if its pubkey hash matches which means you would be able to sign it, and hence,  it derives what is needed for that!

I'm sure it'll get cleaner,  but that is mostly the gist, with a few exceptions for `scripthash` outputs.

dcousens:
ping @fanatid @rubensayshi @afk11 for merge

src/transaction_builder.js
```diff
@@ -436,29 +446,36 @@ TransactionBuilder.prototype.__build = function (allowIncomplete) {
...
+    input.prevOutScript !== undefined &&
+    input.pubKeys !== undefined &&
+    input.signatures !== undefined &&
+    input.signatures.length === input.pubKeys.length &&
```
afk11: Is this true for a 1-of-3 multisig?

dcousens: @afk11 yes,  the number of signatures required is not `===` to the number of possible signatures,  which is what `.signatures` represents.

E.g,  for P2PKH,  before signing,  `signatures === [undefined]`

dcousens:
@runn1ng no issues I assume?

runn1ng:
I was busy. Seems working (no issue on signing), but I haven't tried to actually send the tx. Will try later if I  have time.

PR #648, opened 2016-10-06
HDNode: remove unnecessary this context

dcousens:


dcousens:
tests pass except for a integration test timing out

PR #653, opened 2016-10-07
block: use merkle-lib instead of inline calculation

dcousens:
Ref: https://github.com/bitcoinjs/merkle-lib

While I had @jprichardson's comments about NIH syndrome in the back of mind RE this.
Ultimately I wanted a library I could use for other projects that was returning the tree format in a flat array,  used `Buffer`s internally and was very easy to review.

I also wanted to continue to use the constant-space root calculation algorithm which we were using in this library.

### DIFF
--- a/src/block.js
+++ b/src/block.js
@@ -1,7 +1,7 @@
-var createHash = require('create-hash')
 var bufferutils = require('./bufferutils')
 var bcrypto = require('./crypto')
 var bufferReverse = require('buffer-reverse')
+var fastMerkleRoot = require('merkle-lib/fastRoot')
 
 var Transaction = require('./transaction')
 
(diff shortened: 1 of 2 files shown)


src/block.js
```diff
@@ -134,25 +134,13 @@ Block.calculateTarget = function (bits) {
 }
 
 Block.calculateMerkleRoot = function (transactions) {
-  var length = transactions.length
-  if (length === 0) throw TypeError('Cannot compute merkle root for zero transactions')
+  if (transactions.length === 0) throw TypeError('Cannot compute merkle root for zero transactions')
```
jprichardson: Should we have the following instead:

```js
if (!Array.isArray(transactions) || transactions.length === 0) ...
```

dcousens: @jprichardson perhaps,  `.length` would probably throw anyway.

dcousens: I suppose @jprichardson we should decide what parts of the API should have type checking,  and what shouldn't.  Its definitely not consistent right now.

dcousens: @jprichardson see https://github.com/bitcoinjs/bitcoinjs-lib/pull/653/commits/47609bc3cc6faefd5b8e77681dc4db60e3a4f99b

dcousens:
If you have any further comments,  please still write them POST merge.
This will ultimately all need to be "confirmed" before release... but just so we can move forward on the `master` branch,  I'm pushing ahead.

If anyone thinks we should move slower,  let me know.

PR #654, opened 2016-10-07
seperate coverage into a seperate travis task.

rubensayshi:
also running coverage and standards using node v6 now.


dcousens:
@rubensayshi looks like the testnet faucet's unspents have been drained entirely haha

PR #656, opened 2016-10-09
Transaction: add Witness support - WIP

dcousens:
(TODO - Untested)

@afk11 thoughts on this?

### DIFF
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -287,6 +287,8 @@ Transaction.prototype.hashForSignature = function (inIndex, prevOutScript, hashT
 }
 
 Transaction.prototype.hashForWitnessV0 = function (inIndex, prevOutScript, amount, hashType) {
+  typeforce(types.tuple(types.UInt32, types.Buffer, types.Satoshi, types.UInt32), arguments)
+
   var hashOutputs, hashPrevouts, hashSequence
 
   var tbuffer, toffset
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -309,7 +309,9 @@ Transaction.prototype.hashForWitnessV0 = function (inIndex, prevOutScript, amoun
     hashPrevouts = bcrypto.hash256(tbuffer)
   }
 
-  if (!(hashType & Transaction.SIGHASH_ANYONECANPAY) && (hashType & 0x1f) !== Transaction.SIGHASH_SINGLE && (hashType & 0x1f) !== Transaction.SIGHASH_NONE) {
+  if (!(hashType & Transaction.SIGHASH_ANYONECANPAY) &&
+       (hashType & 0x1f) !== Transaction.SIGHASH_SINGLE &&
+       (hashType & 0x1f) !== Transaction.SIGHASH_NONE) {
     tbuffer = new Buffer(4 * this.ins.length)
     toffset = 0
 
@@ -320,7 +322,8 @@ Transaction.prototype.hashForWitnessV0 = function (inIndex, prevOutScript, amoun
     hashSequence = bcrypto.hash256(tbuffer)
   }
 
-  if ((hashType & 0x1f) !== Transaction.SIGHASH_SINGLE && (hashType & 0x1f) !== Transaction.SIGHASH_NONE) {
+  if ((hashType & 0x1f) !== Transaction.SIGHASH_SINGLE &&
+      (hashType & 0x1f) !== Transaction.SIGHASH_NONE) {
     var txOutsSize = this.outs.reduce(function (sum, output) {
       return sum + 8 + scriptSize(output.script)
     }, 0)
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -6,6 +6,12 @@ var opcodes = require('./opcodes.json')
 var typeforce = require('typeforce')
 var types = require('./types')
 
+function scriptSize (someScript) {
+  var length = someScript.length
+
+  return bufferutils.varIntSize(length) + length
+}
+
 function Transaction () {
   this.version = 1
   this.locktime = 0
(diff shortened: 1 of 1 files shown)


src/transaction.js
```diff
@@ -246,6 +286,84 @@ Transaction.prototype.hashForSignature = function (inIndex, prevOutScript, hashT
...
+Transaction.prototype.hashForWitnessSignature = function (inIndex, prevOutScript, amount, hashType) {
+  var hashOutputs, hashPrevouts, hashSequence
+
+  var tbuffer, toffset
+  function writeSlice (slice) { toffset += slice.copy(tbuffer, toffset) }
+  function writeUInt32 (i) { toffset = tbuffer.writeUInt32LE(i, toffset) }
+  function writeUInt64 (i) { toffset = bufferutils.writeUInt64LE(tbuffer, i, toffset) }
+  function writeVarInt (i) { toffset += bufferutils.writeVarInt(tbuffer, i, toffset) }
+
+  if (!(hashType & Transaction.SIGHASH_ANYONECANPAY)) {
```
dcousens: this area needs to be commented as well as the other `hashForSignature`

dcousens:
Ultimately after added tests for this,  we could release `3.0.0` with preliminary segregated witness support. 

Naturally `TransactionBuilder` will need more work to make it friendly, but it should *work* and just throw `witness* is not supported`

src/transaction.js
```diff
@@ -69,6 +97,12 @@ Transaction.fromBuffer = function (buffer, __noStrict) {
     })
   }
 
+  if (hasWitnesses) {
+    for (i = 0; i < vinLen; ++i) {
+      tx.inputs[i].witness = readScript()
```
afk11: Re above & witness format:

    0x03 # number of witnesses/inputs
      00 # say the first has no data elements in the witness
      0x02 0x41 [pubkey] 0x.... [sig] # this one has two elements. each element is prefixed with it's length, but using a varint(not a script push)
      0x01 0x00 # this input has a witness. it contains one element, a byte of \x00

src/transaction.js
```diff
@@ -51,13 +68,24 @@ Transaction.fromBuffer = function (buffer, __noStrict) {
...
+
   var vinLen = readVarInt()
   for (var i = 0; i < vinLen; ++i) {
     tx.ins.push({
       hash: readSlice(32),
       index: readUInt32(),
       script: readScript(),
-      sequence: readUInt32()
+      sequence: readUInt32(),
+      witness: EMPTY_SCRIPT
```
afk11: Oh - also, that was probably intentional, because segwit is a pathway to introducing script types other than bitcoin script. So those Buffers we extract from witness data could wind up in a crazy new segwit version.

dcousens: A vector of `Buffer`s?  Maybe I'm missing how that would be more useful seeing as it needs to serialize to one `Buffer`.

src/transaction.js
dcousens: Well,  we don't decompile the script here,  so technically  `readScript` is just `readSlice` + a var int.
Shall I change that?

src/transaction.js
```diff
@@ -93,15 +127,23 @@ Transaction.prototype.isCoinbase = function () {
...
 
-var EMPTY_SCRIPT = new Buffer(0)
-
 Transaction.prototype.addInput = function (hash, index, sequence, scriptSig) {
+  return this._addInput(hash, index, sequence, scriptSig, undefined)
+}
+
+Transaction.prototype.addWitnessInput = function (hash, index, sequence, witness) {
+  return this._addInput(hash, index, sequence, undefined, witness)
+}
```
dcousens: @afk11 I haven't seen that used yet,  so would that literally be an empty P2SH script?  (aka, only the redeemScript, not scriptSig).

afk11: https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki#P2WSH_nested_in_BIP16_P2SH

src/transaction.js
afk11: Witness data is meant to be interpreted as providing the script stack for the witness program. https://github.com/bitcoin/bitcoin/blob/fc4f4547b7f6ea13d273ad926738fcdffee36295/src/script/interpreter.cpp#L1376

Storing it as an array of Buffers makes parsing signatures super easy! Your functions can now operate on an array (just like classifyOutput does), except the origin of the data can now be witness data, or a decompiled scriptSig.

A stack has multiple elements, and the serialization depends on whether it's segwit or not. The lowest common denominator between a scriptSig and witnessData is that they both decompile to an array of buffers.

src/transaction.js
rubensayshi: I think we can safely say we'll almost exclusively see P2SH wrapped segwit because 
1) there's no address for non P2SH segwit
2) it will result in addresses and scripts that can be send to by old nodes

src/transaction.js
afk11: Sorry, the number of witnesses is implicit. The rest is fine: you know there are `ninputs` witnesses. the first character you read is the number of elements for that witness. The first witness has zero elements, so it didn't use segwit. The second would be a P2PKH embedded in Segwit.

I'll whip up something to parse them to arrays

src/transaction.js
```diff
@@ -93,15 +127,23 @@ Transaction.prototype.isCoinbase = function () {
...
+
+Transaction.prototype._addInput = function (hash, index, sequence, scriptSig, witness) {
   typeforce(types.tuple(
     types.Hash256bit,
     types.UInt32,
     types.maybe(types.UInt32),
+    types.maybe(types.Buffer),
     types.maybe(types.Buffer)
   ), arguments)
+  if (scriptSig && witness) throw TypeError('scriptSig must be blank if witness is provided')
```
afk11: Same problem, scriptSig's aren't ALWAYS blank now. For P2WPKH and P2WSH, the scriptSig's are empty *unless* it's pay-to-script-hash, in which case the witnessProgram is the only element in the scriptSig.

afk11:
https://github.com/bitcoinjs/bitcoinjs-lib/compare/witnesstx...afk11:witness-tx-patched

dcousens:
Updated milestone to `3.1.0`

afk11:
Updated my branch with tests for the new hashing operation.

dcousens:
Closing in favour of https://github.com/bitcoinjs/bitcoinjs-lib/pull/684

PR #660, opened 2016-10-12
TransactionBuilder: further cleanup and refactoring

### LINKED ISSUE #642: s/hashType/hashTypes in `TransactionBuilder.inputs`
Reference https://github.com/bitcoinjs/bitcoinjs-lib/issues/417.

We currently assume that signatures are signing with the same hash type, but as described by @dthorpe,  that assumption can break in some scenarios:

(paraphrased and simplified)
```
* Alice and Bob send funds to a 2-of-2 multisig
* Bob builds a transaction spending the funds, and sends it to Alice
* Alice signs the relevant inputs with SIGHASH_ALL | SIGHASH_ANYONECANPAY and returns the transaction
* Carol adds some (signed) inputs to the transaction
* Bob signs the transaction with SIGHASH_ALL
* Transaction is broadcasted
```

dcousens:
Just part of the effort of making this more review-able. 

Cleaned up a few assumptions that weren't correct too in `fixMultisigOrder`,  it simply won't ever work unless a `redeemScript` was provided (and hence, `pubKeys` defined)

Also resolves #642

### DIFF
--- a/src/transaction_builder.js
+++ b/src/transaction_builder.js
@@ -16,7 +16,7 @@ var Transaction = require('./transaction')
 function expandInput (scriptSig, redeemScript) {
   var scriptSigChunks = bscript.decompile(scriptSig)
   var prevOutType = bscript.classifyInput(scriptSigChunks, true)
-  var hashType, pubKeys, signatures, prevOutScript
+  var pubKeys, signatures, prevOutScript
 
   switch (prevOutType) {
     case 'scripthash':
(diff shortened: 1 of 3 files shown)


dcousens:
Now includes fix for https://github.com/bitcoinjs/bitcoinjs-lib/issues/642

src/transaction_builder.js
```diff
@@ -102,8 +85,12 @@ function fixMultisigOrder (input, transaction, vin) {
       // skip if undefined || OP_0
       if (!signature) return false
 
+      // TODO: avoid O(n) hashForSignature
+      var parsed = ECSignature.parseScriptSignature(signature)
+      var hash = transaction.hashForSignature(vin, input.redeemScript, parsed.hashType)
```
dcousens: We could cache on `hashType`

PR #662, opened 2016-10-12
Transaction & block version is int32le

afk11:
https://en.bitcoin.it/wiki/Protocol_documentation#Block_Headers
https://en.bitcoin.it/wiki/Protocol_documentation#tx

The transaction and block version fields are both int32's (little endian encoded), so technically the range is -2^31 to 2^31 (not zero to 2^32)

### DIFF
--- a/src/block.js
+++ b/src/block.js
@@ -31,8 +31,14 @@ Block.fromBuffer = function (buffer) {
     return i
   }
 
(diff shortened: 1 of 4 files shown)


dcousens:
@afk11 good spot!
We'll need some more test cases for this just to make sure that doesn't change in the future :+1:

PR #665, opened 2016-10-13
HDNode: remove HDNode.toString for safety

dcousens:
Not sure why this was ever added, when we came to decision back in `1.0.0` IIRC to disable `toString` possibilities for private keys in general.
If you want the Base58... `toBase58` is the same number of letters... and is generally a more explicit opt-in.

Breaking change as someone could be relying on it for serialisation.

### DIFF
--- a/src/hdnode.js
+++ b/src/hdnode.js
@@ -317,6 +317,4 @@ HDNode.prototype.derivePath = function (path) {
   }, this)
 }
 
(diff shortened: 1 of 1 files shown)


dcousens:
Heh,  we got a txn-memory-pool conflict... I guess we'll avoid tests being run in series I suppose.

PR #666, opened 2016-10-14
travis: no need for before_install

dcousens:
I can't see why this is necessary,  so I'm removing it.  If anyone knows why,  I'll put it back.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 sudo: false
 language: node_js
-before_install:
-  - npm install npm -g
 node_js:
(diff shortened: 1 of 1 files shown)


dcousens:
Before

> $ node --version
> v6.8.0
> $ npm --version
> 3.10.8
> $ nvm --version
> 0.32.0

After

> $ node --version
> v6.8.0
> $ npm --version
> 3.10.8
> $ nvm --version
> 0.32.0

Saves 20-30 seconds of test time too.

PR #668, opened 2016-10-14
Transaction: segregated witness (alternative) (feedback requested)

afk11:
This continues @dcousens work, with some modifications to internals that should set us up well for signing, and some minor tests.

I've made some test vectors out of examples in BIP-1430. I decided to skip adding the OP_CODESEPARATOR cases, because they require special handling, and the code for that doesn't belong in this PR. 

It could certainly be squashed, but wanted to keep the commits separate for now.


afk11:
Rebased

test/transaction.js
```diff
@@ -233,4 +233,51 @@ describe('Transaction', function () {
...
+        new Buffer('00', 'hex'),
+        new Buffer('51', 'hex')
+      ]
+      tx.setWitness(0, witness)
+      assert.strictEqual(witness, tx.ins[0].witness)
+    })
+  })
+
+  describe('addWitnessInput', function () {
+    it('should add a witness bearing input', function () {
```
dcousens: what is a witness 'bearing' input?

dcousens:
I think I'll aggregate all the test fixtures first,  so we can be sure about the use cases we need to support,  then put them in a failing PR which we can then merge,  and then resolve with this PR :+1: .

src/transaction.js
```diff
@@ -261,17 +403,33 @@ Transaction.prototype.getId = function () {
   return bufferReverse(this.getHash()).toString('hex')
 }
 
-Transaction.prototype.toBuffer = function (buffer, initialOffset) {
+var SerializationFlags = {}
+SerializationFlags.NONE = 0
+SerializationFlags.ALLOW_SEGWIT = 1 << 1
```
dcousens: What is the purpose of these?

test/transaction.js
afk11: An input where the witness vector is not empty

src/transaction.js
dcousens: I'd rather no flags at this point

afk11:
Closing in favor of #684

PR #670, opened 2016-10-14
HDNode test coverage

afk11:
- First commit adds an npm script for HTML coverage reports. 
 - Second is test cases that hit simple checks in fromBuffer
 - Third commit removes the check for uncompressed keys whilst parsing from base58. We slice only 33 bytes and `ecurve.Point.decodeFrom` throws given \x04 keys which aren't 65 bytes

I guess it's making assumptions about `ecurve.Point.decodeFrom`, but it's untestable at the moment :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
   ],
   "scripts": {
     "coverage-report": "nyc report --reporter=lcov",
+    "coverage-html": "nyc report --reporter=html",
     "coverage": "nyc --check-coverage --branches 90 --functions 90 mocha",
     "integration": "mocha test/integration/",
     "standard": "standard",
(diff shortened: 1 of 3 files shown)


dcousens:
> Third commit removes the check for uncompressed keys whilst parsing from base58. We slice only 33 bytes and ecurve.Point.decodeFrom throws given \x04 keys which aren't 65 bytes

Great depiction of the issue.
I'm going to merge, but will quickly commit a comment of this assumption :+1:

dcousens:
Had to bypass travis due to issues with the fork not coming from this repository,  and therefore travis didn't carry over the environment variables...

PR #673, opened 2016-10-20
test: include derivation test when private key has leading zeros

braydonf:
- Tests derivation implementation to bip32 spec when private key has leading zeros as it's possible that the private key could be represented with less than 32 bytes.
- Cross tested against:
  - libbtc at https://github.com/braydonf/libbtc/commits/cross-ref
  - hdkey at https://github.com/cryptocoinjs/hdkey/pull/9
  - bcoin at https://github.com/bcoin-org/bcoin/pull/84
  - yours-bitcoin at https://github.com/yoursnetwork/yours-bitcoin/pull/48
  - bitcore-lib at https://github.com/bitpay/bitcore-lib/pull/97

### DIFF
--- a/test/hdnode.js
+++ b/test/hdnode.js
@@ -387,5 +387,13 @@ describe('HDNode', function () {
         }, /Expected BIP32 derivation path/)
       })
     })
+
+    it('works when private key has leading zeros', function () {
+      var key = 'xprv9s21ZrQH143K3ckY9DgU79uMTJkQRLdbCCVDh81SnxTgPzLLGax6uHeBULTtaEtcAvKjXfT7ZWtHzKjTpujMkUd9dDb8msDeAfnJxrgAYhr'
+      var hdkey = HDNode.fromBase58(key)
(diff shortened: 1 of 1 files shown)


dcousens:
Probably would add this as a test fixture instead,  but sure?  Was there any reason why this might not work?

braydonf:
All of the tests will currently pass if the 32 is removed from line
https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/src/hdnode.js#L217, and changed into:
```js
this.keyPair.d.toBuffer().copy(data, 1)
```

Instead of:
```js
this.keyPair.d.toBuffer(32).copy(data, 1)
```

So this test covers that case.

There may be one more at, that isn't covered: https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/src/hdnode.js#L193

Most of the fixtures have a seed, the seed is unknown here.

dcousens:
@braydonf just so we didn't need to change our testing fixtures dramatically with more branching,  I generated a fixture that had 5 leading leading zeros [here](https://github.com/bitcoinjs/bitcoinjs-lib/pull/674/files#diff-79325616f92d85a4322e3ebfd4f596d3R180)

dcousens:
Whatever, we'll merge both

PR #679, opened 2016-10-27
Replace script type strings with constants

afk11:
I find it's good practice to avoid repetitively using a string values throughout code. 
- It generally makes it easier for consumers of the library to check the correct value should they ever change
- IDE's will complain if you've typed the wrong string

Also added an npm script to fix the code style: `npm run standard-fixer`

### DIFF
--- a/src/script.js
+++ b/src/script.js
@@ -12,8 +12,17 @@ var REVERSE_OPS = (function () {
   }
   return result
 })()
-
 var OP_INT_BASE = OPS.OP_RESERVED // OP_1 - 1
+var scriptTypes = {
+  P2PKH: 'pubkeyhash',
+  P2PK: 'pubkey',
+  MULTISIG: 'multisig',
+  P2SH: 'scripthash',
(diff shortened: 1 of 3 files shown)


jprichardson:
ACK

src/script.js
```diff
@@ -12,8 +12,17 @@ var REVERSE_OPS = (function () {
   }
   return result
 })()
-
 var OP_INT_BASE = OPS.OP_RESERVED // OP_1 - 1
+var scriptTypes = {
```
dcousens: Can this be a constant too? (caps)

dcousens:
concept ACK

dcousens:
@afk11 could you do your changes from branches within the repository?
I could resolve some of my naming nits pretty quickly and just get your final ACK rather than put the burden on you. :+1:

dcousens:
Needs rebase

dcousens:
Merged into non-master branch so I can rebase it

PR #680, opened 2016-10-27
Script: add function for parsing multisig type scripts

afk11:
Figure this is handy since it's duplicated in script.js

### DIFF
--- a/src/script.js
+++ b/src/script.js
@@ -2,7 +2,6 @@ var bip66 = require('bip66')
 var bufferutils = require('./bufferutils')
 var typeforce = require('typeforce')
 var types = require('./types')
-
 var OPS = require('./opcodes.json')
 var REVERSE_OPS = (function () {
   var result = {}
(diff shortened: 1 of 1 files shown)


src/script.js
```diff
@@ -251,27 +250,55 @@ function isMultisigInput (script, allowIncomplete) {
...
+  var m = chunks[0] - OP_INT_BASE
+  var n = chunks[chunks.length - 2] - OP_INT_BASE
 
-  if (!types.Number(mOp)) return false
-  if (!types.Number(nOp)) return false
+  if (m <= 0) {
+    throw new Error('Number of signatures required must be greater than zero')
+  }
+  if (n > 16) {
+    throw new Error('Number of public keys must be less than 17')
```
dcousens: Isn't that policy?

src/script.js
```diff
@@ -251,27 +250,55 @@ function isMultisigInput (script, allowIncomplete) {
...
+    throw new Error('Non-canonical pubic key found')
+  }
 
-  // 0 < m <= n <= 16
-  if (m <= 0) return false
-  if (m > n) return false
-  if (n > 16) return false
-  if (n !== chunks.length - 3) return false
+  return {
+    nRequiredSigs: m,
```
dcousens: Could we just name this `m`?

src/script.js
```diff
@@ -251,27 +250,55 @@ function isMultisigInput (script, allowIncomplete) {
...
+    nPublicKeys: n,
```
dcousens: And this `n`?

src/script.js
```diff
@@ -251,27 +250,55 @@ function isMultisigInput (script, allowIncomplete) {
...
+    publicKeyBuffers: keys
```
dcousens: ACK

src/script.js
```diff
@@ -441,6 +461,7 @@ module.exports = {
   isMultisigOutput: isMultisigOutput,
   isNullDataOutput: isNullDataOutput,
 
+  parseMultisigScript: parseMultisigScript,
```
dcousens: I wonder if a `encode/decode/is(or check)` style module setup for each of these scripts types would be a great way to set it up?

dcousens:
Awesome @afk11

src/script.js
afk11: Checking intepreter.cpp, it seems 20 is the consensus enforced limit:
https://github.com/bitcoin/bitcoin/blob/fc4f4547b7f6ea13d273ad926738fcdffee36295/src/script/interpreter.cpp#L914
https://github.com/bitcoin/bitcoin/blob/57b34599b2deb179ff1bd97ffeab91ec9f904d85/src/script/script.h#L28
Limited due to other policy reasons I guess, but for now I guess 20 since it's used by OP_CHECKMULTISIG?

src/script.js
afk11: Will shorten both!

src/script.js
afk11: Yeah, it probably makes sense to do this for the scripts we support signing. While classifying we basically turn up the same information, so unifying that is good (avoids the need to call decompile a second time to get your script hash or whatever)

afk11: If we are, we probably shouldn't use `m` and `n` as names because that convention came from multisig? least it might be less obvious what they're for if used for p2pkh.

dcousens: @afk11 true.... `m-of-n` makes sense... now that I know it,  but it does require incantation to remember what it means.

dcousens:
Needs rebase

dcousens:
Closed in favour of https://github.com/bitcoinjs/bitcoinjs-lib/pull/681

PR #681, opened 2016-11-01
decode* functions for output scripts

afk11:
As brought up in #680, there is a case for including functions capable of decoding the supported output scripts. While #680 only adds this for multi-signature scripts, this PR adds methods for pubkey / pubkeyhash / multisig / nulldata / scripthash / witnesskeyhash / witnessscripthash. They were written based on the conditions used by the previous `is*` functions, and test fixtures for these conditions are included.

While starting this work I noticed something awkward - that various is* functions call `compile` on input, whereas others used `decompile`. I opted to restrict the input accepted by `is*Output` and `classifyOutput` to scripts only.

The intent for the decode* functions is to return:

  - the type of script
  - the data a wallet uses to check if it can spend the funds (pubkey for P2PK, a hash for P2PKH, a list of keys for multisig, a hash for P2SH, a hash for P2WSH, a hash for P2WPKH)
  - future supported types might return other fields: csv_locktime, cltv..

Each is* function now attempts to decode against the script type - if an exception is thrown, return false, otherwise return true. 

**NB** A nice break would be to return the output of decode after calling `is*`. Maybe this could also cascade up to classifyOutput, so once you classify you get all information about the output, cos quite often as soon as you determine it's P2SH (eg), you need to check the scriptHash against something. This takes two decodes at the moment.

### DIFF
--- a/src/script.js
+++ b/src/script.js
@@ -12,6 +12,16 @@ var REVERSE_OPS = (function () {
   return result
 })()
 
+var LIST_DECODE_NAMES = [
+  decodePubKeyOutput,
+  decodePubKeyHashOutput,
+  decodeMultisigOutput,
+  decodeNullDataOutput,
+  decodeScriptHashOutput,
+  decodeWitnessPubKeyHashOutput,
+  decodeWitnessScriptHashOutput
+]
+
 var OP_INT_BASE = OPS.OP_RESERVED // OP_1 - 1
 
 function compile (chunks) {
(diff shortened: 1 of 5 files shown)


dcousens:
> While starting this work I noticed something awkward - that various is* functions call compile on input, whereas others used decompile. I opted to restrict the input accepted by is*Output and classifyOutput to scripts only.

An optimisation actually.   They all use `script` Buffers... which are either decompiled for use or used as-is.

src/script.js
```diff
@@ -168,15 +177,34 @@ function isPubKeyHashInput (script) {
...
+    throw new Error('PubKeyHash script missing OP_EQUALVERIFY')
+  }
+  if (buffer[24] !== OPS.OP_CHECKSIG) {
+    throw new Error('PubKeyHash script missing OP_CHECKSIG')
+  }
+
+  return {
+    type: 'pubkeyhash',
+    pubKeyHash: chunks[2]
+  }
```
dcousens: Just return `chunks[2]`

src/script.js
```diff
@@ -251,54 +331,90 @@ function isMultisigInput (script, allowIncomplete) {
...
-  if (m <= 0) return false
-  if (m > n) return false
-  if (n > 16) return false
-  if (n !== chunks.length - 3) return false
+  return {
+    type: 'multisig',
+    nRequiredSigs: m,
+    nPublicKeys: n,
+    publicKeyBuffers: keys
+  }
```
dcousens: This should return the same arguments that were used to `encode` it.

That is,

`function multisigOutput (m, pubKeys) {`

So, lets just return `{ m: m,  pubKeys: pubKeys }`

dcousens:
All great. 
Epic work @afk11,  just a few nits which I'll bring up in another PR.

dcousens:
On a second note,  the performance implications of throwing exceptions repeatedly is a bit lame for the `is` checks and classification.

src/transaction_builder.js
```diff
@@ -108,7 +108,7 @@ function expandOutput (script, scriptType, ourPubKey) {
 
   var scriptChunks = bscript.decompile(script)
   if (!scriptType) {
-    scriptType = bscript.classifyOutput(scriptChunks)
+    scriptType = bscript.classifyOutput(script)
```
dcousens: Going to rever this... it was done specifically as an optimization.  Will comment that though.

dcousens: Will review

test/address.js
```diff
@@ -36,15 +36,6 @@ describe('address', function () {
...
     })
 
-    fixtures.valid.forEach(function (f) {
-      it('parses (as chunks) ' + f.script.slice(0, 30) + '... (' + f.network + ')', function () {
-        var chunks = bscript.decompile(bscript.fromASM(f.script))
-        var address = baddress.fromOutputScript(chunks, networks[f.network])
-
-        assert.strictEqual(address, f.base58check)
-      })
-    })
```
dcousens: ACK,  this was weird anyway.

But it will need to be marked as breaking change if it has truly been removed (and not just something we're trying to "discourage").

src/script.js
```diff
@@ -251,54 +331,90 @@ function isMultisigInput (script, allowIncomplete) {
...
   }
+}
 
-  return 'nonstandard'
+function determinesTypeOrNonstandard (functions, script) {
+  if (!types.Array(functions)) {
+    throw new Error('Must provide an array of functions to determinesTypeOrNonstandard')
+  }
+  if (!types.Buffer(script)) {
+    throw new Error('Must provide a script to determinesTypeOrNonstandard')
```
dcousens: Right, so this is why this is a breaking change.

Why can't we just coerce from `Buffer`?
Also,  use `typeforce`... :+1:

dcousens: Not sure if this'll make the cut

PR #682, opened 2016-11-02
Scripts: transition to encode/decode/check style

dcousens:
Other considerations:

* `bitcoin.script.*.output(...)` could be (more visually/backwards compatibly similar) short-hand for `bitcoin.script.*.output.encode(...)`

* `bitcoin.script.*.input/output` could be `bitcoin.templates.*`,  which would make the importing cleaner,  but obviously means we are separating it from `script`.
* Alternatively,  `bitcoin.script.templates.*.input/output` would be nice... but oh my god its so much text.
* Alternate alternatively,  `require('bitcoinjs-script')`,  with a `.templates` export


dcousens:
`bscript.witness*.input`'s since they are witness, not inputs.
Maybe `bscript.witness*.witness.encode` would be the best way forward there.

Also,  `classifyWitness` was introduced,  but remains untested for now.
Tests fixtures wanted! :+1:

dcousens:
ping @afk11 this is what I was thinking in terms of #681

src/address.js
```diff
@@ -38,8 +38,8 @@ function toOutputScript (address, network) {
   network = network || networks.bitcoin
 
   var decode = fromBase58Check(address)
-  if (decode.version === network.pubKeyHash) return bscript.pubKeyHashOutput(decode.hash)
-  if (decode.version === network.scriptHash) return bscript.scriptHashOutput(decode.hash)
+  if (decode.version === network.pubKeyHash) return bscript.pubKeyHash.output.encode(decode.hash)
+  if (decode.version === network.scriptHash) return bscript.scriptHash.output.encode(decode.hash)
```
dcousens: `bscript.scriptHash.output(` IMHO is cleaner, but I went with verbosity for now.
Could be added in another PR if wanted.

test/transaction_builder.js
```diff
@@ -357,7 +357,7 @@ describe('TransactionBuilder', function () {
                 var signatures = bscript.decompile(scriptSig).slice(1, -1).filter(function (x) { return x !== ops.OP_0 })
 
                 // rebuild/replace the scriptSig without them
-                var replacement = bscript.scriptHashInput(bscript.multisigInput(signatures), redeemScript)
+                var replacement = bscript.scriptHash.input.encode(bscript.multisig.input.encode(signatures), redeemScript)
```
dcousens: The verbosity is strong

dcousens:
ping @jprichardson - I think this is worth it,  but I don't like the verbosity.  Penny for your thoughts?

src/script.js
```diff
@@ -169,410 +159,19 @@ function isCanonicalSignature (buffer) {
...
-  pubKeyHashInput: pubKeyHashInput,
-  multisigInput: multisigInput,
-  scriptHashInput: scriptHashInput,
-  witnessScriptHashInput: witnessScriptHashInput
+  isDefinedHashType: isDefinedHashType
+}
+
+var templates = require('./templates')
+for (var key in templates) {
+  module.exports[key] = templates[key]
```
dcousens: As mentioned, maybe this would be better as `bitcoin.templates`... though tbh,  `bitcoin.script.templates` makes more contextual sense.

src/templates/scripthash/input.js
```diff
@@ -37,8 +37,8 @@ function decode (buffer) {
   typeforce(check, chunks)
 
   return {
-    scriptSignature: chunks.slice(0, -1),
-    scriptPubKey: chunks[chunks.length - 1]
+    redeemScriptSig: bscript.compile(chunks.slice(0, -1)),
+    redeemScript: chunks[chunks.length - 1]
```
dcousens: I renamed these to be internally consistent

jprichardson:
> It does smell strongly like a great time for breaking it out to another module.

😊

dcousens:
@jprichardson after this PR though?
Shouldn't affect the API for now IMHO.

Any other comments?

test/templates.js
```diff
@@ -227,16 +251,25 @@ describe('script-templates', function () {
...
-      it('returns ' + f.scriptPubKey, function () {
+      it('encodes ' + f.scriptPubKey, function () {
         assert.strictEqual(bscript.toASM(scriptPubKey), f.scriptPubKey)
       })
+
+      it('decodes to original arguments', function () {
+        assert.deepEqual(bscript.multisig.output.decode(scriptPubKey), {
+          m: m,
+          pubKeys: pubKeys
+        })
```
afk11: Agreed!

jprichardson:
I actually really love the concept of script templates. I hadn't thought of that before. I also really like the consistent interface provided. Not a fan of the API `bscript.script.XXXX.YYYY`... but I like it better than anything else AND it's consistent and predictable - that's the most important.

So a big LGTM.

afk11:
echo @jprichardson's thoughts - this will do nicely!

@dcousens: For witness inputs, better to follow with another PR, since they're not scripts. The are easily produced if you have a scriptSig: decompile - rejecting on any non-pushdata opcodes - and handle OP_1NEGATE, OP_1-16 using scriptNumber.

dcousens:
@afk11 agreed.

PR #684, opened 2016-11-04
Transaction: add Witness serialization and hashForWitnessV0

dcousens:
This much has basically become concrete in terms of our API.
So lets merge it and move on to the rest of the API.

@afk11 I couldn't rebase cleanly, but this is a lot of your work,  including parts from https://github.com/bitcoinjs/bitcoinjs-lib/pull/656.
Big thanks to @afk11 and @rubensayshi.

Almost there!


dcousens:
It only adds the serialization support and `hashForWitnessV0`... so,  its basically an entirely clean addition.
Which is great for progress!

Next up, `addWitnessInput` and co... but we need more test vectors.

src/transaction.js
```diff
@@ -6,6 +6,20 @@ var opcodes = require('./opcodes.json')
...
+}
+
+function vectorSize (someVector) {
+  var length = someVector.length
+
+  return bufferutils.varIntSize(length) + someVector.reduce(function (sum, witness) {
+    return sum + varSliceSize(witness)
+  }, 0)
+}
+
```
dcousens: @afk11 I renamed `varString` to `varSlice`, since that was more consistent with the rest.

src/transaction.js
```diff
@@ -252,6 +298,92 @@ Transaction.prototype.hashForSignature = function (inIndex, prevOutScript, hashT
...
+    })
+
+    hashOutputs = bcrypto.hash256(tbuffer)
+  } else if ((hashType & 0x1f) === Transaction.SIGHASH_SINGLE && inIndex < this.outs.length) {
+    var output = this.outs[inIndex]
+
+    tbuffer = new Buffer(8 + varSliceSize(output.script))
+    toffset = 0
+    writeUInt64(output.value)
+    writeVarSlice(output.script)
```
dcousens: @afk11 the PR includes your fixes/addition of this too,  nicely done :+1:

dcousens:
So,  this PR doesn't resolve the remaining API issues:

* Transaction.prototype.addInput,  but for SegWit
* TransactionBuilder SegWit support
* Transaction.toHex that allows you to choose whether you support SegWit or not.  As of this PR, it'll automatically choose the format such that `fromHex === toHex`.   Naturally, we'll need to allow users to do both.

test/fixtures/transaction.json
```diff
@@ -289,5 +289,177 @@
...
+      "id": "bd38915e2aa04b8efa307e0e29ab95582954738a7062cf88730a6276224bd917",
+      "hash": "17d94b2276620a7388cf62708a7354295895ab290e7e30fa8e4ba02a5e9138bd",
+      "hex": "0100000000010113ae35a2063ba413c3a1bb9b3820c76291e40e83bd3f23c8ff83333f0c64d623000000002322002086b2dcecbf2e0f0e4095ef11bc8834e2e148d245f844f0b8091389fef91b69ffffffffff0180969800000000001976a914b1ae3ceac136e4bdb733663e7a1e2f0961198a1788ac03004730440220121a629bb5fee3ecaf3e7a0b111101c51de816f427eaedd992b57f49b69b228e0220402ecd144a7321b4bad6ba3bfa5876b755b9c52a8c8ab17a33830d5929a76cbe0125512103b848ab6ac853cd69baaa750c70eb352ebeadb07da0ff5bbd642cb285895ee43f51ae00000000",
+      "raw": {
+        "version": "1",
+        "ins": [
+          {
+            "hash": "23d6640c3f3383ffc8233fbd830ee49162c720389bbba1c313a43b06a235ae13",
+            "index": "0",
+            "script": "002086b2dcecbf2e0f0e4095ef11bc8834e2e148d245f844f0b8091389fef91b69ff"
```
dcousens: @afk11 we basically need these fixtures, but with their witnesses.

src/transaction.js
afk11: Nice, and `vectorSize` too :)

afk11: Nice, and `vectorSize` too! :)

src/transaction.js
```diff
@@ -132,19 +181,22 @@ Transaction.prototype.addOutput = function (scriptPubKey, value) {
...
   return (
-    8 +
+    (hasWitnesses ? 10 : 8) +
     bufferutils.varIntSize(this.ins.length) +
     bufferutils.varIntSize(this.outs.length) +
-    this.ins.reduce(function (sum, input) { return sum + 40 + scriptSize(input.script) }, 0) +
-    this.outs.reduce(function (sum, output) { return sum + 8 + scriptSize(output.script) }, 0)
+    this.ins.reduce(function (sum, input) { return sum + 40 + varSliceSize(input.script) }, 0) +
+    this.outs.reduce(function (sum, output) { return sum + 8 + varSliceSize(output.script) }, 0) +
+    (hasWitnesses ? this.ins.reduce(function (sum, input) { return sum + vectorSize(input.witness) }, 0) : 0)
```
afk11: Much cleaner than mine!

src/transaction.js
afk11: Nice, those were after some of jl2012's tests. The ones we are missing are meant to test usage of FindAndDelete and OP_CODESEPARATOR, but we're not likely to need this (ever)

test/fixtures/transaction.json
afk11: I'll have a look or generate some

test/fixtures/transaction.json
```diff
@@ -289,5 +289,177 @@
...
     ]
-  }
-}
+  },
+  "hashForWitnessV0": [
+    {
+      "description": "Native P2WPKH with SIGHASH_ALL",
+      "txHex": "0100000002fff7f7881a8099afa6940d42d1e7f6362bec38171ea3edf433541db4e4ad969f0000000000eeffffffef51e1b804cc89d182d279655c3aa89e815b1b309fe287d9b2b55d57b90ec68a0100000000ffffffff02202cb206000000001976a9148280b37df378db99f66f85c95a783a76ac7a6d5988ac9093510d000000001976a9143bde42dbee7e4dbe6a21b2d50ce2f0167faa815988ac11000000",
+      "inIndex": 1,
+      "script": "OP_DUP OP_HASH160 1d0f172a0ecb48aee1be1f2687d2963ae33f71a1 OP_EQUALVERIFY OP_CHECKSIG",
```
afk11: Maybe name this hashScriptCode since we're talking about field 5: https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki#specification This scriptCode is what's left after FindAndDelete / OP_CODESEPARATOR handling. Since they have undergone this treatment, we should distance it from a raw script.

test/fixtures/transaction.json
```diff
@@ -289,5 +289,177 @@
...
+        ],
+        "locktime": "0"
+      },
+      "coinbase": false
+    },
+    {
+      "description": "Multisig",
+      "id": "3cd3dfe7312d34798b87351d9b72b3356e6ec109d841919cec510f3efa5d291f",
+      "hash": "1f295dfa3e0f51ec9c9141d809c16e6e35b3729b1d35878b79342d31e7dfd33c",
+      "hex": "0100000000010113ae35a2063ba413c3a1bb9b3820c76291e40e83bd3f23c8ff83333f0c64d623000000004a00483045022100e332e8367d5fee22c205ce1bf4e01e39f1a8decb3ba20d1336770cf38b8ee72d022076b5f83b3ee15390133b7ebf526ec189eb73cc6ee0a726f70b939bc51fa18d8001ffffffff0180969800000000001976a914b1ae3ceac136e4bdb733663e7a1e2f0961198a1788ac0000000000",
```
afk11: Doesn't decode

test/fixtures/transaction.json
```diff
@@ -289,5 +289,177 @@
...
+        ],
+        "locktime": "0"
+      },
+      "coinbase": false
+    },
+    {
+      "description": "P2PK",
+      "id": "6cf3688270c95846d3ced68b7fcf90f30daddd63e951ed04791cbf1f8c731c77",
+      "hash": "771c738c1fbf1c7904ed51e963ddad0df390cf7f8bd6ced34658c9708268f36c",
+      "hex": "0100000000010113ae35a2063ba413c3a1bb9b3820c76291e40e83bd3f23c8ff83333f0c64d623000000006a47304402203f965bdf792ea80c1f96e4292d1edb52ca62f22c7511aefca967fb9f3067063402204acbfaa4e7f1d5631227d491426d89d954d5e2abfed6f0dbc300216f01916baa012103b848ab6ac853cd69baaa750c70eb352ebeadb07da0ff5bbd642cb285895ee43fffffffff0100000000000000001976a914b1ae3ceac136e4bdb733663e7a1e2f0961198a1788ac0000000000",
```
afk11: Doesn't decode

afk11:
The above can't be decoded by Bitcoin Core's decoderawtransaction

test/fixtures/transaction.json
dcousens: Interesting @afk11,  do we need to be doing the `OP_CODESEPARATOR` then like in the traditional `hashForSignature`?

dcousens:
@afk11 these were your vectors, where did they come from?
I'm assuming from the fact they don't decode, that they were generated with our library and that somehow there is an encoding issue on our part?

afk11:
It looks like Core rejects transactions serialized as witness but without witness data, and I serialized them that way.. Oops, will revert with something better.

afk11:
I pushed new test vectors now, made sure they broadcast on testnet this time.

dcousens:
> It looks like Core rejects transactions serialized as witness but without witness data

Poses an interesting question as to whether we should do the same.
But that is separate to this PR,  because that is our behaviour currently.

afk11:
The problem with those cases was this:

RPC decoderawtransaction calls DecodeHexTx
https://github.com/bitcoin/bitcoin/blob/fc4f4547b7f6ea13d273ad926738fcdffee36295/src/rpc/rawtransaction.cpp#L511
which is here:
https://github.com/bitcoin/bitcoin/blob/57b34599b2deb179ff1bd97ffeab91ec9f904d85/src/core_read.cpp#L93
calls SerializeTransaction:
https://github.com/bitcoin/bitcoin/blob/master/src/primitives/transaction.h#L290
which decode witnesses here:
https://github.com/bitcoin/bitcoin/blob/master/src/primitives/transaction.h#L316
which fails, because it knows the entire structure is empty:
https://github.com/bitcoin/bitcoin/blob/master/src/primitives/transaction.h#L265

The same would happen when asked to serialize using the witness serialization, but all fields are empty: we should probably throw the same error.

This might imply having two separate methods for {serialize,deserialize}, because the caller should specify what form they wish to serialize. fromWitnessTx toWitnessBuffer?

afk11:
Rebased

dcousens:
> Maybe add a test case checking we reject {,de-}serializing one of those earlier ones? :)

Sounds a plan.  Are you able to add it?

dcousens:
I added https://github.com/bitcoinjs/bitcoinjs-lib/pull/684/commits/2b871fb3c3ebfb79da8feeeee201433fb3e9aa3b,  which deserializes the transactions in those vectors and matches the inputs 1:1.

It'd be great if there was just plain JSON test vectors for decoding/serialization, matching signature hashes etc without having to worry about the interpreter or signing every time.

Maybe we should make our own vectors @afk11?
In any case, I'm happy to merge once you've added the rejected test vectors

afk11:
Added the vector commit again :)

afk11:
Pushed a test case where the witness flags/structure is unnecessary, and prevents the transaction from being parsed. I can't test the case for serialization yet, because we rely on `_hasWitnesses` for witness serialization. We'll need to change this anyway, because it's not yet possible to convert a witness tx to just the base data.

src/transaction.js
```diff
@@ -435,10 +441,15 @@ Transaction.prototype.toBuffer = function (buffer, initialOffset) {
...
-  if (hasWitnesses) {
+  if (serializeWitnesses) {
+    var isNull = true
     this.ins.forEach(function (input) {
       writeVector(input.witness)
+      isNull = isNull && input.witness.length === 0
     })
+    if (isNull) {
+      throw new Error('Transaction has superfluous witness data')
+    }
```
dcousens: `this.ins.every(function (input) { return input.witness.length === 0 })` is clearer IMHO

src/transaction.js
```diff
@@ -120,8 +120,13 @@ Transaction.fromBuffer = function (buffer, __noStrict) {
   }
 
   if (hasWitnesses) {
+    var isNull = true
     for (i = 0; i < vinLen; ++i) {
       tx.ins[i].witness = readVector()
+      isNull = isNull && tx.ins[i].witness.length === 0
+    }
+    if (isNull) {
+      throw new Error('Transaction has superfluous witness data')
```
dcousens: Do we need this?
I understand its not ideal,  but,  it doesn't really hurt any one.

dcousens: > It looks like Core rejects transactions serialized as witness but without witness data, and I serialized them that way.. Oops, will revert with something better.
 :+1: 1  

Right,  but is that Core or just the RPC?

@btcdrak ?

dcousens: Nevermind... from https://bitcoincore.org/en/segwit_wallet_dev/

> 
If all txins in a transaction are not associated with any witness data, the transaction MUST be serialized in the original transaction format, without marker, flag, and witness. For example, if none of the txins are coming from segwit UTXO, it MUST be serialized in the original transaction format. (exception: coinbase transaction)

test/transaction.js
```diff
@@ -118,30 +118,19 @@ describe('Transaction', function () {
...
-    })
-
-    it('defaults to empty script', function () {
+    it('defaults to empty script, witness and 0xffffffff SEQUENCE number', function () {
       var tx = new Transaction()
       tx.addInput(prevTxHash, 0)
 
       assert.strictEqual(tx.ins[0].script.length, 0)
+      assert.strictEqual(tx.ins[0].witness.length, 0)
+      assert.strictEqual(tx.ins[0].sequence, 0xffffffff)
```
dcousens: Stopped referring to `Transaction.DEFAULT_SEQUENCE` here, as its a bit circular otherwise

dcousens:
> The txid remains the primary identifier of a transaction:
> It MUST be used in the txin when referring to a previous output.
> If a wallet or service is currently using txid to identify transactions, it is expected to use the same after upgrade.

Can't be merged as is since `getId` output would be ambiguous.
Bleh, @afk11 let's organise a catchup on slack

afk11:
Lets get this merged! :dancing_men: :dancing_women:

btcdrak:
Remember to update https://github.com/bitcoin-core/bitcoincore.org/blob/gh-pages/_data/segwitsupport.csv when merged.

PR #689, opened 2016-11-05
Add bscript.toWitness and tests

afk11:
May come in useful for some of the witness templates, and when parsing signatures from normal or witness transactions (trick: convert normal sigs to witnesses, and extract signatures that way)


src/script.js
```diff
@@ -161,6 +176,7 @@ function isCanonicalSignature (buffer) {
 module.exports = {
   compile: compile,
   decompile: decompile,
+  toWitness: toWitness,
```
dcousens: maybe we should just have `bitcoin.witness` and have `compile/decompile` there instead?

afk11: It could be called evalPushOnly (there's a complementary function PushAll from Core which does the reverse: 
https://github.com/bitcoin/bitcoin/blob/2f71490d21796594ca6f55e375558944de9db5a0/src/script/sign.cpp#L124)

It feels like it belongs in script. I plan on passing scriptSig's through this to prepare for signature extraction, because then extraction can work against an eval'd scriptSig, or a witness as is, making the function mostly agnostic to segwit.

afk11:
How about an interpreter module for stuff like this? Supporting the standard script types would be cool, since you can verify signatures that way.

`DUP`, `HASH160`, `EQUALVERIFY`, `CHECKSIG`, `EQUAL`, `CHECKSIG`, `CHECKMULTISIG` being handled, plus integers as handled in this PR would cover the standard types for an `EvalScript` like function, then support for P2SH and witness like in `VerifyScript`.

src/script.js
dcousens: >  I would name it decompilePushOnly, but the function changes some opcodes to Buffers.

I haven't got time at the moment, but could you reference to this behaviour in the BIPs?

dcousens:
@afk11 I've avoided an interpreter thus far in this library for the sole reason,  we can't easily test every case.
And I'd rather not try and become completely consensus compatible... as that is a monumental task to do right.

src/script.js
afk11: > https://github.com/bitcoin/bitcoin/blob/fc4f4547b7f6ea13d273ad926738fcdffee36295/src/script/interpreter.cpp#L1352
> Witness data is NOT script. https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki#specification

In normal scripts, you arrive at `stack` by interpreting a scriptSig. Witness data is basically the stack, ready to be operated on by the witness script. So, if you're parsing signatures, if it's P2SH, a regular p2pk/p2pkh/multisig, you have to decompile scriptSig and do some work. 

But if it's witness, you've already got the Buffers ready to parse, and the layout of the element (`sig pubkey` for p2pkh, `sig` for P2PK, `OP_0 sig sig` for multisig) is exactly the same. 

It's because witnesses don't have opcodes that I'd like a function that has the `script` -> `push only elements` semantics. It has useful side effects when trying to extract sigs/sign/serialize signatures.

afk11:
That makes sense. I don't mean all out, and the templated approach works just as well. I might find some spare time, I'd be nice to check signatures against the data used to produce them.

dcousens:
@afk11 do you still feel this is the best API for this OP?
You mentioned on slack that push only might be useful somewhere else?

afk11:
Yep, closing in favor of #700

PR #690, opened 2016-11-08
types: add Script ASM type check [WIP]

dcousens:
This does result in a double parse, but it is useful for avoiding weird cases where `Buffer` will accept `0xff` but returns `Buffer< >`...


dcousens:
Somehow that regex is causing an infinite loop.  None of the tests pass.

Awesome.

dcousens:
... or not?
I have no idea what is going wrong with this PR.

PR #691, opened 2016-11-08
Alternate to 99cc40f

dcousens:


src/transaction.js
```diff
@@ -188,7 +186,7 @@ Transaction.prototype.addOutput = function (scriptPubKey, value) {
 
 Transaction.prototype._hasWitnesses = function () {
   return this.ins.some(function (x) {
-    return x.witness !== EMPTY_WITNESS
+    return x.witness.length !== 0
```
dcousens: This was probably a bug anyway :+1:

src/transaction.js
```diff
@@ -441,15 +438,10 @@ Transaction.prototype.toBuffer = function (buffer, initialOffset) {
...
-  if (serializeWitnesses) {
-    var isNull = true
+  if (hasWitnesses) {
     this.ins.forEach(function (input) {
       writeVector(input.witness)
-      isNull = isNull && input.witness.length === 0
     })
-    if (isNull) {
-      throw new Error('Transaction has superfluous witness data')
-    }
```
dcousens: I think a better way to do this would be with something similar to `clearWitnesses()` or `.setWitness(i, EMPTY_WITNESS)` prior to encoding.

dcousens:
For `Transaction.prototype.getId`,  @afk11 how about something like:

``` javascript
  Transaction.prototype.getId = function () {
    // transaction hash's are displayed in reverse order
   return bufferReverse(this.withoutWitnesses().getHash()).toString('hex')
  }
```

src/transaction.js
afk11: How does it know to call `.setWitness(i, EMPTY_WITNESS)` ? 

It's up to the caller to decide what way to serialize for witness. Separate issue, the `isNull` check only prevents a non-segwit transaction being serialized as a segwit one. The checks prevent an invalid case from being serialized/deserialized.

src/transaction.js
```diff
@@ -120,14 +120,12 @@ Transaction.fromBuffer = function (buffer, __noStrict) {
...
     for (i = 0; i < vinLen; ++i) {
       tx.ins[i].witness = readVector()
-      isNull = isNull && tx.ins[i].witness.length === 0
-    }
-    if (isNull) {
-      throw new Error('Transaction has superfluous witness data')
     }
+
+    // was this pointless?
+    if (!tx._hasWitnesses()) throw new Error('Transaction has superfluous witness data')
```
afk11: Since it addresses the only erroneous conversion between base & witness formats I'm not sure it is. Our test cases were happily accepting transactions that Bitcoin Core wouldn't, this check would have prevented that.

src/transaction.js
afk11: We can't even test this check yet because there's no way to ask for only the base serialization.

afk11:
ACK - no way to add a review post-merge

PR #693, opened 2016-11-09
Remove .getHash

dcousens:
Thoughts?

Does anyone use these methods in a situation where it is actually relevant?
The cases we currently use internally are:

* `Block.calculateMerkleRoot`
* `TransactionBuilder.addInput`

Both of which are places which use `hash256` in near proximity anyway, so it wouldn't even be weird to remove it.

If you are using this in production somewhere, let us know, it can be re-added if needed.

--- Just a random thought when trying to figure out an ideal solution for the SegWit changes.

Don't merge without discussion... not even that keen on it tbh.


dcousens:
Actually, the subtleties that may be required here in terms of SegWit make this worthwhile keeping around...

PR #696, opened 2016-11-09
TransactionBuilder: add fee safety

dcousens:
**edit:** For future reference,  the merged constant is a maximum, minimum fee-rate, of 1000 satoshis per byte.


src/transaction_builder.js
```diff
@@ -288,21 +291,30 @@ TransactionBuilder.prototype.addInput = function (txHash, vout, sequence, prevOu
...
-  return this.__addInputUnsafe(txHash, vout, sequence, null, prevOutScript)
+  return this.__addInputUnsafe(txHash, vout, {
+    sequence: sequence,
+    prevOutScript: prevOutScript,
+    value: value
+  })
 }
 
-TransactionBuilder.prototype.__addInputUnsafe = function (txHash, vout, sequence, scriptSig, prevOutScript) {
+TransactionBuilder.prototype.__addInputUnsafe = function (txHash, vout, options) {
```
dcousens: Maybe this is how we should handle `addInput` universally for SegWit and the upcoming changes?

src/transaction_builder.js
```diff
@@ -272,7 +272,10 @@ TransactionBuilder.fromTransaction = function (transaction, network) {
 
   // Copy inputs
   transaction.ins.forEach(function (txIn) {
-    txb.__addInputUnsafe(txIn.hash, txIn.index, txIn.sequence, txIn.script)
+    txb.__addInputUnsafe(txIn.hash, txIn.index, {
+      sequence: txIn.sequence,
+      script: txIn.script
+    })
```
dcousens: I could probably just pass `txIn` here as options, haha

src/types.js
```diff
@@ -10,7 +10,7 @@ function BIP32Path (value) {
 }
 BIP32Path.toJSON = function () { return 'BIP32 derivation path' }
 
-var SATOSHI_MAX = 2.1 * 1e15
+var SATOSHI_MAX = 21 * 1e14
```
dcousens: This was annoying me

src/transaction_builder.js
```diff
@@ -479,4 +499,18 @@ TransactionBuilder.prototype.__canModifyOutputs = function () {
...
+  var incoming = this.inputs.reduce(function (a, x) { return a + (x.value >>> 0) }, 0)
+
+  // but all outputs do, and if we have any input value
+  // we can immediately determine if the outputs are too small
+  var outgoing = this.tx.outs.reduce(function (a, x) { return a + x.value }, 0)
+  var fee = incoming - outgoing
+
+  // its not fool-proof, but, it might help somebody
+  // fee > 0.2BTC
+  return fee > (0.2 * 1e8)
```
dcousens: I also don't think anything beyond a hard-coded constant is reasonable without annoying real users

dcousens: Also, is this value suitable?

dcousens:
ping @afk11 for ACK/NACK

src/transaction_builder.js
```diff
@@ -367,6 +384,9 @@ TransactionBuilder.prototype.__build = function (allowIncomplete) {
   if (!allowIncomplete) {
     if (!this.tx.ins.length) throw new Error('Transaction has no inputs')
     if (!this.tx.outs.length) throw new Error('Transaction has no outputs')
+
+    // do not rely on this, its merely a last resort
+    if (this.__hasAbsurdFee()) throw new Error('Transaction has absurd fees')
```
afk11: Is there any way to disable this?

src/transaction_builder.js
afk11: - :+1: for exporting the constant. 
 - Not sure if it's suitable at all :D

If it's not a requirement to provide a value, it seems fee can be negative which would cause the function to declare an absurd fee. Is it intended to introduce this behavior in here? https://github.com/bitcoinjs/bitcoinjs-lib/pull/696/files#diff-52af0f8b5da51627aa8d01193636ba73R384

afk11:
I guess my questions are: 
 - do we always want to check the fee for transactions? Can the check be disabled?
The output value is optional even with segwit (it's not always required, so can safely allow the user not to provide it in those cases), but we can make it required going forward since we're introducing a break?

 - It doesn't seem to check the transaction size, which could more than warrant such a large fee. Could maybe use ABSURD_FEERATE instead?

src/transaction_builder.js
dcousens: > it seems fee can be negative which would cause the function to declare an absurd fee

What? How?
The fee can be negative,  but that is fine.  It will still return false as it it less than absurd.

dcousens:
> Can the check be disabled?

In the cases where you wouldn't want it (incomplete transactions),  it is avoided by using `buildIncomplete`.

> Should we require the value going forward?

No? Its optional obviously.

> It doesn't seem to check the transaction size, which could explain such a large fee. Could maybe use ABSURD_FEERATE instead? 

Sure,  we have the signatures at this point too so should be straight forward.

dcousens:
- [ ] Change to ABSURD_FEERATE instead of ABSURD_FEE
- [ ] Expose as constant on `TransactionBuilder`

dcousens:
The assumptions this makes is:

* Inputs values are `undefined` or `> 0`
* Output values are all `> 0`
* `fee = incoming - outgoing`
* If we have **any** input values,  and their sum is larger than the outgoing,  we can determine the **minimum** fee for the transaction (not the actual fee rate, the *minimum*).

At that point,  we can say,  if the minimum is absurd,  it doesn't matter what the rest of the inputs are, because they only make it worse.

Naturally,  in most cases,  the [detected] minimum fee for the transaction is negative,  because `.value` is `undefined` (and treated as `0`).
But, that is fine,  because this isn't to be relied on,  its simply an aid to prevent [potential] algorithmic mistakes.

src/transaction_builder.js
afk11: Oops, my mistake

src/transaction_builder.js
afk11: The fee wouldn't be absurd if the transaction is very large, or you really want it to confirm. It feels it should be possible without setting `allowIncomplete = true`

dcousens: @afk11 if you're doing 150kb transactions 120 satoshis/byte,  only then,  will you hit this limit.
With a constant `TransactionBuilder.ABSURD_FEERATE`,  a user could change that to by-pass these limits themselves,  as they will be a *very* select few people.

dcousens:
I think an absurd fee rate of >300 satoshis per byte is a good limit for now.

dcousens:
The only remaining part of the TODO for this PR is:

> Add canModifyOutputs check which would stop this from being correct

Personally,  I can't decide if this is a good or a bad thing... so I'm going to leave it to a patch PR later rather than do it.

dcousens:
Merging once Travis passes.

dcousens:
@afk11 this works a treat :+1:

dcousens:
ping @afk11 on 6a019ee

afk11:
ACK 6a019ee

Issue #695, opened 2016-11-09
SegWit: the TransactionBuilder API


dcousens:
Just a dumping ground for thoughts in regard to the TransactionBuilder and segregated witness support

afk11:
There seems to be this kind of motif throughout working a segwit transaction, whether it's extracting signatures, signing, or serializing them. 

I try to keep 'stuff' general, so `extractSigsFromChunks(type, chunks)` is there to handle regular scriptSigs [decompiled], or P2SH script sigs [decompiled], witnesses [they already are an array!]

```
var SIGNABLE_SCRIPTS = [bscript.types.P2PK, bscript.types.P2PKH, bscript.types.MULTISIG]
var type = bscript.classifyOutput(prevOutScript)
if (SIGNABLE_SCRIPTS.indexOf(type)) {
  // do stuff
}

if (type === bscript.types.P2SH) {
  // take redeemScript, and doStuff with it instead
  // set type to redeemScriptType
}

if (type === bscript.types.P2WPKH) {
  // do stuff, by generating the p2pkh scriptCode in situ
} else if (type === bscript.types.P2WSH) {
  // take witnessScript, and doStuff with it
  // set type to witnessScriptType
}

if (SIGNABLE_SCRIPTS.indexOf(type)) {
  // it worked
}
```

afk11:
As you can see, this relies on knowing what prevOutScript is. I dislike coming up with this by inspecting the scriptSig. With the right data it should be able to keep you safe, not help you do silly things. 

That opinion aside, the library needs the scriptPubKey, and the value if the scriptPubKey (directly or by P2SH) uses segwit. With that, I suggest we pass the full txOut field in going forward. It's a break, lets make it worthwhile. 

Next part: redeem & witness scripts. With the full txOut field, we can now decide whether to accept or reject any attempted P2SH or P2WSH scripts. This is probably saner than deciding on the presence of a redeemScript, or all else, assume P2PKH. 

With the txOut, redeemScript, and witnessScript all referencing eachother, you can now encode/decode any scriptSig / witness, and verify signatures. We want to reach this state before we attempt to sign.  We also need the txOut & redeemScript/witnessScript/sigVersion around just to parse partially-signed multisigs.

So, lets fail hard if the correct conditions aren't met to sign an input. 
 - (tx, nIn, txOut) are now required, optionally including options with redeemScript / witnessScript
 - txOut is checked against redeemScript and witnessScript. save them all for later.
[...]

PR #697, opened 2016-11-09
New function: script.solveOutput

afk11:
This function builds on `classifyOutput` and the various `decode` functions. It returns an object with the following properties based on the script. 
  - `type` the type of script
  - `script` we include the script itself. 
  - `solvedBy` - the result of calling `decode` for that type of output script
  - `requiredSigs` is null except for P2PKH/P2PK/multisig. 
  - <strike>`canSign` indicates whether the script is signable (defined later). </strike>
 
<strike>It also adds two constants: `SIGNABLE_SCRIPTS`, ie, the ones we support signing (p2pk/p2pkh/multisig). `P2SH_SCRIPTS` is SIGNABLE_SCRIPTS and also P2PKH & P2WSH.</strike>

```js
{
  type: "pubkey",                  // any of the bscript.types
  script: "4104ababababab......ac" // as is
  solvedBy: "04ababababab..",    // ** this is the output of the respective output.decode() function
  requiredSigs: 1                  // set for P2PK, P2PKH, Multisig, P2WPKH
}
```
This is useful when determining if there is enough information to sign, and guides the approach outlined in https://github.com/bitcoinjs/bitcoinjs-lib/issues/695


src/templates/index.js
```diff
@@ -54,16 +57,69 @@ function classifyWitness (script, allowIncomplete) {
...
+  }
+
+  return {
+    type: outputType,
+    script: scriptCode,
+    solvedBy: solvedBy,
+    canSign: SIGNABLE_SCRIPTS.indexOf(outputType) !== -1,
+    requiredSigs: requiredSigs
+  }
+}
```
dcousens: `scripts` has been so far purely about templates... this feels like the wrong place

afk11:
Anything else required for this?

dcousens:
Could you add an integration test to show-case how it might be used?
Or show how it might be more idiomatic then doing it yourself?

afk11:
Yeah, but it captures basically all the state you are deducing with, and gives you something more useful to save as state than `prevOutScript` `redeemScript` `witnessScript` since you probably have to work with the same material again. (signing, then later you need to classify to determine where things go)

dcousens:
@afk11 did you see my comments on slack?

PR #698, opened 2016-11-09
[for witnesshash] More info for witness fixtures

afk11:
Adds the vectors provided in #684 by including the scriptPubKey/redeemScript/witnessScript/utxo amount, should come in useful when it comes to signing.


dcousens:
My only concern is this information, as you mentioned, is only useful for signing.
Completed unrelated to the `transaction` fixtures... but for now we'll keep it.

Should be in `transaction_builder` fixtures though.

PR #699, opened 2016-11-12
tests: add minimally encoded 0-length buffer script

dcousens:
Just a curiosity worth noting/testing


afk11:
If you add `"scriptPubKey": "OP_0",` the rest of the diff can be dropped?

dcousens:
@afk11 in my mind that read a bit confusingly... but yeah you're probably right.

PR #701, opened 2016-11-14
Transaction: Witness serialization support default only for some func…

dcousens:
…tions

- [x] Add [originally failing] tests to ensure TXID is encoded without SegWit


src/transaction.js
```diff
@@ -294,9 +298,9 @@ Transaction.prototype.hashForSignature = function (inIndex, prevOutScript, hashT
   }
 
   // serialize and hash
-  var buffer = new Buffer(txTmp.byteLength() + 4)
+  var buffer = new Buffer(txTmp.__byteLength(false) + 4)
   buffer.writeInt32LE(hashType, buffer.length - 4)
-  txTmp.toBuffer(buffer, 0)
+  txTmp.__toBuffer(buffer, 0, false)
```
dcousens: @afk11 this is my understanding,  `hashForSignature` (the original) should sign the non-segwit serialized transaction.

dcousens:
- [ ] Add tests for `hasWitnesses`

src/transaction.js
```diff
@@ -395,7 +399,11 @@ Transaction.prototype.getId = function () {
 }
 
 Transaction.prototype.toBuffer = function (buffer, initialOffset) {
-  if (!buffer) buffer = new Buffer(this.byteLength())
+  return this.__toBuffer(buffer, initialOffset, true)
+}
```
dcousens: Note the public API does not change

src/transaction.js
```diff
@@ -386,7 +390,7 @@ Transaction.prototype.hashForWitnessV0 = function (inIndex, prevOutScript, value
 }
 
 Transaction.prototype.getHash = function () {
-  return bcrypto.hash256(this.toBuffer())
+  return bcrypto.hash256(this.__toBuffer(undefined, undefined, false))
 }
```
dcousens: This has to use an internal function such that getHash is always returning the legacy TXID of the transaction,  not WTXID

dcousens:
@afk11 the only part of any of this: naming things,  and finding test fixtures

afk11:
I'll dig up the wtxid tomorrow - the witness fixtures are on testnet!

dcousens:
Rebased and now adds failing fixtures before adding a fix commit.

For example, https://github.com/bitcoinjs/bitcoinjs-lib/pull/701/commits/16abf3ed2b6d10ed8f70fd991a6ca9f39ac565b0 was failing , but fixed with https://github.com/bitcoinjs/bitcoinjs-lib/pull/701/commits/2683391df2cc64edbe752d934628b68edae11366, as `setWitness` was not accepting valid arguments.

https://github.com/bitcoinjs/bitcoinjs-lib/pull/701/commits/7a54826c3e27245d1bbe592296bd562e2c664055 was failing due to the default serialization being to always serialize witnesses if available.
Fixed by https://github.com/bitcoinjs/bitcoinjs-lib/pull/701/commits/44e386ec6b2f726f9659cadaba700ea905c23a23.

test/fixtures/transaction.json
```diff
@@ -229,60 +229,18 @@
...
       "hash": "6a3ec142a4fea834667fdfbabcfad413b224ce1bb3ee651a33b3e95b9933478d",
       "hex": "010000000193aef40ae141694895e99e18e49d0181b086dd7c011c0241175c6eaf320099970000000049483045022100e57eba5380dcc8a7bdb5370b423dadd43070e1ca268f94bc97b2ded55ca45e9502206a43151c8af03a00f0ac86526d07981e303fc0daea8c6ed435abe8961533046d01ffffffff0160ea0000000000001976a914851a33a5ef0d4279bd5854949174e2c65b1d450088ac00000000",
-      "hasWitness": false,
-      "witnessHex": "",
+      "whex": "",
       "raw": {
         "version": 1,
         "ins": [
           {
-            "hash": "97990032af6e5c1741021c017cdd86b081019de4189ee995486941e10af4ae93",
```
dcousens: Bitcoin core has ambiguity between `hash` and an `id`.
For testing purposes,  we always call the `hash` the raw hex string which would have been returned by `sha256`, and the `id` is reversed.

This amends the test fixtures to reflect that.

test/fixtures/transaction.json
```diff
@@ -311,13 +269,12 @@
       "id": "d4ebb5e09c06dd375f265b51eb7f1a14d12deba71329500e2a00a905fdfdc27d",
       "hash": "7dc2fdfd05a9002a0e502913a7eb2dd1141a7feb515b265f37dd069ce0b5ebd4",
       "hex": "0100000001a30e865fa60f6c25a8b218bb5a6b9acc7cf3f1db2f2e3a7114b51af5d6ae811f000000006c473044022026d2b56b6cb0269bf4e80dd655b9e917019e2ccef57f4b858d03bb45a2da59d9022010519a7f327f03e7c9613e0694f929544af29d3682e7ec8f19147e7a86651ecd012321038de63cf582d058a399a176825c045672d5ff8ea25b64d28d4375dcdb14c02b2bacffffffff0160ea0000000000001976a914851a33a5ef0d4279bd5854949174e2c65b1d450088ac00000000",
-      "hasWitness": false,
-      "witnessHex": "",
```
dcousens: no need having both

test/transaction.js
```diff
@@ -73,14 +78,25 @@ describe('Transaction', function () {
...
-        var actual = fromRaw(f.raw)
+        if (f.whex) {
+          var wactual = fromRaw(f.raw)
+          assert.strictEqual(wactual.toHex(), f.whex)
+        }
 
-        assert.strictEqual(actual.toHex(), f.hex, actual.toHex())
+        var actual = fromRaw(f.raw, true)
+        assert.strictEqual(actual.toHex(), f.hex)
       })
```
dcousens: might as well test both, because we can

dcousens:
@afk11 we can put `getWitnessHash` and `getWitnessId` in another PR.
They are not limiting this one in any way, and not really related either.

dcousens:
`hasWitnesses` being added to the public API should be discussed in https://github.com/bitcoinjs/bitcoinjs-lib/issues/703 first.

This PR purely focuses on fixing the `getId` and import/export ambiguity issues.

dcousens:
Good to merge!

dcousens:
@afk11 nothing controversial here AFAIK,  so I've merged to ease the rebase hell.  But please still review.

test/fixtures/transaction.json
afk11: Doh, I should have double checked that.

src/transaction.js
afk11: For now that makes sense! There might be room for a convenience function that takes (scriptCode, sigVersion, sigHashType, [amount]) and returns the appropriate hash? It's not very long!

PR #708, opened 2016-11-14
New function: script.solveOutput

afk11:
Replaces https://github.com/bitcoinjs/bitcoinjs-lib/pull/697 because that targeted another base.

### DIFF
--- a/src/templates/index.js
+++ b/src/templates/index.js
@@ -17,9 +17,6 @@ var types = {
   P2WSH: 'witnessscripthash'
 }
 
-var SIGNABLE_SCRIPTS = [types.MULTISIG, types.P2PKH, types.P2PK]
-var P2SH_SCRIPTS = SIGNABLE_SCRIPTS.concat([types.P2WPKH, types.P2WSH])
-
 function classifyOutput (script) {
   if (witnessPubKeyHash.output.check(script)) return types.P2WPKH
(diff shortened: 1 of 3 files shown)


afk11:
This tests both sighash implementations against live vectors, plus the test logic may well end up incorporated into signing somewhere

afk11:
Rebased

afk11:
Rebased, time for merge?

afk11:
One of the motivations of this PR is to make stuff like this easier: 
https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/src/transaction_builder.js#L150

We could run solveOutput, save that as prevOutput, then we get the following without having to duplicate work:
```
 {
    type: outputType,
    script: scriptCode,
    solvedBy: solvedBy,
    requiredSigs: requiredSigs
}
```

dcousens:
@afk11 we could maybe make it private for now,  try out the use cases then decide?

afk11:
I still think we need this, mind merging it?

dcousens:
ACK,  but I don't think the name is right as it implies `solveOutput(ouput) -> solvedOutput` ... aka, it is a verb.
It is actually akin `whatSolvesThisOutput(output)`

dcousens:
@afk11 thoughts on this?
It is useful,  but,  I struggled with the exact use-case... maybe an integration test?

Issue #688, opened 2016-11-04
SegWit: support legacy Transaction export


dcousens:

afk11:
https://github.com/bitcoin/bitcoin/pull/9194

dcousens:
I love how they had all the same questions we do :+1: .

`toLegacyBuffer`?

dcousens:
Closing in favour of #715

PR #715, opened 2016-12-06
Transaction: add toLegacyBuffer

afk11:
Adds `Transaction.getLegacyBuffer`

Related issue: https://github.com/bitcoinjs/bitcoinjs-lib/issues/688

### DIFF
--- a/test/transaction.js
+++ b/test/transaction.js
@@ -56,6 +56,10 @@ describe('Transaction', function () {
         var actual = Transaction.fromHex(txHex)
 
         assert.strictEqual(actual.toHex(), txHex)
+        if (!f.whex) {
+          // fromHex tolerates either format, so only test this if witness is not set
+          assert.strictEqual(actual.toLegacyBuffer().toString('hex'), txHex)
+        }
       })
 
       if (f.whex) {
@@ -63,6 +67,10 @@ describe('Transaction', function () {
           var actual = Transaction.fromHex(f.whex)
 
           assert.strictEqual(actual.toHex(), f.whex)
+          // if we have the non-witness format, check legacy export matches
+          if (f.hex) {
+            assert.strictEqual(actual.toLegacyBuffer().toString('hex'), f.hex)
+          }
         })
       }
     }
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -398,6 +398,10 @@ Transaction.prototype.getId = function () {
   return bufferReverse(this.getHash()).toString('hex')
 }
 
+Transaction.prototype.toLegacyBuffer = function (buffer, initialOffset) {
+  return this.__toBuffer(buffer, initialOffset, false)
+}
+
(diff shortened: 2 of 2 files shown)


dcousens:
> getWitnessId returns a Buffer - it could have been hex, but since a witness-hash isn't a thing..

That is too confusing IMHO,  its almost not worth having it if it is that confusing.
Separate PR?

afk11:
Yeah getId kind of sets the convention for returning hex. I'll axe getWitnessId until it's discussed further, though the wtxid is used in the witness coinbase commitment.

test/transaction.js
```diff
@@ -56,13 +56,29 @@ describe('Transaction', function () {
         var actual = Transaction.fromHex(txHex)
 
         assert.strictEqual(actual.toHex(), txHex)
+        if (!f.whex) {
+          // fromHex tolerates either format, so only test this if witness is not set
+          assert.strictEqual(actual.toLegacyBuffer().toString('hex'), txHex)
+        }
+        if (f.id) {
+          // id is always whatever's in the file
```
dcousens: what do you mean in the file?

dcousens:
LGTM

test/transaction.js
dcousens: so why do `if (f.id)` at all?

afk11: Actually, removing that one since the getId test will test the witness if available.

dcousens:
ACK, will merge on travis pass

dcousens:
rebased, will merge

dcousens:
@afk11 for documentation,  could you list the reasons why we need this?

dcousens:
So,  to list the reason from: https://github.com/bitcoin/bitcoin/pull/9194

* Downstream users (not users of this library) may not support SegWit yet,  so this is to allow clients to pass them `txHex` that may be defaulting to the newer format.

I'm inclined to think this change is nice to have, but really should be discouraged by putting the burden on the downstream users to upgrade instead.

Perhaps we'll keep the PR open for as long as until some complains/asks about it?

dcousens:
> Maybe we should work out better names for toBuffer / toLegacyBuffer, since it's only possible to have a witness serialization if one of the outputs called for it?

You mean,  if one of the inputs (previous outputs) required it?
`toBuffer` is meant to represent the default serialization,  on the blockchain.

If this libraries assumes segregated witness is in use (it will in `3.0.0`),  then,  that is representative.

The only reason to use `toLegacyBuffer`,  is supporting non-upgraded clients... AFAIK.
I think my point in https://github.com/bitcoinjs/bitcoinjs-lib/pull/715#issuecomment-267202804 holds best,  we don't merge this unless someone asks for it?

afk11:
Agree, we can hold this one off until someone calls for it

dcousens:
Closing (but keeping (not deleting) the branch)

Issue #702, opened 2016-11-14
SegWit: add tests to verify hashForSignature only uses the legacy transaction serialization


dcousens:

dcousens:
Closing in favour of #716

PR #716, opened 2016-12-07
Test: verify Transaction.hashForSignature uses non-witness format.

afk11:
This PR adds a small test that `tx.__toBuffer` is "never" called with `__allowWitnesses=true` during hashForSignature. This would only arise due to error since __toBuffer gets directly called in this function, and passes false always. 

Related issue: https://github.com/bitcoinjs/bitcoinjs-lib/issues/702

### DIFF
--- a/test/transaction.js
+++ b/test/transaction.js
@@ -205,6 +205,29 @@ describe('Transaction', function () {
   })
 
   describe('hashForSignature', function () {
+    it('only uses V0 serialization', function () {
+      var randScript = new Buffer('6a', 'hex')
+
+      var tmp = new Transaction()
+      tmp.addInput(new Buffer('0000000000000000000000000000000000000000000000000000000000000000', 'hex'), 0)
+      tmp.addOutput(randScript, 5000000000)
+      tmp.___toBuffer = tmp.__toBuffer
+      tmp.__toBuffer = function (a, b, c) {
+        if (c !== false) {
+          throw new Error('Not meant to pass true to __toBuffer in hashForSignature')
+        }
+        return this.___toBuffer(a, b, c)
+      }
+
+      assert.throws(function () {
+        tmp.__toBuffer(undefined, undefined, true)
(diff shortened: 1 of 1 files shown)


test/transaction.js
```diff
@@ -205,6 +205,29 @@ describe('Transaction', function () {
   })
 
   describe('hashForSignature', function () {
+    it('only uses V0 serialization', function () {
+      var randScript = Buffer.from('6a', 'hex')
+
+      var tx = new Transaction()
+      tx.addInput(Buffer.from('0000000000000000000000000000000000000000000000000000000000000000', 'hex'), 0)
+      tx.addOutput(randScript, 5000000000)
+      tx.___toBuffer = tx.__toBuffer
```
dcousens: `var tmp` was no good? haha

dcousens:
```
 TypeError: undefined is not a function
      at Context.<anonymous> (test/transaction.js:209:31)
```

afk11:
Fixed

test/transaction.js
```diff
@@ -205,6 +205,29 @@ describe('Transaction', function () {
   })
 
   describe('hashForSignature', function () {
+    it('only uses V0 serialization', function () {
+      var randScript = new Buffer('6a', 'hex')
+
+      var tx = new Transaction()
+      tx.addInput(new Buffer('0000000000000000000000000000000000000000000000000000000000000000', 'hex'), 0)
+      tx.addOutput(randScript, 5000000000)
+      tx.___toBuffer = tx.__toBuffer
```
dcousens: I'd change it myself, but, your hot on the topic.

Could you just use `var tmp` or something that isn't almost visually identical to its predecessor?

afk11: Done :)

dcousens: Hahaha, no not the transaction.  

I meant,  `tx.__original_toBuffer = tx.__toBuffer` or something, rather than just an extra `_`

PR #717, opened 2016-12-07
Transaction: add getWitnessId (returning buffer)

afk11:
As per #715 I've separated out getWitnessId. I went with returning a Buffer this time, converting the id to hex by convention seems to be a convenience measure.. 

Useful to protocol implementors who wish to verify witness commitments, so should be available in some shape or form (commitment specification states that the leaves of the Merkle tree are the wtxid, not the witness buffer)

### DIFF
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -398,6 +398,10 @@ Transaction.prototype.getId = function () {
   return bufferReverse(this.getHash()).toString('hex')
 }
 
+Transaction.prototype.getWitnessId = function () {
+  return bufferReverse(bcrypto.hash256(this.__toBuffer(undefined, undefined, true))).toString('hex')
+}
+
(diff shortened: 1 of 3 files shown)


dcousens:
IMHO, not `3.0.0`

dcousens:
NACK for now, lets re-open in `3.1.0`,  if needed?

afk11:
Why not? seems it's deciding for people what parts of the protocol to implement?

dcousens:
> verify witness commitments

Maybe a test case for this would be good :+1:

afk11:
That could be fun actually, let me look at the witness commitment templates & verification, see if I ca

dcousens:
Going to close this in favour of https://github.com/bitcoinjs/bitcoinjs-lib/pull/722,  which,  if this is needed, will be added there instead.

But I suspect it won't be needed at all.

PR #720, opened 2016-12-10
tests: less ambiguous naming, fix exception check

dcousens:
As pointed out in https://github.com/bitcoinjs/bitcoinjs-lib/pull/716,  this PR just makes the changes I was after.

It also fixes an exception message not being caught.

### DIFF
--- a/test/transaction.js
+++ b/test/transaction.js
@@ -213,27 +213,28 @@ describe('Transaction', function () {
   })
 
   describe('hashForSignature', function () {
-    it('only uses V0 serialization', function () {
+    it('only uses legacy serialization', function () {
       var randScript = new Buffer('6a', 'hex')
 
-      var tmp = new Transaction()
-      tmp.addInput(new Buffer('0000000000000000000000000000000000000000000000000000000000000000', 'hex'), 0)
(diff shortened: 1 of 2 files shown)


test/transaction.js
```diff
@@ -213,27 +213,28 @@ describe('Transaction', function () {
...
+        if (c !== false) throw new Error('hashForSignature MUST pass false')
+
+        return original.call(this, a, b, c)
       }
 
       assert.throws(function () {
-        tmp.__toBuffer(undefined, undefined, true)
-      }, 'Verify our replacement of __toBuffer can lead to an error if using witness')
+        tx.__toBuffer(undefined, undefined, true)
+      }, /hashForSignature MUST pass false/)
```
dcousens: This must be a regex, or it will just be ignored.

test/transaction.js
```diff
@@ -213,27 +213,28 @@ describe('Transaction', function () {
   })
 
   describe('hashForSignature', function () {
-    it('only uses V0 serialization', function () {
+    it('only uses legacy serialization', function () {
```
dcousens: or,  `it does not use Witness V0 serialization`

PR #721, opened 2016-12-10
Add dual-key stealth example

dcousens:
Adds a dual key example,  and hopefully changes the comments to clarify a bit more RE how a leak would affect you, and how the overall DH process works.

I think the notation used in the comments really helps explain the public/private nature of all three functions


dcousens:
ping @fanatid,  could you look over this for validity? (source https://github.com/genjix/bips/blob/master/bip-stealth.mediawiki)

fanatid:
yep, looks like validity by reference

PR #722, opened 2016-12-13
Script templates: add witness commitment template

afk11:


dcousens:
If you could add the necessary function witness commitment hash creation function to `Block`,  tests for that,  then a final integration test that checks segnet,  downloads a segwit block and verifies it... this feature is good to go :+1:

dcousens:
ACK 6c78b36 good to merge,  but will leave PR open for the above

dcousens:
@afk11 can you rebase this?  We'll merge it now and I'll add an issue for:

> Needs an integration test showing how this would be used... and potentially a Block method to build the commitment hash

PR #723, opened 2016-12-13
Clarify its using non-witness, not necessarily legacy

dcousens:
I think implying the previous serialisation methods are `legacy` actually may be the wrong way to go about this.


dcousens:
ping @afk11

afk11:
Maybe we should work out better names for toBuffer / toLegacyBuffer, since it's only possible to have a witness serialization if one of the outputs called for it?

PR #725, opened 2016-12-14
Add TransactionBuilder sign support for P2WPKH

dcousens:
Shouldn't be far off full support for P2WSH too.

Please see/merge https://github.com/bitcoinjs/bitcoinjs-lib/pull/726 first


afk11:
P2WPKH is just a special case for P2PKH (only difference is where you put sigs, and the sigVersion). Is there any way to exploit this more in your PR so there's less code hacking it in?

afk11:
Yay, encodeStack looks good, because once you have the stack you can use compile _not caring what the type of script is_

src/transaction_builder.js
```diff
@@ -14,9 +14,20 @@ var Transaction = require('./transaction')
...
-  var prevOutType = bscript.classifyInput(scriptSigChunks, true)
+function expandInput (scriptSig, redeemScript, witnessStack) {
+  var witnessType
+  if (witnessStack) {
+    witnessType = bscript.classifyWitness(witnessStack)
+  }
+
+  var prevOutType, scriptSigChunks
+  if (scriptSig.length === 0 && witnessStack) {
+    prevOutType = witnessType
```
afk11: You could have expandInput solve these values, and eventually call expandScript (which takes out sigs/pubkeys/nRequiredSigs)

src/transaction_builder.js
```diff
@@ -14,9 +14,20 @@ var Transaction = require('./transaction')
...
-  var scriptSigChunks = bscript.decompile(scriptSig)
-  var prevOutType = bscript.classifyInput(scriptSigChunks, true)
+function expandInput (scriptSig, redeemScript, witnessStack) {
+  var witnessType
+  if (witnessStack) {
+    witnessType = bscript.classifyWitness(witnessStack)
+  }
+
+  var prevOutType, scriptSigChunks
+  if (scriptSig.length === 0 && witnessStack) {
```
afk11: ScriptSig's aren't empty if the transaction is P2SH | P2WPKH - the scriptSig would be a push of the `[OP_0 [20 byte hash]]`

src/transaction_builder.js
```diff
@@ -164,43 +196,55 @@ function prepareInput (input, kpPubKey, redeemScript) {
...
+var EMPTY_SCRIPT = new Buffer(0)
+
 function buildInput (input, allowIncomplete) {
   var signatures = input.signatures
   var scriptType = input.redeemScriptType || input.prevOutType
-  var scriptSig
+  var stack
 
   switch (scriptType) {
+    case scriptTypes.P2WPKH:
```
afk11: The stack encoding of P2WPKH and P2PKH is the same, maybe do something like this? 
```
    case scripttypes.P2WPKH:
    case scriptTypes.P2PKH:
             ...
             break
```

src/transaction_builder.js
```diff
@@ -164,43 +196,55 @@ function prepareInput (input, kpPubKey, redeemScript) {
...
+
     case scriptTypes.P2PKH:
-    case scriptTypes.P2PK:
       if (signatures.length < 1 || !signatures[0]) throw new Error('Not enough signatures provided')
-      if (scriptType === scriptTypes.P2PKH) {
-        scriptSig = bscript.pubKeyHash.input.encode(signatures[0], input.pubKeys[0])
-      } else {
-        scriptSig = bscript.pubKey.input.encode(signatures[0])
-      }
+      stack = bscript.pubKeyHash.input.encodeStack(signatures[0], input.pubKeys[0])
```
dcousens: @afk11 indeed,  this was an artefact of the refactoring process

dcousens:
Rebased

dcousens:
We might need to drop the `compile(x) => x` (if `x` is a `Buffer`) convenience of this stack ambiguity to drop out.

Unfortunately, the optimisation,  while it has been OK to date, if you knew what you were doing,  completely screws us over since a stack is a `[]` and should **not** be considered a decompiled script... in many cases.

Or...

We need to figure out how we are going to deal with the ambiguity in `check`s for `pubKeyHash` and `witnessPubKeyHash`... maybe `checkStack` could be added which forces an Array input?

dcousens:
ping @afk11

afk11:
>  Unfortunately, the optimisation, while it has been OK to date, if you knew what you were doing, completely screws us over since a stack is a [] and should not be considered a decompiled script... in many cases.

Is it that P2WPKH's witness would look too close to a decompiled P2PKH scriptsig for comfort?

junderw:
This PR is milestone 3.1.0... currently at 3.1.1

Wondering if this will be picked back up now that Segwit is looking to activate in a few weeks

PR #726, opened 2016-12-14
Pre-#725 cleanup

dcousens:


src/script.js
```diff
@@ -151,9 +151,9 @@ function decompilePushOnly (script) {
 
   return chunks.map(function (op) {
     if (Buffer.isBuffer(op)) return op
-    if (op === OPS.OP_0) return new Buffer(0)
+    if (op === OPS.OP_0) return op
 
-    return scriptNumber.encode(op - OP_INT_BASE)
+    return op
   })
```
dcousens: ping @afk11 ,  did I miss something here?
Why are OP_INTs being encoded as script numbers?

src/templates/pubkeyhash/input.js
```diff
@@ -1,4 +1,4 @@
-// {pubKey} OP_CHECKSIG
```
dcousens: :8ball:

src/transaction_builder.js
```diff
@@ -358,7 +358,7 @@ TransactionBuilder.prototype.__addInputUnsafe = function (txHash, vout, options)
 
   var vin = this.tx.addInput(txHash, vout, options.sequence, options.scriptSig)
   this.inputs[vin] = input
-  this.prevTxMap[prevTxOut] = true
```
dcousens: this isn't necessary anywhere, but it is certainly nice to have

test/fixtures/transaction.json
```diff
@@ -164,8 +164,7 @@
           {
             "hash": "0000000000000000000000000000000000000000000000000000000000000000",
             "index": 4294967295,
-            "data": "032832051c4d696e656420627920416e74506f6f6c20626a343a45ef0454c5de8d5e5300004e2c0000",
-            "sequence": 4294967295
+            "data": "032832051c4d696e656420627920416e74506f6f6c20626a343a45ef0454c5de8d5e5300004e2c0000"
           }
```
dcousens: All this data was not needed, and just served to confuse me as to why it was in the fixtures.

If we need it, we can add it back later by looking at this commit.

test/fixtures/transaction.json
```diff
@@ -802,7 +672,7 @@
       "value": 600000000
     },
     {
-      "description": "P2SH(P2WPKH) with SIGHASH_ALL",
+      "description": "P2WPKH with SIGHASH_ALL",
       "txHex": "0100000001db6b1b20aa0fd7b23880be2ecbd4a98130974cf4748fb66092ac4d3ceb1a54770100000000feffffff02b8b4eb0b000000001976a914a457b684d7f0d539a46a45bbc043f35b59d0d96388ac0008af2f000000001976a914fd270b1ee6abcaea97fea7ad0402e8bd8ad6d77c88ac92040000",
```
dcousens: P2SH is irrelevant here, it is hashing P2WPKH.
Again, confusing.
Removed.

test/transaction_builder.js
```diff
@@ -354,7 +354,7 @@ describe('TransactionBuilder', function () {
   describe('multisig', function () {
     fixtures.valid.multisig.forEach(function (f) {
       it(f.description, function () {
-        var txb = construct(f, false)
+        var txb = construct(f, true)
         var tx
         var network = NETWORKS[f.network]
```
dcousens: Falsy is simpler

src/script.js
afk11: They have the effect of pushing a vch onto the stack like every other push, but we don't rely on this behavior much (example, I've never once checked the check mulltisig dummy value), but it's the semantics we need.

test/fixtures/transaction.json
afk11: It's all contained in the hex or whex, I put this data here for convenience during testing (it's nice to check the sigHash if you can before signing, etc)

afk11: Just build all these unsigned tx's, and have the tests check it comes up with the same witness tx?

src/script.js
afk11: if you don't do this, you end up with something like: 
decompiled script == [OP_0, Buffer(sig1), Buffer(sig2), Buffer(redeemScript)

but the witness == ['', Buffer(sig1), Buffer(sig2), Buffer(redeemScript)]

which obviously aren't the same. The script stack is the common representation between the two though!

test/fixtures/transaction.json
dcousens: @afk11 if you want to build those test fixtures, go for it! But I'd say we keep them separate from the vectors strictly needed for `hashForWitnessV0`

test/fixtures/transaction.json
dcousens: @afk11 indeed, but `hashForWitnessV0` doesn't care.
It just takes the types of arguments it takes, it doesn't care about the context.
The fixtures should reflect that where possible.

src/script.js
dcousens: https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/src/transaction.js#L449 - maybe we should just be putting the script number encoding code there?

dcousens:
@afk11 / @rubensayshi please see 59a878e

src/script.js
```diff
@@ -145,23 +145,18 @@ function fromASM (asm) {
...
 
   return chunks.map(function (op) {
     if (Buffer.isBuffer(op)) return op
-    if (op === OPS.OP_0) return op
+    if (op === OPS.OP_0) return new Buffer(0)
 
-    return op
+    return scriptNumber.encode(op - OP_INT_BASE)
   })
 }
```
dcousens: There is no such thing as `fromStack`... its a 1 way process because you can't tell the difference between a data buffer and what may have been originally an OP_INT.

dcousens: Sort of,  I guess in our case most of the standard templates ARE push only with NO OP_INTs... but certainly its one of those cases that is difficult to determine what `fromStack` would/should return...

dcousens: It also leaves up for question when we decide to decode as script number compared to a `Buffer`...

PR #730, opened 2016-12-15
Test data removal and rename

dcousens:


afk11:
NACK, tt's not unused, we don't have the code to support it yet. We're going to need this stuff soon.

afk11:
I think we're going to use this quite soon - we can build those transactions from unsigned and sign ourselves to check we're still compatible. Since I added them here I've also incorporated them elsewhere: https://github.com/Bit-Wasp/bitcoin-php/blob/master/tests/Transaction/Factory/SignerTest.php#L219

You can test so much with these vectors!

dcousens:
@afk11 if these vectors are "official",  maybe they should be seperate.
But I try to avoid "generic" all-round vectors,  because it makes it harder to add extra tests to account for just 1 branch when you have to add extra data to satisfy all the other tests.

afk11:
Aye I accept that point of view, it can be a pain, I've also ended up just doing `if (f.xx)` to compensate. I can put them somewhere else, it's probably better we have a consistent set to go all out on :)

afk11:
Hmm, could be good to leave what remains in transaction.json, but can you add transaction_sign.json or something with the full data from 'P2PK' to 'P2SH P2WPKH' ?

dcousens:
Weren't these official vectors in some way?
Put them in [`tests/fixtures/core`](https://github.com/bitcoinjs/bitcoinjs-lib/tree/master/test/fixtures/core)?

dcousens:
@afk11 `git checkout 08ca1e73c2ed713dce5d1d317cc90f94f8e6c839~1` to get the data you need,  then put it in a file, when it is needed.

PR #731, opened 2016-12-15
Remove of pushOnly, added toStack

dcousens:
Non-breaking change as *pushOnly wasn't released yet.


afk11:
Looks good, though on review https://github.com/bitcoin/bips/blob/master/bip-0062.mediawiki#push-operators mentions OP_1NEGATE which is missed in `compile`, so compile(toStack()) mightn't always hold. Annoyingly, it's 0x81 on the stack so it's a special case to be added to compile if we're bothered.

dcousens:
> compile(toStack()) mightn't always hold

What do you mean by that?

afk11:
The current bscript.compile is going to end up encoding OP_1NEGATE as 0x0181, instead of using it's minimal encoding of 0x4f, so `compile` needs something to support that.

dcousens:
Closing in favour of https://github.com/bitcoinjs/bitcoinjs-lib/pull/732

PR #735, opened 2016-12-16
Segwit: add buildStack

afk11:
Adds buildScript which avoids recursive methods which try to solve the input while trying to serialize. After expanding/signing you already know all of this, so we use some light code that decides where to put the results of buildScript.

It's kind of weird trying to do anything without making this assumption - the alternative is stuff like `var script=witnessScript||redeemScript||prevOutScript`


afk11:
As mentioned, I can remove some code from that function once calling prepareInput guarantees that the input will be signable
 * that the txOut script is provided
 * if it's bare, it can only be [P2WPKH, P2WSH, P2SH, P2PKH, P2PK, Multisig]
 * if the txOut script calls for a P2SH, the redeemScript must be provided, and of [P2WPKH, P2WSH, P2PKH, P2PK, Multisig] ie, standard types and witness types. never p2sh.
 * if the txOut script calls for P2WSH, the witnessScript must be provided, and of [P2PKH, P2PK, Multisig] (ie, same types as standard-bare scriptPubKey)

I found changing expandOutput / prepareInput to accommodate this to be a real challenge (without the txOut), but would really suit the rest of signing.

src/transaction_builder.js
```diff
@@ -462,15 +467,17 @@ TransactionBuilder.prototype.__build = function (allowIncomplete) {
...
     var result = buildInput(input, allowIncomplete)
 
     // skip if no result
-    if (!result && allowIncomplete) return
-    if (result && result.script) return tx.setInputScript(i, result.script)
-    if (result && result.witness) return tx.setWitness(i, result.witness)
+    if (!allowIncomplete) {
+      if (SIGNABLE.indexOf(result.type) === -1 && result.type !== bscript.types.P2WPKH) {
+        throw new Error(result.type + ' not supported')
+      }
```
dcousens: Why do we need this?

src/transaction_builder.js
```diff
@@ -224,72 +226,75 @@ function prepareInput (input, kpPubKey, redeemScript) {
...
-    case scriptTypes.MULTISIG:
-      signatures = signatures.map(function (signature) {
-        return signature || ops.OP_0
-      })
-
-      if (!allowIncomplete) {
-        // remove blank signatures
-        signatures = signatures.filter(function (x) { return x !== ops.OP_0 })
-      }
+function buildScript (type, signatures, pubKeys, allowIncomplete) {
```
dcousens: `buildStack` more like

afk11: +1 that would be more correct

src/transaction_builder.js
afk11: Normally we'd end up with [p2pkh, p2pk, multisig], but p2wpkh is just p2pkh, so when saving our knowledge about the inner-most script, we set the P2PKH solution (yep, solveOutput) for scriptCode if it's P2WPKH.

dcousens: Man,  I know we both agree the complexity explosion in this area has always sucked, and made it so hard to review changes by any one.

Its all just so gross.

afk11:
Bad rebase, easier to resubmit

PR #738, opened 2016-12-17
script: add fromASM failing test case and fix

dcousens:


dcousens:
In favour of https://github.com/bitcoinjs/bitcoinjs-lib/pull/690

PR #741, opened 2016-12-21
package: bump bs58check, drop node 0.12

dcousens:


fanatid:
Why also not use `Buffer.from`/`Buffer.alloc`?

dcousens:
@fanatid changes take time,  time I didn't have when I was bumping this dependency in a project that needed it :+1: 

If you can change them over,  absolutely go for it,  otherwise I'll do it asap

dcousens:
After reaching out,  even companies on `bitcoinjs-lib < 2.0.0` are using node 4... so,  I'm happy with this.

PR #743, opened 2016-12-22
Segwit: buildStack for signatures

afk11:

PR #746, opened 2016-12-29
Segwit P2WSH

afk11:
I rewrote prepareInput to tried and adhere to the same semantics as before, but covers P2SH and witness types, and bare witness types. 

Final thing needed here is tests, I'll dig up that set I generated on testnet.

This PR is based on #743, which is based on #725 but adds buildInput + test fixes.


afk11:
refactoring expandInput now to get TransactionBuilder.fromTransaction working with P2WSH

afk11:
Remembered your comment here @dcousens https://github.com/bitcoinjs/bitcoinjs-lib/pull/725/files#r92329243
definitely the bit I'm having trouble with now!

afk11:
Ok, some of these functions aren't in a nice way, but whoop it works and tests pass! I'd say this is as good as it gets BC-wise.

Once this is merged, I strongly recommend we look at:
 * options break - motivation for this is to avoid the ever increasing number of signing parameters (witnessScript, txout value, redeemScript)
 * txOut break - absolutely always require the txOut, makes it easier to check extra `opts`, or data already in the transaction.

Not necessarily at the same time - I believe txOut now really *should* be maintained as a separate structure to `opts`. There is no need for opts when the transaction is fully/part signed - the necessary data would be embedded, whereas a builder initialized with fromTransaction could be missing parts. 

I did get as far as a separate class for input signer before - I think it's an idea worth revisiting, since it means the internal signing api gets simpler:

+-------------------+------------------+
|            |         |
+--------------------+------------------+

afk11:
```bash
 1) script-templates classifyInput classifies 000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000 as nonstandard:

      AssertionError: 'scripthash' === 'nonstandard'
      + expected - actual

      -scripthash
      +nonstandard
      
      at Context.<anonymous> (test/templates.js:19:16)

  2) script-templates scriptHash.input.check returns false for 000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000:

      AssertionError: true === false
      + expected - actual

      -true
      +false
```

afk11:
Needs more tests

afk11:
Segwit's working fully now, and added a tonne of test vectors

afk11:
@pernas this should be working, want to give this a go?

pernas:
Nice, looks that now is working. I produced that transaction p2wpkh into p2sh

https://www.blocktrail.com/tBTC/tx/9b96df06f06dbea8fdb8a3bf1d1787c8205695507adfd9d1f1f9312dad6600a7

01000000000101c70f1761d0f179867de6d554d12f345f2f4f2ff5296ccc0a212382df15b56b580100000017160014b1a819968646a6c80db62a0172a1aea3a93f01d7ffffffff01905f01000000000017a91420028c4f260ea7b7814570f2b771be1d056f9be58702483045022100e7bb84d2b4f06c31984e5395751ca5e00312833d43f65f47b23a0e5a2cd6f522022015357ccabc3ef01049f4ad84cbb936215c2d75d2acde0b2d13ad594604e77712012103cdc2e40e331be000c4a599baf93823796197dd07d0a8e8623da1ccf32d94757600000000

dcousens:
Will review this week :+1:

afk11:
Latest commit fixes an issue with serializing P2WPKH tx's when they weren't fully signed. Has me thinking a nice test would be to keep track of signatures and stuff as we call sign, and check that they're getting applied.

src/transaction_builder.js
```diff
@@ -339,6 +338,10 @@ function prepareInput (input, kpPubKey, redeemScript, witnessValue, witnessScrip
     signScript = prevOutScript
   }
 
+  if (witness && !types.Satoshi(witnessValue)) {
+    throw new Error('Input was witness but not given witness value')
```
dcousens: ACK

dcousens:
This is looking good to merge :+1: ,  just need to test it a bit myself

dcousens:
@afk11 I can't review this further at this time,  so I'm going to leave merge responsibility on you,  lets see if we don't get some feedback based on that.

@fanatid if you could overlook it that'd be great too.

src/transaction_builder.js
dakk: WitnessValue won't be mandatory if Transaction.addInput provides the value, right? (In some use cases the signer has no informations about the unspent value he's signing)

afk11: Transaction.addInput probably shouldn't know the input value, since serializing would discard it. That information must _always_ be known by the signer (or even for extracting signatures) since otherwise they would fail to produce the sigHash. 

Requiring the scriptpubkey/value is a mission for signing operation is a mission for me atm, so the signer should always have this information

dcousens: @afk11 it is worth capturing that information though.

dcousens: See https://github.com/bitcoinjs/bitcoinjs-lib/pull/882

dakk: got it

PR #748, opened 2017-01-03
Experimental TxSigner abstraction

afk11:
For posterity, this is a brand new transaction signer. It assumes that transaction building is separate, so is only ever initialized by a transaction. 

It works by defering all script handling until everything is fully classified. it also *requires* the txOut from the caller. This way, all assumptions are checked against information known from the time the address was created. It's mostly to give people ideas, as it has a lot of similarities to the current transaction_builder, but some differences also.

It currently handles P2PKH, P2PK, Multisig in any representation, and to extend this support, barely anything new has to be added (actually, it's the script-hash commitments working in tandem that's the hard part)

No effort on tests this time, this is to give people ideas about what the next signer could look like. Although it's not tested, it does work nicely.


dcousens:
Needs rebase

afk11:
Had some notes looking over this one

 * Adding the `txOut` parameter to TransactionBuilder.sign is necessary (extra param)
 * Consolidate solve & extraction functions to allow optional opts (w/ txout, already know where to find things), (fewer required params for part signed txs, data already there)
 * Compartmentalizing the input signing means (nin, txOut, [opts, or: rs, ws, value]) all disappear as arguments to it's signing function. nice for multisig users (only parameter is the key/sigHashType)
 * Why incapsulate input signing again? creating the instance is our chance to say it's unsignable (no logic required later for canSign).. also allows us to expose some interesting per-input functions 
 * An opts instance isn't utterly necessary, but it does make the TransactionBuilder.sign API nicer.. (optional arguments no longer positional; proposals like MAST warrant more parameters)

I can find some time this week to look at the first steps of applying some of this to master,

dcousens:
@afk11 what were your thoughts on this compared to a BIP174 approach?

PR #750, opened 2017-01-04
node 4.0.0 or higher

dcousens:
Ref https://github.com/bitcoinjs/bitcoinjs-lib/pull/741

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "Client-side Bitcoin JavaScript library",
   "main": "./src/index.js",
   "engines": {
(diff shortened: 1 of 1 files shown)


rubensayshi:
is this based on any particular adoption charts or something of Node versions?

dcousens:
@rubensayshi no it was based on switching to `Buffer.from` as a safety precaution standard... which we still need to do throughout.

dcousens:
@rubensayshi that said,  if any such chart/data exists... I'd love to know.

PR #770, opened 2017-03-28
Added a note to the "Browser" section of README.md.

### LINKED ISSUE #769: Invalid checksum on iOS devices
I'm currently receiving the following error on an iOS device for a React Native project ("Invalid checksum"). The error is triggered when attempting to use the following:
```
const HDNode = require("./bitcoinjs-lib/src/hdnode");
const xpub = "xpub6AoynT2YixgjHK2yNTDPJGmbFaNfiKTVNb4zVygVmSfhWktotj8woL2nT7FALSFEhggNemiy5KEGWxRpNYtsHVVvWCC99WuhzvAvZTJD1SH";
let node = HDNode.fromBase58(xpub);
```
The error appears to be coming from `base58check.decode` where `decodeRaw` doesn't pass the checksum check. 

One important thing to note is that this error does not occur within the iOS simulator, Android emulator or on an Android device. It appears to be isolated to iOS devices only. Does anyone have any thoughts as to why this error seems to only pop up on iOS devices?

screen shot 2017-03-25 at 4 42 32 pm

coreyphillips:
This closes #769.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,8 @@ Example:
 uglifyjs ... --mangle --reserved 'Array,BigInteger,Boolean,ECPair,Function,Number,Point'
 ```
 
+**NOTE**: If you expect this library to run on an iOS 10 device, ensure that you are using [buffer@5.0.5](https://www.npmjs.com/package/buffer) or greater. 
+
 ### Flow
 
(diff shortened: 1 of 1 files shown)


dcousens:
Thanks @corey-phillips

PR #771, opened 2017-03-28
block: use writeUInt32BE for target calculation

dcousens:


src/block.js
```diff
@@ -131,16 +131,11 @@ Block.prototype.toHex = function (headersOnly) {
...
 
   var target = new Buffer(32)
   target.fill(0)
-
-  target[i] = mantissa & 0xff
-  target[i - 1] = mantissa >> 8
-  target[i - 2] = mantissa >> 16
-  target[i - 3] = mantissa >> 24
-
+  target.writeUInt32BE(mantissa, i)
```
fanatid: why not put `28 - exponent` here instead `i`? in any case, nice solution!

dcousens: done

PR #783, opened 2017-04-15
Decred integration

peterzen:
This adds support for Decred (DCR)  altcoin: https://www.decred.org

### DIFF
--- a/src/networks.js
+++ b/src/networks.js
@@ -31,5 +31,15 @@ module.exports = {
     pubKeyHash: 0x30,
     scriptHash: 0x05,
     wif: 0xb0
-  }
+  },
+  decred: {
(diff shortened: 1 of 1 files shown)


dcousens:
Wouldn't this fail because of the reasons you pointed out in `bs58check`?

fanatid:
@peterzen for network data you can use [coininfo](https://github.com/cryptocoinjs/coininfo/):
```js
const = require('coininfo')
const mainnet = coininfo.decred.main.toBitcoinJS()
```

@dcousens there should be too much changes for Decred, Transaction encoding/new Opcodes... you can check required changes here https://github.com/decred/bitcore/commits/master-new

I think this should be closed.

PR #784, opened 2017-04-18
Add Block.prototype.byteLength

dcousens:


dcousens:
Necessary for https://github.com/bitcoinjs/bitcoinjs-lib/issues/513,  but I don't have time for that yet so this covers off on the use cases I have.

dcousens:
POST ACK should be OK,  as the feature is pretty uncontroversial

fanatid:
ACK

PR #790, opened 2017-04-19
Safe Buffer usage

dcousens:
@fanatid review would be appreciated :+1: .

Non-controversial,  but, sensitive and I'd like at least 2 extra eyes since it touches the ECDSA module.

This is **not** a breaking change (we're already at a minimum of Node 4.5.0)

### DIFF
--- a/src/address.js
+++ b/src/address.js
@@ -1,3 +1,4 @@
+var Buffer = require('safe-buffer').Buffer
 var bs58check = require('bs58check')
 var bscript = require('./script')
 var networks = require('./networks')
--- a/src/address.js
+++ b/src/address.js
@@ -9,7 +9,7 @@ function fromBase58Check (address) {
   if (payload.length < 21) throw new TypeError(address + ' is too short')
   if (payload.length > 21) throw new TypeError(address + ' is too long')
 
-  var version = payload[0]
+  var version = payload.readUInt8(0)
   var hash = payload.slice(1)
 
   return { hash: hash, version: version }
@@ -18,7 +18,7 @@ function fromBase58Check (address) {
 function toBase58Check (hash, version) {
   typeforce(types.tuple(types.Hash160bit, types.UInt8), arguments)
 
-  var payload = new Buffer(21)
+  var payload = Buffer.allocUnsafe(21)
   payload.writeUInt8(version, 0)
   hash.copy(payload, 1)
 
--- a/src/script_number.js
+++ b/src/script_number.js
@@ -1,3 +1,5 @@
+var Buffer = require('safe-buffer').Buffer
+
 function decode (buffer, maxLength, minimal) {
   maxLength = maxLength || 4
   minimal = minimal === undefined ? true : minimal
--- a/src/script_number.js
+++ b/src/script_number.js
@@ -16,8 +16,8 @@ function decode (buffer, maxLength, minimal) {
     var a = buffer.readUInt32LE(0)
     var b = buffer.readUInt8(4)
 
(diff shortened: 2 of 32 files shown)


fanatid:
I like `from`/`alloc`/`allocUnsafe`, but this is not supported by node < 4.5.0 :(
do you sure about this?

dcousens:
@fanatid we already bumped to `4.5.0` minimum with `3.0.0` release whether we like it or not,  as we use `bs58check`@2.0.0 ... we could revert that if necessary, but doesn't change the fact that this PR, right now,  isn't a breaking change.

dcousens:
Ref https://github.com/bitcoinjs/bitcoinjs-lib/pull/741

src/script_number.js
```diff
@@ -16,8 +16,8 @@ function decode (buffer, maxLength, minimal) {
     var a = buffer.readUInt32LE(0)
     var b = buffer.readUInt8(4)
 
-    if (b & 0x80) return -((b & ~0x80) * 0x100000000 + a)
-    return b * 0x100000000 + a
+    if (b & 0x80) return -(((b & ~0x80) * 0x100000000) + a)
+    return (b * 0x100000000) + a
```
fanatid: extra parentheses for readable?

dcousens: `standard`@latest warns on these, haven't bumped yet in repository though

fanatid:
Currently `engines.node` is `>=4.0.0`, but all nodes < 4.5.0 not support from/alloc/allocUnsafe ...
https://github.com/bitcoinjs/bitcoinjs-lib/blob/v3.0.0/package.json#L7

src/address.js
```diff
@@ -18,7 +18,7 @@ function fromBase58Check (address) {
 function toBase58Check (hash, version) {
   typeforce(types.tuple(types.Hash160bit, types.UInt8), arguments)
 
-  var payload = new Buffer(21)
+  var payload = Buffer.allocUnsafe(21)
```
fanatid: I see two variants of this PR: `Buffer.alloc(length)` and `Buffer.allocUnsafe(length)`. Can you choose one?

dcousens: @fanatid I've used `Buffer.alloc` all through the tests... and `Buffer.allocUnsafe` in parts of the code where were previously using `new Buffer`; so no change.

We should probably only need need `Buffer.alloc` in cases of user input no?

dcousens:
Wait... `bs58check` was breaking because it relied on `bs58` which had the `Buffer.from` introduced... but that has since been reverted and we only lost `0.12`... so actually this **is** a breaking change.

Blimey.
Waiting on travis to confirm.

fanatid:
travis will be fine, because `4` there is a link to latest 4.x

fanatid:
base-x was bumped correctly https://github.com/cryptocoinjs/base-x/blob/v3.0.0/package.json#L41

dcousens:
@fanatid only [after a screw around](https://github.com/cryptocoinjs/base-x/issues/35#issuecomment-279305099).
Hence my confusion; and why yes,  **this is a breaking change**.

src/address.js
fanatid: `allocUnsafe` (from docs) uses internal memory pool, maybe we should use `alloc` everywhere?

dcousens: I think this code is tested well enough that we can use both?

dcousens:
Re-targeting `3.0.0` by using `safe-buffer`

dcousens:
Rebased,  using `safe-buffer` and good to merge for a patch release

PR #796, opened 2017-04-22
Update Litecoin P2SH

losh11:
### Litecoin Core 0.13.3 (reference client)

https://github.com/litecoin-project/litecoin/blob/master/src/chainparams.cpp#L129

### LoafWallet (iOS Wallet)

https://github.com/litecoin-association/LoafWallet/blob/master/BreadWallet/NSString%2BBitcoin.h#L30

### DIFF
--- a/src/networks.js
+++ b/src/networks.js
@@ -29,7 +29,7 @@ module.exports = {
       private: 0x019d9cfe
     },
     pubKeyHash: 0x30,
-    scriptHash: 0x05,
+    scriptHash: 0x50,
     wif: 0xb0
   }
 }
--- a/src/networks.js
+++ b/src/networks.js
@@ -29,7 +29,7 @@ module.exports = {
       private: 0x019d9cfe
     },
     pubKeyHash: 0x30,
-    scriptHash: 0x50,
+    scriptHash: 0x32,
     wif: 0xb0
   }
 }
--- a/test/fixtures/address.json
+++ b/test/fixtures/address.json
@@ -15,6 +15,20 @@
       "script": "OP_HASH160 cd7b44d0b03f2d026d1e586d7ae18903b0d385f6 OP_EQUAL"
     },
     {
+      "network": "litecoin",
+      "version": 50,
+      "hash": "332bdfb31f688c0be0137c7c038a6d0fea0de0b6",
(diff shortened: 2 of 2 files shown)


dcousens:
Do you have a citation or reference?
And can you provide some test fixtures maybe?

losh11:
Sorry, I was about to do that, and got disturbed by stuff.

Sure, I can fix tests, but all tests seem to pass on my machine `2795 passing (12s)`.

dcousens:
@losh11 then we need tests! :+1:

src/networks.js
```diff
@@ -29,7 +29,7 @@ module.exports = {
       private: 0x019d9cfe
     },
     pubKeyHash: 0x30,
-    scriptHash: 0x05,
+    scriptHash: 0x50,
```
junderw: `0x32` ?

The link in litecoin is base-10

dcousens:
Not going to merge until we have some example Litecoin P2SH addresses added to tests to ensure the constant is actually valid...

dcousens:
@junderw @losh11 if either of you have example Litecoin P2SH addresses... please either update this PR,  submit a different PR, or just post them.

afk11:
Someone grab a LTC P2SH address and run base58 decode on it?

@losh11 as mentioned, the constant should be expressed as 0x32 or 50, but not 0x50.

afk11:
Guessing everyones happy, whoever second passes this please merge?

dcousens:
If someone just adds fixtures... We're good to merge.

afk11:
will do!

losh11:
That's great guys! Sorry got really busy, so I didn't notice this...

dcousens:
Published as `3.0.3`

rplopes:
This change makes it impossible to work with legacy P2SH litecoin addresses, which will still show up in live systems.

Is there a recommended approach for dealing with such valid legacy transaction versions? I'm guessing adding a `legacyScriptHash` property to all networks might not be the best long term strategy. Adding a `litecoinLegacy` network wouldn't work, as a transaction builder only accepts one network, and a transaction can mix the old and new addresses.

If it's something that is not possible to do, would it be something worthy to build support for?

PR #800, opened 2017-05-25
fix typo in Readme

eliyastein:
The link under examples in the Readme for "Generate a address and WIF for Litecoin" has a typo in it that goes to github 404. Here's the fix.


dcousens:
Thanks

Issue #521, opened 2016-01-20
Recognize an opt-in RBF transaction


Sjors:
Opt-in Replace By Fee is part of [Core 0.12](https://github.com/bitcoin/bitcoin/blob/0.12/doc/release-notes.md).

I'm still reading up on this, but it seems useful to detect if a given transaction if RBF-able. Someone has already done a [proof of concept](https://gist.github.com/anonymous/b935887caec15849cb15). 

You also have to apply this check to every unconfirmed parent transaction. A fancy method could ask for each of those, like so:

```js
const getParentTxIfUnconfirmed = (hash) -> {
   // lookup transaction somewhere
   const tx = ....;
   if(tx.confirmed) {
     return nil;
   } else {
     return tx.toHex();
   }
}

Transaction.fromHex(...).isRbf(getParentTxIfUnconfirmed).then((res) ->
   if(res) {
     // Warn user
   }
)
```

dcousens:
Detecting RBF isn't difficult,  the question is,  at what part of this code base is it relevant?
Perhaps we should make a separate module for it?

Its policy to be honest,  and policy varies with the wind.

dcousens:
@Sjors did you have any ideas how you'd like to see this integrated with this library?

jprichardson:
> Perhaps we should make a separate module for it?

:+1:

dcousens:
This is policy related code.
In reading what was discussed (even if only by myself) here: https://github.com/bitcoinjs/bitcoinjs-lib/pull/617#issuecomment-243666879

I think it is safe to say our "policy" on policy related code is the following:

* `bitcoinjs.script.is*` functions adhere to the community recognized script templates relating to their type,  this is a matter of localized consensus,  but I think it is safe to say this doesn't change much.
They are certainly just pattern matching templates,  and should be treated as such.
As bitcoin-core is a great reference for these templates,  we will keep them "in-line" with bitcoin-cores policy in how they pattern match.
This is consistent with our changes to date.

* `bitcoinjs.ECPair.sign` ensures that we create signatures that adhere to the `LOW_S` anti-malleability bitcoin-core policy (which is not soft-forked [yet](https://github.com/bitcoin/bitcoin/pull/8533)),  certainly our verification code still allows `LOW_S` signatures however.
This is probably not very consistent,  but I certainly wouldn't want it changed...

[...]

PR #814, opened 2017-07-01
Starting rbf

runn1ng:
First shot at implementing RBF and RBF detection.

The problem is that logically, RBF belongs to the transaction, but the signalling is in the inputs.

### DIFF
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -218,6 +218,29 @@ Transaction.prototype.__byteLength = function (__allowWitness) {
   )
 }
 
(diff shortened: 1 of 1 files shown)


runn1ng:
Oh, it seems you already discussed RBF here, and it seems like your decision was to not implement it, if I understand correctly?

https://github.com/bitcoinjs/bitcoinjs-lib/issues/521

runn1ng:
see discussion #521 , closing here, will make a new module later (maybe)

dcousens:
This code looks like a good candidate for that module though. :+1:

PR #823, opened 2017-07-20
remove errant argument to buildInput in __build

Melvillian:
`buildInput` was passing `buildStack` an incorrect number of arguments. Based on the call signature of `buildStack`, I removed `input.script`.

This caused the "Incomplete transaction w/ prevTxScript defined" test to fail because it no longer threw an exception. However I don't think that exception should be thrown, since if `allowIncomplete` is set to `true` then a lack of signatures shouldn't be a problem.

### DIFF
--- a/src/transaction_builder.js
+++ b/src/transaction_builder.js
@@ -397,7 +397,7 @@ function buildInput (input, allowIncomplete) {
   var sig = []
   var witness = []
   if (SIGNABLE.indexOf(scriptType) !== -1) {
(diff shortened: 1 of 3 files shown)


dcousens:
>  However I don't think that exception should be thrown, since if allowIncomplete is set to true then a lack of signatures shouldn't be a problem.

100% agreed.

Thanks!

dcousens:
Re-added 0b9b30a based on its original intent (see https://github.com/bitcoinjs/bitcoinjs-lib/commit/4c9fd6010eeb5bf364bb101466c8781e7d312bd8).

It simply shouldn't have had `incomplete: true` set.

PR #826, opened 2017-07-27
Add BitcoinCash sighash support

junderw:
This should help anyone building apps based off bitcoinjs-lib support Bitcoin Cash.

### DIFF
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -319,7 +319,7 @@ Transaction.prototype.hashForSignature = function (inIndex, prevOutScript, hashT
     // which uses a fork id of 0. 0 | anything = anything,
     // so bitcoin cash's replay protection is simply unaltered
     // segwit instead of legacy signature-hashing.
-    return this.hashForWitnessV0(inIndex, prevOutScript, hashType, inAmount)
+    return this.hashForWitnessV0(inIndex, prevOutScript, inAmount, hashType)
   } else { // if not BitcoinCash BIP143
     // serialize and hash
     var buffer = Buffer.allocUnsafe(txTmp.__byteLength(false) + 4)
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -315,6 +315,10 @@ Transaction.prototype.hashForSignature = function (inIndex, prevOutScript, hashT
       throw new Error('Bitcoin Cash sighash requires value of input to be signed.')
     }
 
+    // this abstraction only holds for Bitcoin Cash,
+    // which uses a fork id of 0. 0 | anything = anything,
+    // so bitcoin cash's replay protection is simply unaltered
+    // segwit instead of legacy signature-hashing.
     return this.hashForWitnessV0(inIndex, prevOutScript, hashType, inAmount)
   } else { // if not BitcoinCash BIP143
     // serialize and hash
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -32,6 +32,7 @@ Transaction.DEFAULT_SEQUENCE = 0xffffffff
 Transaction.SIGHASH_ALL = 0x01
 Transaction.SIGHASH_NONE = 0x02
 Transaction.SIGHASH_SINGLE = 0x03
+Transaction.SIGHASH_BITCOINCASHBIP143 = 0x40
 Transaction.SIGHASH_ANYONECANPAY = 0x80
 Transaction.ADVANCED_TRANSACTION_MARKER = 0x00
 Transaction.ADVANCED_TRANSACTION_FLAG = 0x01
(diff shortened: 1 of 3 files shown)


dcousens:
Personal NACK,  but will leave open as a reference for others to review.

I don't want this to become a political PR,  the feature is clear,  whether it is useful will depend on events spanning the next few weeks,  and therefore I'd rather defer any action until then.

dcousens:
Needs tests

src/transaction.js
```diff
@@ -251,8 +252,8 @@ Transaction.prototype.clone = function () {
  * hashType, and then hashes the result.
  * This hash can then be used to sign the provided transaction input.
  */
-Transaction.prototype.hashForSignature = function (inIndex, prevOutScript, hashType) {
-  typeforce(types.tuple(types.UInt32, types.Buffer, /* types.UInt8 */ types.Number), arguments)
+Transaction.prototype.hashForSignature = function (inIndex, prevOutScript, hashType, inAmount) {
```
dcousens: Maybe it'd be better to take the approach used for SegWit,  a different function not `hashForSignature`,  avoids positional parameters

junderw:
ToDo:

1. Write tests.
2. Move to separate function
3. Change the fixtures in core/sighash.json to not accidentally trigger 0x40 masking for tests.

I will look into brushing things up once BCC is a clear fixture in the community. For now I'll leave this PR on the back burner.

Thanks for the comments.

src/transaction.js
junderw: I agree.

junderw:
Other problems:

This breaks extractInput for multisigs with partial signatures. As the function for fixing order of sigs requires generating the hash for signing, which means using TransactionBuilder.fromTransaction is not possible.

ie. 3-of-5 multisig rawtx that only contains 2 sigs etc.

src/transaction.js
```diff
@@ -308,10 +309,96 @@ Transaction.prototype.hashForSignature = function (inIndex, prevOutScript, hashT
...
+    var outputsBufOffset = 0
+    outputsBufferArray.forEach(function(outputBuffer) {
+      outputBuffer.copy(outputsBuf, outputsBufOffset)
+      outputsBufOffset += outputBuffer.length
+    })
+
+    // HASH256 to get the hashPrevouts, hashSequence, and hashOutputs
+    var hashPrevouts = bcrypto.hash256(prevOutsBuf)
+    var hashSequence = bcrypto.hash256(inputSequencesBuf)
+    var hashOutputs = bcrypto.hash256(outputsBuf)
```
afk11: This all comes from BIP143 - could you maybe use the functions from that part instead of duplicating the code again here?

jprichardson:
NACK on my end as well. Couldn't have said it any better than this https://github.com/bitcoinjs/bitcoinjs-lib/pull/826#issuecomment-318262587

afk11:
@junderw @dcousens https://github.com/afk11/bitcoinjs-lib/commit/abd54be36dcf761e0eacd7a132f26cf58f6b34e5

Since the reference implementation performs the call like this (within SignatureHash) it seems appropriate to abstract this logic since bitcoinjs still supports segwit. 

The change as is breaks existing tests (signature hash fixtures came from fuzzing, so are specific to the implementation being exercised (Bitcoin Core without the hard fork). @junderw if you have any test fixtures, I'm sure they'd be appreciated.

Aside: deferring comment on inclusion until we see exactly where this is going.. But I do have a need for this code right now.

junderw:
merged commits from @afk11 via Github UI.

Thanks a bunch.

I agree that this probably shouldn't be merged into master, but rather into another branch maybe... but since most people will come to github and search for "BitcoinCash" under your repo, having this pull request has some meaning imo.

I'll keep my political opinion off here, but I think it helps people of both the anti and pro side to make it easier for BCC users to split off from the community and go do their own thing.

afk11:
@junderw have you got a test case working with this yet? 

trying to sign a transaction (on regtest) and broadcast it but somehow my signature is not correct, added a bunch of debug lines to both bitcoin-abc and bitcoinjs and the signaturehash is the same, just somehow ... the signatures are not!

afk11:
Oops - that patch I introduced to bitcoin-abc was the problem, not my branch xD

https://github.com/afk11/bitcoinjs-lib/tree/hash4cash works for a P2PKH ALL|FORKID. The test case can be reproduced with this:
bitcoin-cli -regtest signrawtransaction 020000000113aaf49280ba92bddfcbdc30d6c7501c2575e4a80f539236df233f9218a2c8400000000000ffffffff0100f2052a010000001976a914cd29cc97826c37281ac61301e4d5ed374770585688ac00000000 '[]' 'null' 'ALL|FORKID'
```
{
  "hex": "020000000113aaf49280ba92bddfcbdc30d6c7501c2575e4a80f539236df233f9218a2c8400000000049483045022100c5874e39da4dd427d35e24792bf31dcd63c25684deec66b426271b4043e21c3002201bfdc0621ad4237e8db05aa6cad69f3d5ab4ae32ebb2048f65b12165da6cc69341ffffffff0100f2052a010000001976a914cd29cc97826c37281ac61301e4d5ed374770585688ac00000000",
  "complete": true
}
```

dcousens:
@junderw close in favour of #831?

junderw:
Yes. Close in favor of #831

Issue #824, opened 2017-07-21
Add BIP49 tests and example


iancoleman:
Does bitcoinjs-lib support BIP49 (Derivation scheme for P2WPKH-nested-in-P2SH based accounts)?

If so, tests for the bip49 vectors would be great to have, and make for a useful example.

https://github.com/bitcoin/bips/blob/master/bip-0049.mediawiki#test-vectors

dcousens:
> Does bitcoinjs-lib support BIP49
`bitcoinjs-lib` will likely not *directly* support this,  but it could definitely be a new module under the bitcoinjs-lib organization.

This isn't any specific opinion of BIP49,  simply that we even want to move BIP39 to [bitcoinjs/bip32](https://github.com/bitcoinjs/bip32].

I suspect,  given that BIP49 is a derivation scheme for BIP32,  you'll sooner see that scheme in the above or [bitcoinjs/bip32-utils](https://github.com/bitcoinjs/bip32-utils).

dcousens:
@iancoleman that said,  an integration example could definitely be added to showcase how you might do it as is. :+1:

dcousens:
Closing in favour of https://github.com/bitcoinjs/bitcoinjs-lib/pull/840

PR #840, opened 2017-08-05
add testnet bip49 example

iancoleman:
Based on [BIP49 Test Vectors](https://github.com/bitcoin/bips/blob/master/bip-0049.mediawiki#test-vectors) and addresses https://github.com/bitcoinjs/bitcoinjs-lib/issues/824.

The readme is not updated with a link to this example yet.


kabutosamourai:
ACK

PR #841, opened 2017-08-06
WIP: Implement segwit_address.js against bech32 commit

afk11:
WIP, but about time for feedback - this moves the segwit address logic from bech32 to this library, and implements the encode/decode functions with some tests.

Related: https://github.com/bitcoinjs/bech32/pull/8 (the PR uses this branch)

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
   ],
   "dependencies": {
     "bigi": "^1.4.0",
+    "bech32": "git@github.com:bitcoinjs/bech32.git#81bc22761b3c8d286d6d8dc7bc10bda855adeb78",
     "bip66": "^1.1.0",
     "bitcoin-ops": "^1.3.0",
     "bs58check": "^2.0.0",
--- a/src/segwit_address.js
+++ b/src/segwit_address.js
@@ -1,0 +1,50 @@
+var bech32 = require('bech32')
+
+function decode (expectedPrefix, string) {
+  var result = bech32.decode(string)
+  if (result.prefix !== expectedPrefix) {
+    throw new Error('Unexpected prefix')
+  }
+
+  if (!(result.bitData.length > 0) && (result.bitData.length < 66)) {
+    throw new Error('Invalid length of encoded data')
+  }
+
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -52,6 +52,7 @@
   ],
   "dependencies": {
     "bigi": "^1.4.0",
+    "bech32": "git@github.com:bitcoinjs/bech32.git#81bc22761b3c8d286d6d8dc7bc10bda855adeb78",
```
kabutosamourai: Needs to be `https` rather than `git@` in order to check out for non-members of `bech32`

kabutosamourai:
This looks good as a first-cut transfer of the functionality from `bitcoinjs/bech32`.

TODO: tie into `src/index.js` as an include. We could all bike-shed about the API name exposed by the library (`segwit_address` vs `segwitAddress` etc).

Usability suggestion: perhaps the Bech32 human-readable prefix should be part of the `networks` parameters, so we can dispense with the `expectedPrefix` business. This would change the API to:

```
// Auto-detect network based on prefix. Throws if no network found
decode(string) -> { version, program, network }
- or -
// Throws if expected network does not match string prefix
decode(string, network) -> { version, program }

// Encodes program according to given network parameters
encode(version, program, network) -> string
```

afk11:
Adding `hrp` to network makes sense. It does raise questions about altcoins which may not have segwit.. but I guess we can test if it's set for now.

dcousens:
@kabutosamourai it should export under `.address` tbh.
`fromOutputScript` can handle it fine.

`.address` makes no guarantee about being `base58check` only.

kabutosamourai:
@dcousens That sounds great. Do you have any thoughts about adding the `hrp` to `networks`?

dcousens:
> Do you have any thoughts about adding the hrp to networks?

Of course!  Where else would it go? :+1:

dcousens:
@afk11 I might close this in favour of a `bitcoinjs/bitcoinjs-lib` branch,  if you do that in future I could easily just add to this :+1:

Issue #641, opened 2016-10-05
Uncompressed keys considered non-standard in witness scripts


dcousens:
Reference: https://bitcoincore.org/en/meetings/2016/09/29/

Relevant code: https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/src/script.js#L134-L147,  and subsequently all scripts that function use it would need to know they are in a `SegWit` context.

dcousens:
So, `function isPubKeyHashInput (script) {` would probably become `function isPubKeyHashInput (script, inWitness) {` or some such.

dcousens:
As mentioned in https://github.com/bitcoin/bips/pull/459/files#diff-d6a595a37a5d316054e79e53e9c2382fR128

afk11:
Not easy currently, but did something recently lead me to something that could help. Maybe adding some state to the script object: no state means it behaves like a scriptPubKey, and otherwise, it can be in a P2SH state, a P2WSH state, or a P2SH+P2WSH state.

Allowed state transitions:
```
none -> p2wsh
p2wsh -> p2sh
none -> p2sh
```

Forbidden:
```
p2sh -> p2wsh (p2sh happens after p2wsh, so this wouldn't work)
p2sh -> p2sh (nesting forbidden)
p2wsh -> p2wsh (nesting forbidden)
none -> p2wsh IF THE SCRIPT HAS UNCOMPRESSED KEYS
```

Mutating a script with `P2WSH()` would cause it to check the pubkeys and maybe other details (nesting). NB for later, going from P2WSH -> P2SH should mean the actual script is replaced with it's witness program.*

It might help us in other ways - you could expose `getScriptHash` and `getWitnessScriptHash` functions (opportunity to cache also). After making it P2SH, you'd replace getWitnessScriptHash with a function causing an error (cos P2SH scripts can't have a P2WSH output script)

This is only necessary because `script.scriptHash.output.encode()` etc take hashes (as they should) so there's no opportunity to check what we're hashing.. unless we offer an API that allows it :) 

When I was working on this I didn't really think about the uncompressed issue, though it seems to work ok.

What this would hopefully give us: people only use script output templates for the raw script, and decorate those scripts with contextual behavior. It's probably easier to use than calling the script templates by hand..

```js
var raw = script.multisig.output.encode(2, [])
var ws = p2wsh(raw)
var rs = p2sh(ws); // See note at *, replace witness script to witness program
var spk = rs.getOutputScript()
[...]

Issue #640, opened 2016-10-05
Bitcoin-core Policy following... policy


dcousens:
I think it is safe to say our "policy" for node policy related code is the following:
- ~~`bitcoinjs.script.*`~~ `bitcoinjs.template.*` functions should follow the script "templates" typically used in the community (and consequently, the bitcoin blockchain).
  This is a matter of community consensus,  but is probably going to always be closely related to what bitcoin core does by default.
  They are quite plainly just pattern matching functions, and should be treated as such.

`bitcoinjs.script.decompile` is consensus bound only,  it does not reject based on policy.
`bitcoinjs.script.compile` will adhere to bitcoin-core `IsStandard` policies rules such as minimalpush in https://github.com/bitcoinjs/bitcoinjs-lib/pull/638.

All ECDSA `sign` operations will adhere to `IsStandard` policies such as `LOW_S`, but `verify` should not reject them.

For rejecting non-standard `decoding`, you should use external modules to this library (which we could probably build a list of, and may even be in this organization).

**TLDR:** 
Where "standards compliant" refers to the default policies of bitcoin-core,  we adhere to the following:
- Any "creation" event must create standards-compliant data
- Any "validation" event must allow for non-standards compliant data (aka, consensus only)

For stricter validation,  use an external module which we [might have] provided.
We will use these modules to test `bitcoinjs-lib` against to ensure standards-compliant "encoding".

dcousens:
@junderw I think we can move this and the contributing issue to a CONTRIBUTING.md?

dcousens:
See https://github.com/bitcoinjs/bitcoinjs-lib/pull/1042

PR #857, opened 2017-08-22
Reject uncompressed keys for witness scripts

dcousens:
From https://github.com/bitcoinjs/bitcoinjs-lib/issues/641

- [x] Prevent `TransactionBuilder` creating these at all

Compliant with our policy in https://github.com/bitcoinjs/bitcoinjs-lib/issues/640

> `bitcoinjs.template.*` functions should follow the script "templates" typically used in the community (and consequently, the bitcoin blockchain).
This is a matter of community consensus, but is probably going to always be closely related to what bitcoin core does by default.
They are quite plainly just pattern matching functions, and should be treated as such.

`bitcoin.template.witnessPubKeyHash.decode` should fail if the pattern matched doesn't community policy [of rejected uncompressed pubkeys], these are templates afterall.
You can easily `bitcoin.script.decompile` yourself if you want to DIY.

As per [BIP143](https://github.com/bitcoin/bips/pull/459/files#diff-d6a595a37a5d316054e79e53e9c2382fR128)

### DIFF
--- a/src/templates/witnesspubkeyhash/input.js
+++ b/src/templates/witnesspubkeyhash/input.js
@@ -1,9 +1,44 @@
 // {signature} {pubKey}
 
-var pkh = require('../pubkeyhash/input')
+var bscript = require('../../script')
+var typeforce = require('typeforce')
+
+function isCompressedCanonicalPubKey (pubKey) {
+  return bscript.isCanonicalPubKey(pubKey) && pubKey.length === 33
+}
+
+function check (script) {
+  var chunks = bscript.decompile(script)
+
+  return chunks.length === 2 &&
+    bscript.isCanonicalSignature(chunks[0]) &&
+    isCompressedCanonicalPubKey(chunks[1])
+}
+check.toJSON = function () { return 'witnessPubKeyHash input' }
+
+function encodeStack (signature, pubKey) {
+  typeforce({
+    signature: bscript.isCanonicalSignature,
+    pubKey: isCompressedCanonicalPubKey
+  }, {
+    signature: signature,
+    pubKey: pubKey
+  })
+
+  return [signature, pubKey]
+}
+
+function decodeStack (stack) {
+  typeforce(check, stack)
+
+  return {
+    signature: stack[0],
+    pubKey: stack[1]
+  }
+}
 
 module.exports = {
-  check: pkh.check,
-  decodeStack: pkh.decodeStack,
-  encodeStack: pkh.encodeStack
+  check: check,
+  decodeStack: decodeStack,
+  encodeStack: encodeStack
 }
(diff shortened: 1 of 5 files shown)


src/templates/witnesspubkeyhash/input.js
```diff
@@ -1,9 +1,44 @@
...
+  }, {
+    signature: signature,
+    pubKey: pubKey
+  })
+
+  return [signature, pubKey]
+}
+
+function decodeStack (stack) {
+  typeforce(check, stack, false)
```
dcousens: Liberal in what we accept

src/templates/witnesspubkeyhash/input.js
```diff
@@ -1,9 +1,44 @@
...
+  return chunks.length === 2 &&
+    bscript.isCanonicalSignature(chunks[0]) &&
+    (strict ? isCompressedCanonicalPubKey(chunks[1]) : bscript.isCanonicalPubKey(chunks[1]))
+}
+check.toJSON = function () { return 'witnessPubKeyHash input' }
+
+function encodeStack (signature, pubKey) {
+  typeforce({
+    signature: bscript.isCanonicalSignature,
+    pubKey: isCompressedCanonicalPubKey
```
dcousens: Strict in what produce

src/templates/witnesspubkeyhash/input.js
```diff
@@ -1,9 +1,44 @@
...
 
-var pkh = require('../pubkeyhash/input')
+var bscript = require('../../script')
+var typeforce = require('typeforce')
+
+function isCompressedCanonicalPubKey (pubKey) {
+  return bscript.isCanonicalPubKey(pubKey) && pubKey.length === 33
+}
+
+function check (script, strict) {
```
dcousens: Issue here,  `check` is usually `check(script, allowIncomplete)`,  not `check(script, strict)`... I don't think this is OK?

src/transaction_builder.js
```diff
@@ -676,6 +676,12 @@ TransactionBuilder.prototype.sign = function (vin, keyPair, redeemScript, hashTy
   var signed = input.pubKeys.some(function (pubKey, i) {
     if (!kpPubKey.equals(pubKey)) return false
     if (input.signatures[i]) throw new Error('Signature already exists')
+    if (!keyPair.compressed && (
+      input.prevOutType === scriptTypes.P2WPKH ||
+      input.prevOutType === scriptTypes.P2WSH ||
+      input.redeemScriptType === scriptTypes.P2WPKH ||
+      input.redeemScriptType === scriptTypes.P2WSH
+    )) throw new Error('BIP143 rejects uncompressed public keys in P2WPKH or P2WSH')
```
dcousens: @afk11 not an amazing place to put this,  but I intend to refactor this area hopefully soon

afk11:
Looks good! 

It'd be cool if there an approach to making scripts script that was easier than calling `script.X.encode(the hashXXX())` primitives directly.. 

maybe a factory that takes a script, the user calls p2sh(), and/or p2wsh(), before a final method is called that looks at what was done and applies contextual checks? (like conditionally forbidding uncompressed public keys)

dcousens:
> script.X.encode(the hashXXX())

I concur,  this is lame.
But,  if with powerful primitives,  then we can build powerful convenience wrappers :+1: .

We should always be open to ideas as to how we can improve UX,  with the fallback of a consistent, strict and safe API :smiley: .

PR #859, opened 2017-08-22
Fix nullData output encode and decode

### LINKED ISSUE #858: OP_RETURN scriptPubkey can be up to 80 bytes
https://github.com/bitcoinjs/bitcoinjs-lib/blob/b552c63fb0a76e82d55a8d562cf6df76a25f3fc5/src/templates/nulldata.js#L25

This assumes OP_RETURN then PUSHDATA (1 Byte) and slices 2 bytes off the head.

If you push 77 bytes, the PUSHDATA is 2 bytes long, so decode will give the second byte of the PUSHDATA on the head of the decoded buffer.

Since OP_RETURN allows for multiple pushes as long as the scriptPubkey is 80 bytes or lower, the decode should be

1. slice 1 byte
2. read PUSHDATA and slice it
3. slice the number of bytes in 2
4. If ended, return Buffer.
5. If not ended read PUSHDATA again
6. slice off the number of bytes in 5
7. convert the return value into an array of buffers and push the result of 6
8. repeat 5-7 until end of buffer reached.

junderw:
Fixes #858

### DIFF
--- a/src/templates/nulldata.js
+++ b/src/templates/nulldata.js
@@ -14,15 +14,23 @@ function check (script) {
 check.toJSON = function () { return 'null data output' }
 
 function encode (data) {
-  typeforce(types.Buffer, data)
+  typeforce(typeforce.oneOf(types.Buffer, types.Array), data)
 
-  return bscript.compile([OPS.OP_RETURN, data])
+  if (!types.Array(data)) {
+    data = [ data ]
+  }
+
+  return bscript.compile([OPS.OP_RETURN].concat(data))
 }
 
 function decode (buffer) {
   typeforce(check, buffer)
 
-  return buffer.slice(2)
+  var chunks = bscript.decompile(buffer)
+
+  chunks.shift()
+
+  return chunks.length === 1 ? chunks[0] : chunks
 }
 
 module.exports = {
--- a/test/fixtures/script.json
+++ b/test/fixtures/script.json
@@ -96,6 +96,10 @@
       "script": "6a14deadffffffffffffffffffffffffffffffffbeef"
     },
     {
+      "asm": "OP_RETURN deadffffffffffffffffffffffffffffffffbeef deadffffffffffffffffffffffffffffffffbeef",
+      "script": "6a14deadffffffffffffffffffffffffffffffffbeef14deadffffffffffffffffffffffffffffffffbeef"
+    },
+    {
       "asm": "OP_0 OP_0 3044022001ab168e80b863fdec694350b587339bb72a37108ac3c989849251444d13ebba02201811272023e3c1038478eb972a82d3ad431bfc2408e88e4da990f1a7ecbb263901 3045022100aaeb7204c17eee2f2c4ff1c9f8b39b79e75e7fbf33e92cc67ac51be8f15b75f90220659eee314a4943a6384d2b154fa5821ef7a084814d7ee2c6f9f7f0ffb53be34b01",
       "script": "0000473044022001ab168e80b863fdec694350b587339bb72a37108ac3c989849251444d13ebba02201811272023e3c1038478eb972a82d3ad431bfc2408e88e4da990f1a7ecbb263901483045022100aaeb7204c17eee2f2c4ff1c9f8b39b79e75e7fbf33e92cc67ac51be8f15b75f90220659eee314a4943a6384d2b154fa5821ef7a084814d7ee2c6f9f7f0ffb53be34b01",
       "stack": [


src/templates/nulldata.js
```diff
@@ -14,15 +14,23 @@ function check (script) {
 check.toJSON = function () { return 'null data output' }
 
 function encode (data) {
-  typeforce(types.Buffer, data)
+  typeforce(typeforce.oneOf(types.Buffer, types.Array), data)
 
-  return bscript.compile([OPS.OP_RETURN, data])
+  if (!types.Array(data)) {
+    data = [ data ]
```
dcousens: `concat` should work either way

src/templates/nulldata.js
```diff
@@ -14,15 +14,23 @@ function check (script) {
...
+  return bscript.compile([OPS.OP_RETURN].concat(data))
 }
 
 function decode (buffer) {
   typeforce(check, buffer)
 
-  return buffer.slice(2)
+  var chunks = bscript.decompile(buffer)
+
+  chunks.shift()
```
dcousens: Could technically modify `buffer` if `buffer` was an `Array`...

src/templates/nulldata.js
```diff
@@ -14,15 +14,23 @@ function check (script) {
...
+
+  return chunks.length === 1 ? chunks[0] : chunks
```
dcousens: Although it is semi backwards compatible,  it might break existing assumptions that users have about what is returned (`Buffer`).
I don't know if there is going to be a better answer though...

test/fixtures/script.json
```diff
@@ -96,6 +96,10 @@
       "script": "6a14deadffffffffffffffffffffffffffffffffbeef"
     },
     {
+      "asm": "OP_RETURN deadffffffffffffffffffffffffffffffffbeef deadffffffffffffffffffffffffffffffffbeef",
+      "script": "6a14deadffffffffffffffffffffffffffffffffbeef14deadffffffffffffffffffffffffffffffffbeef"
+    },
```
dcousens: Nice

src/templates/nulldata.js
```diff
@@ -14,15 +14,23 @@ function check (script) {
 check.toJSON = function () { return 'null data output' }
```
dcousens: Hmmm, I wonder if `check` should enforce <80...

dabura667: That's a standard rule... I don't see any other standardness rules being enforced. (Unless I'm missing something)

src/templates/nulldata.js
dabura667: Hmmmmmmm... 99.99999999% of OPReturns are one push......

It's their fault for not testing edge cases, ;-P

src/templates/nulldata.js
dcousens: See https://github.com/bitcoinjs/bitcoinjs-lib/issues/640,  these templates are "standard"ness

src/templates/nulldata.js
afk11: Should we assert that the first opcode == OP_RETURN here?

src/templates/nulldata.js
dabura667: I guess it should check if it's push only as well.

src/templates/nulldata.js
dcousens: Done in `check`

src/templates/nulldata.js
afk11: I suggest this might worth clarifying if we're thinking about a break. Although the majority of `OP_RETURN` usage is just one push, they can also have zero or several data-pushes, it might be good to return an array here. The content after OP_RETURN hardly gets executed, does it even have to be valid script?

Whatever templates we have, it's clear that people expect them not to conditionally fail (suppose the function couldn't recognize `OP_RETURN [DATA] [DATA]` and classified it non-standard) and that whatever we call 'nullData' doesn't change in the future..

I noticed we don't check the first opcode == OP_RETURN - this is another degree of freedom to consider when we call this script 'nullData'..

Maybe we should outline what exactly we consider a nullData script.. Thoughts?

dcousens: @afk11 remember `decode` has an arbitrary return type,  it merely needs to be able to unpack its outputs back to `encode` to be "complete".
```
encode(...) -> Buffer
decode(Buffer) -> ...
```

src/templates/nulldata.js
dabura667: Really? How so?

I thought Array.prototype.shift just pops the first item of an Array, and wouldn't modify the Buffers contained in the Array?

dcousens: https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/src/script.js#L86

afk11: Oops, you are correct

dabura667: Why would anyone insert chunks into a decode function?

I thought the whole point was that encode always outputs a buffer and decode requires a buffer as input...?

dcousens: @dabura667 it is an unfortunate optimization that has lead to way too much confusion.

dcousens:
@junderw we should accept a `[Buffer]` to `encode`,  and `decode` should return `[Buffer]`... it is the only real solution.
But, it will be `4.0.0`.
For now,  we simply can't really do much here.

dcousens:
It is atleast consistent that `encode` only encodes 1 push item.

afk11:
Concept ACK, but since in this case the solution doesn't retain BC and the classification of 'nulldata' would now be slightly different, we should hold this off for v4.0.0 and make a point out of mentioning it in the release notes so people can update accordingly

dcousens:
See https://github.com/bitcoinjs/bitcoinjs-lib/pull/1041

PR #860, opened 2017-08-22
Formatting improvement

benhc123:
license


dcousens:
@benhc123 interesting,  I wonder if this could be related to the difference between

``` js
if (license) MIT()
```

compared to

``` js
if (license) {
  MIT()
}

PR #863, opened 2017-08-23
Stack template fixes

dcousens:
The `OP_0` / `Buffer< >` interchange is a nightmare.

I don't know if there is a solution other than minimizing interchange between Stack/Chunks/Script types,  because the distinction is immediately lost without classification.

As for decoding from a stack... we can't make assumptions about whether it is OP_INT or `Buffer`.


src/script.js
```diff
@@ -127,7 +133,10 @@ function toASM (chunks) {
 
   return chunks.map(function (chunk) {
     // data?
-    if (Buffer.isBuffer(chunk)) return chunk.toString('hex')
+    if (Buffer.isBuffer(chunk)) {
+      if (chunk.length === 0) return 'OP_0'
```
dcousens: Gross

dcousens:
This isn't turning out as elegant as I think we hoped.
The intrinsic types at play are: 

``` haskell
OP_INT = Number
OP = Number
Chunks = [OP_INT | OP | Buffer]
Script = Buffer
Stack = [Buffer]

encode :: ... _> Script
encodeRaw :: ... -> Chunks
encodeStack :: ... -> Stack

decode :: Script -> ...
decodeStack :: Stack -> ...
```

dcousens:
The issue was `decodeStack` was returning `chunks`

dcousens:
Hopefully we can use `asMinimalOP` to avoid this issue

dcousens:
I think maybe `decompile` should decompile minimally?  Maybe?  Thoughts?
We need to be strict about this.

afk11:
> Should script.decompile return Buffer or OP_ if it is ambiguous
It should return OP_... and leave interpreting the opcodes up to the caller. The more information they have, the better :)

afk11:
weak nack, I would avoid using opcodes as an intermediate (as encodeRaw seems to do) It would mean doing this: chunks -> stack -> scriptSig which doesn't make a good constraint, since not all chunks have a stack? (scriptPubKeys cannot have a stack, they operate on a stack)

Whereas the following are invariant when SIGPUSHONLY and MINIMALDATA are in effect:

stack -> chunks -> scriptSig
stack -> witness

dcousens:
>  I would avoid using opcodes as an intermediate

The opcodes are used intermediately as they are how those script templates are typically defined.  Not as a stack?
For a witness stack,  they are encoded to be a push-only vector of `Buffer`s (`[Buffer]`). 

> It would mean doing this: chunks -> stack -> scriptSig

Wha?
We have two typical flows...
``` haskell
Chunks -> Stack             -- a witness stack
Chunks -> Script             -- scriptSig, scriptPubKey
```

`encodeRaw` outputs `Chunk`s,  not a stack,  nor a script

dcousens:
As mentioned on slack

* `encodeStack` to be deprecated, renamed as `encodeRaw`.
* `toStack` to be deprecated, renamed as `toWitness`.
* `encodeWitness` to be added.

dcousens:
This could be insanely simplified by removing the encoding to `Script` or `Witness`,  and instead leave that to the user as an option.

``` js
bitcoin.script.p2pk.encode(pubKey) // existing
bitcoin.script.toScript(bitcoin.script.p2pk.encode(pubKey)) // replacement? booooo

// other ideas
bitcoin.p2pkh.fromKeyPair(kp).address
bitcoin.p2pkh.fromKeyPair(kp).script
bitcoin.p2pkh.fromAddress(address).script
bitcoin.p2pkh.fromOutputScript(script).address

// but then how redeem... ?
bitcoin.p2pkh.redeem(kp, signature).script
bitcoin.p2pkh.redeem(kp, signature).witness
```

dcousens:
I think `templates.witnessPubKeyHash` and `templates.witnessScriptHash` may have been the wrong approach.

We may have needed instead a plain `witnessOutput` instead,  not a whole extra category.
For example,  as of now,  you can have an error thrown for a uncompressed pubkey in `witnessPubKeyHash`,  but not in a `pubKeyHash` witness script!

dcousens:
Too many bad decisions, and while this PR would make things stricter,  it doesn't help the clumsiness of the API.

I think it will be better to simply move faster towards `4.0.0`.

PR #867, opened 2017-08-24
script: use asMinimalOP for ASM/decompile

dcousens:
Everything is now consistent in `compile/decompile`.


dcousens:
Merging as we generally each agreed to this,  and this is from https://github.com/bitcoinjs/bitcoinjs-lib/pull/863 which had had, _that_ part reviewed.

Any ACKs still appreciated however.

PR #871, opened 2017-08-29
Add witness is true to signing

junderw:
This was causing problems with multisig signing in rare cases... (why not all cases... not sure)

`var witness = false` is declared near the top but it is never used... so I figured it should be in master.


dabura667:
I need to add a test case... I'll try to get to it today. :-/

junderw:
OK, added the test case. This should be good to merge now.

dcousens:
> This was causing problems with multisig signing in rare cases... (why not all cases... not sure)

Ready to merge,  but will investigate what you mean first,  uncertainty is never a great sign.

dabura667:
The test case I included fails on master, just FYI.

PR #875, opened 2017-08-31
Fix txb.__overMaximumFees for segwit

junderw:


dcousens:
We need a test to verify this :+1:

junderw:
Actually, this doesn't fix the problem at all... this maximumFee check is broken when TransactionBuilder is imported from Transaction.

ex.

```javascript
var tx = bitcoin.Transaction.fromHex('01000000000101035811767cb7f7050f7ecad148ccff6f20a8deb3b05fbf8126fdcd38444f919e0000000000ffffffff02a7290100000000001600148d7a0a3461e3891723e5fdf8129caa0075060cff01000000000000001600148d7a0a3461e3891723e5fdf8129caa0075060cff0247304402203aa3f264b05286cd3caffafc2697dea3ef29b528ed5e5a9e0fd3741606b8fbf102201a60f641206e88d30a595ec9a3a5b97b0076d7a6ebd81e8612a279cd209eca6801210283409659355b6d1cc3c32decd5d561abaac86c37a353b52895a5e6c196d6f44800000000')
var txb = bitcoin.TransactionBuilder.fromTransaction(tx)
txb.inputs[0].value = 189200
txb.__overMaximumFees(txb.tx.virtualSize())
txb.__overMaximumFees(txb.tx.byteLength())
txb.tx.virtualSize()
// should be 141 but gives us 113
txb.tx.byteLength()
// should be 222 but gives us 113
```

The reason why is because when a TransactionBuilder imports from Transaction, the txb.tx Transaction object loses its witnesses, so txb.tx.hasWitnesses() returns false even though txb has witnesses.

This problem is a little more hairy than I thought... and this bug actually might push people over 1000 satoshi/byte if they make a large segwit tx and aim for a 500 satoshi-byte rate or something (since the broken Transactions byte count will be less than half the size, it would double the rate for this one function only, and cause an error.

dcousens:
@junderw could you open an issue for that?

dcousens:
@junderw are you sure?
The Transaction has been re-built at this point,  it should have any witnesses it will ever have.

junderw:
Try it for yourself.

Run:

```javascript
var tx = bitcoin.Transaction.fromHex('01000000000101035811767cb7f7050f7ecad148ccff6f20a8deb3b05fbf8126fdcd38444f919e0000000000ffffffff02a7290100000000001600148d7a0a3461e3891723e5fdf8129caa0075060cff01000000000000001600148d7a0a3461e3891723e5fdf8129caa0075060cff0247304402203aa3f264b05286cd3caffafc2697dea3ef29b528ed5e5a9e0fd3741606b8fbf102201a60f641206e88d30a595ec9a3a5b97b0076d7a6ebd81e8612a279cd209eca6801210283409659355b6d1cc3c32decd5d561abaac86c37a353b52895a5e6c196d6f44800000000')
var txb = bitcoin.TransactionBuilder.fromTransaction(tx)
```

Then inspect txb.tx it has no witnesses

junderw:
Perhaps TransactionBuilder only breaks witnesst on P2WPKH (no P2SH) inputs.

The hex example I gave above is 1 P2WPKH input.

dcousens:
> Then inspect txb.tx it has no witnesses

That is expected.
`.tx` is a merely an internal placeholder structure.

dcousens:
We should rename `.tx` to `.__tx`.

junderw:
Here's the current __build where the problem __overMaximumFees is being called.

I guess from what you're saying, the problem lies with `var tx = this.tx.clone()` meaning we shouldn't be doing this?

if this.tx doesn't have a witness (even though it should) then this.tx.clone() also does not have a witness

The below still gives incorrect byte counts and the tx2.hasWitnesses() is still false.

```javascript
var tx = bitcoin.Transaction.fromHex('01000000000101035811767cb7f7050f7ecad148ccff6f20a8deb3b05fbf8126fdcd38444f919e0000000000ffffffff02a7290100000000001600148d7a0a3461e3891723e5fdf8129caa0075060cff01000000000000001600148d7a0a3461e3891723e5fdf8129caa0075060cff0247304402203aa3f264b05286cd3caffafc2697dea3ef29b528ed5e5a9e0fd3741606b8fbf102201a60f641206e88d30a595ec9a3a5b97b0076d7a6ebd81e8612a279cd209eca6801210283409659355b6d1cc3c32decd5d561abaac86c37a353b52895a5e6c196d6f44800000000')
var txb = bitcoin.TransactionBuilder.fromTransaction(tx)
txb.inputs[0].value = 189200
var tx2 = txb.tx.clone()
txb.__overMaximumFees(tx2.virtualSize())
txb.__overMaximumFees(tx2.byteLength())
tx2.virtualSize()
// should be 141 but gives us 113
tx2.byteLength()
// should be 222 but gives us 113
```

dcousens:
@junderw `.clone` does clone witness data,   if it was present.

`txb.tx` does not have witness data **intentionally**,  as discussed on slack.
`TransactionBuilder.prototype.build` uses `(this.__overMaximumFees(tx.byteLength())) `,  where `tx` *has had* `setInputScript` and `setWitness` called,  and therefore **does** have witness data.

dcousens:
I think a safe assumption should be:

``` javascript
var txb = bitcoin.TransactionBuilder.fromTransaction(tx)

assert.strictEqual(txb.build().virtualSize(), tx.virtualSize())
```

If that assumption does not hold,  then yes,  something probably needs to be fixed.

dcousens:
@junderw could you amend the test fixtures and rebase this PR?
It it is a bug that can be fixed.

junderw:
@dcousens Added test case and fixed a fixture that failed due to this PR. (it had high fees, so I increased the output amounts.)

junderw:
I also confirmed this test case fails on master.

junderw:
@dcousens Added test case and fixed a fixture that failed due to this PR. (it had high fees, so I increased the output amounts.)

junderw:
I also confirmed this test case fails on master.

Issue #872, opened 2017-08-30
Link error in section example  of ReadMe 


shareworlds:
Click these links in ReadMe, it shows not found error.

Generate a SegWit address
Generate a SegWit P2SH address
Generate a SegWit 3-of-4 multisig address
Generate a SegWit 2-of-2 P2SH multisig address

PR #877, opened 2017-08-31
Fixed Segwit links

wmpedersen:
The links were redirecting to 404 - I've modified them so they adhere to the beginning of the respective Segwit unit tests.This pull is in reference to Issue #872


wmpedersen:
Perhaps this is being dealt with already, I see an issue #872 which addresses this.

wmpedersen:
I found some more links in the Readme that were moved by the addition of the Segwit unit tests. ff61de9 covers everything in the file so far.

afk11:
Almost worth having a script to generate it at release time, and a travis test to check.. It's unfortunate how prone to rot documentation like it can be, especially when line numbers are involved :P

dabura667:
IMO links should always point to a commit based tree of the file. Will never budge. :-)

dcousens:
@dabura667 I did that,  it didn't help the code rot situation.

Dead-links are at least obvious to users.

> Almost worth having a script to generate it at release time

These links were actually generated... it appears my regex foo failed though.

PR #882, opened 2017-09-06
TransactionBuilder: collect witnessValue as input.value, and match it

dcousens:


dcousens:
@afk11 to be clear,  this means if `input.value` is defined, it **must** match `witnessValue`.
If `witnessValue` is *not* provided,  it still throws.

No change in behaviour there.

But that could be up for discussion, but I tend to agree with you that the signer **should** provide it irrespective of whether it is known already.

test/fixtures/transaction_builder.json
```diff
@@ -428,7 +428,7 @@
 
       {
         "description": "Sighash: SINGLE (random)",
-        "txHex": "01000000012ffb29d53528ad30c37c267fbbeda3c6fce08f5f6f5d3b1eab22193599a3612a010000006b483045022100dad4bd28448e626ecb1ade42a09c43559d50b61b57a06fac992a5ecdd73deb740220524082f83560e2df9afaa283c699dec4c5b01687484d73e7b280e5a506caf1c4032102f1c7eac9200f8dee7e34e59318ff2076c8b3e3ac7f43121e57569a1aec1803d4ffffffff02c0e1e400000000001976a9140de1f9b92d2ab6d8ead83f9a0ff5cf518dcb03b888ac80969800000000001976a91454d0e925d5ee0ee26768a237067dee793d01a70688ac00000000",
+        "txHex": "01000000012ffb29d53528ad30c37c267fbbeda3c6fce08f5f6f5d3b1eab22193599a3612a010000006b483045022100f963f1d9564075a934d7c3cfa333bd1378859b84cba947e149926fc9ec89b5ae02202b5b912e507bae65002aff972f9752e2aeb2e22c5fdbaaad672090378184df37032102f1c7eac9200f8dee7e34e59318ff2076c8b3e3ac7f43121e57569a1aec1803d4ffffffff0260a62f01000000001976a9140de1f9b92d2ab6d8ead83f9a0ff5cf518dcb03b888ac80969800000000001976a91454d0e925d5ee0ee26768a237067dee793d01a70688ac00000000",
         "inputs": [
```
dcousens: these fixtures had to be updated,  as they each had absurd fees (heh)

afk11:
Ah ok, that seems fine!

PR #884, opened 2017-09-06
Fix buildStack & multisig.input.encodeStack

afk11:
multisig.input.encode should convert OP_0's to empty buffers to deal with padding requirement from transaction_builder (otherwise need to duplicate code for empty sigs)


dcousens:
(re-open if I mis-understood the approach we were going to take)

PR #893, opened 2017-09-15
Add alt coins BIP44

gilemon:
Add alt coins BIP44 https://github.com/libbitcoin/libbitcoin/wiki/Altcoin-Version-Mappings


dcousens:
@gilemon hi!
We decided to remove these mappings some time ago due to the "top 10" changing ... well, near daily.

We have only kept Litecoin as it hasn't left that "top 3" position (as a bitcoinjs-lib compatible alt) in years - but we have nearly removed it several times.

gilemon:
Hi, OK no worries - I needed these coins for a project so I thought it could be useful.

PR #897, opened 2017-09-20
fix input classification bug

arik-so:
In some rare cases, transactions input types will get classified incorrectly. Script hash inputs will be classified as nonstandard. This change fixes that.


dcousens:
Can you please add test fixtures?

afk11:
I'd love to see a snippet of how this came about. TransactionBuilder.fromTransaction()? or just by calling the methods yourself?

Whether it's as simple as this fix, or if we have other input handling problems would be useful to know!

arik-so:
Hi @dcousens and @afk11, I updated the pull request to add a transaction that has several examples of where the input classification would previously have been broken, and one input that remains unaffected. I hope the way I built the unit test is ok.

dcousens:
Nice spot @arik-so!  I think the problem is rooter deeper than that though... check https://github.com/bitcoinjs/bitcoinjs-lib/pull/898

arik-so:
I agree, @dcousens, the classification should be order-agnostic. If the other PR works, perhaps simply keeping the unit test changes from this one would make sense. And @junderw, I added the spaces, as requested.

dcousens:
@arik-so if you could add your unit test changes by submitting a PR to the `bitcoinjs:fixes` branch,  that'd be awesome :+1:

arik-so:
Sure thing! https://github.com/bitcoinjs/bitcoinjs-lib/pull/900

dcousens:
closing in favour of #898

PR #898, opened 2017-09-22
scriptHash.check should be agnostic of classification order

dcousens:
Alternative to https://github.com/bitcoinjs/bitcoinjs-lib/pull/897

Thanks and credit to @arik-so for reporting the bug,  and providing a test case!

The issue is that classifications aren't unique and aren't guaranteed to be un-ambiguous.
`scriptHash.check` has assumed that through using `classify*` internally.

Instead,  lets match 1 for 1,  and if there is a double match,  then we can say OK,  it is probably script hash.


src/templates/scripthash/input.js
```diff
@@ -20,12 +20,23 @@ function check (script, allowIncomplete) {
...
+
+  // match types
+  if (bscript.pubKeyHash.input.check(scriptSigChunks) &&
+    bscript.pubKeyHash.output.check(redeemScriptChunks)) return true
+
+  if (bscript.multisig.input.check(scriptSigChunks, allowIncomplete) &&
+    bscript.multisig.output.check(redeemScriptChunks)) return true
+
+  if (bscript.pubKey.input.check(scriptSigChunks) &&
+    bscript.pubKey.output.check(redeemScriptChunks)) return true
```
dcousens: ironically,  this also stops recursive scriptHash

dcousens:
tests pass, @junderw / @afk11

junderw:
LGTM

Issue #902, opened 2017-09-23
Issue when using browserify instructions on bitcoinjs.org


mandelmonkey:
I used the compile for browser instructions on bitcoinjs.org

and the follow error was shown

`as-MacBook-Pro:test a$ sudo npm install -g browserify uglify-js
/usr/local/bin/uglifyjs -> /usr/local/lib/node_modules/uglify-js/bin/uglifyjs
/usr/local/bin/browserify -> /usr/local/lib/node_modules/browserify/bin/cmd.js
+ uglify-js@3.1.1
+ browserify@14.4.0
added 135 packages in 8.086s
as-MacBook-Pro:test a$ browserify -r bitcoinjs-lib -s bitcoin | uglifyjs > bitcoinjs.min.js
Parse error at 0:96,4
let ALPHABET = 'qpzry9x8gf2tvdw0s3jn54khce6mua7l'
    ^
ERROR: Unexpected token: name (ALPHABET)
    at JS_Parse_Error.get (eval at <anonymous> (/usr/local/lib/node_modules/uglify-js/tools/node.js:21:1), <anonymous>:79:23)
    at fatal (/usr/local/lib/node_modules/uglify-js/bin/uglifyjs:268:52)
    at run (/usr/local/lib/node_modules/uglify-js/bin/uglifyjs:225:9)
    at Socket.<anonymous> (/usr/local/lib/node_modules/uglify-js/bin/uglifyjs:163:9)
    at emitNone (events.js:110:20)
    at Socket.emit (events.js:207:7)
    at endReadableNT (_stream_readable.js:1047:12)
    at _combinedTickCallback (internal/process/next_tick.js:102:11)
    at process._tickCallback (internal/process/next_tick.js:161:9)`

junderw:
Hey Christian.

What version of node are you using? Try switching nvm to use v7

mandelmonkey:
seems my handle gave me away
am on v5.0.3 so will upgrade and give it another go thanks!

dcousens:
@mandelmonkey it appears that `uglify-js` doesn't support ES6 `let` statements.
In repository,  we track LTS node versions for tracking backwards compatibility.

I'd recommend using `--transform [ babelify ]` on your browserify build step to downgrade to ES5.

I think this should be something we outline in the README,  with steps.
Thoughts @bitcoinjs/committers ?

junderw:
I agree. Let’s. ;-)

[...]

PR #903, opened 2017-09-24
README: add notes about ES5, Node LTS feature tracking

dcousens:
Following https://github.com/bitcoinjs/bitcoinjs-lib/issues/902,  I think it is good to commit to what language features we track in `bitcoinjs-lib`.

To date,  we have tracked the Node LTS version,  and whatever it supports,  but we haven't officially declared that until now.


junderw:
Maintaining backwards compatibility is fine to a point.

Suggesting babelify is the right move imo

PR #904, opened 2017-09-25
package: rm contributors field, outdated, update wallet estimate

dcousens:
The estimate comes from the number of wallet users reported by https://blockchain.info,  but to be honest,  but that is ignoring every other wallet statistic... alas


dcousens:
asssumed non controversial

PR #909, opened 2017-09-26
transaction_builder fromTransaction

afk11:
should attempt to take input.value from transaction,
since otherwise fixMultisigOrder will not be able
to succeed.

### DIFF
--- a/src/transaction_builder.js
+++ b/src/transaction_builder.js
@@ -174,7 +174,7 @@ function expandInput (scriptSig, witnessStack) {
 
 // could be done in expandInput, but requires the original Transaction for hashForSignature
 function fixMultisigOrder (input, transaction, vin) {
-  if (input.redeemScriptType !== scriptTypes.MULTISIG || !input.redeemScript) return
+  if (input.signType !== scriptTypes.MULTISIG || !input.signScript) return
   if (input.pubKeys.length === input.signatures.length) return
 
   var unmatched = input.signatures.concat()
@@ -190,7 +190,12 @@ function fixMultisigOrder (input, transaction, vin) {
 
       // TODO: avoid O(n) hashForSignature
       var parsed = ECSignature.parseScriptSignature(signature)
-      var hash = transaction.hashForSignature(vin, input.redeemScript, parsed.hashType)
+      var hash
+      if (input.witness) {
+        hash = transaction.hashForWitnessV0(vin, input.signScript, input.value, parsed.hashType)
+      } else {
+        hash = transaction.hashForSignature(vin, input.signScript, parsed.hashType)
+      }
 
       // skip if signature does not match pubKey
       if (!keyPair.verify(hash, parsed.signature)) return false
@@ -517,6 +522,7 @@ TransactionBuilder.fromTransaction = function (transaction, network) {
     txb.__addInputUnsafe(txIn.hash, txIn.index, {
       sequence: txIn.sequence,
       script: txIn.script,
+      value: txIn.value,
       witness: txIn.witness
     })
   })


dcousens:
@afk11 how would it take it from the Transaction?

src/transaction_builder.js
```diff
@@ -517,6 +522,7 @@ TransactionBuilder.fromTransaction = function (transaction, network) {
     txb.__addInputUnsafe(txIn.hash, txIn.index, {
       sequence: txIn.sequence,
       script: txIn.script,
+      value: txIn.value,
```
junderw: Transaction doesn't currently have value in the input... as it is not encoded in the raw transaction that we output........

Like @dcousens said, we can't fix Multisig order if input is witness... since we need a value.

junderw:
Two ways to fix this once and for all:

1. change the serialization for incomplete multisig segwit inputs to encode value in there somewhere.
2. wait for the BIP proposal for 175 to be set in stone and use that for buildIncomplete serialization.

2 might be good... hmm...

afk11:
well my very much ad-hoc solution to get signature extracting working was to set the value on each input before doing `TransactionBuilder.fromTransaction` and that works a treat with this code, though I realize it's never been set on a transaction's inputs before. But how else is fixMultisigOrder expected to work now that segwit captures the value?

I would consider this part of the "we don't have a txout when we try to sign the input" issue. Generally, I'd like to avoid doing _anything_ input-signature related unless we have the txOut and the value (scriptPubKey would have been fine on it's own until segwit)

```js
var txb = TransactionBuilder.fromTransaction(tx)
var input = txb.input(0, txOut, [rs], [ws]); // classification, extraction, fixMultisigInput can happen now
input.sign(key, [hashType])
```

afk11:
@junderw the problem is that not all implementations will pad with OP_0's, I think yet another custom trick would make this problem worse. 

Re BIP174, at the moment it's impossible to deliver the data even when required, I think the library should be able to facilitate this use (only do on-demand classification/extraction of inputs) without the need for external specs for representing said data

src/transaction_builder.js
afk11: But that's all it needs, and __addInputUnsafe already tolerates `value` being provided in the `opts`. It's so close to making this work, because `witness` tells us the `sigVersion` already. 

I'm not proposing everyone should expect the field to be there for anything else, it at least makes it work for all multisigs, not just P2SH ones (it doesn't work with bare either, I guess?)

dcousens:
@afk11 my point was,  is `fixMSOrder` even required if we have a higher level format?

afk11:
@dcousens Imagine you want to produce the standardized serialized data encoding _for_ a transaction without one. why wouldn't fixMSOrder be needed in this case? :)

dcousens:
I think BIP174 is a better approach,  we should probably start blocking partial transaction encoding as a plausible option.
It is the worst.

afk11:
I'm not sure how an encoding standard such as that helps this - it's an API problem if you can't do fixMultisigInput today.

dcousens:
Does this need to be merged?  Or do we close?

afk11:
Still unsure how best to proceed. It's not urgent (== I've merged it in my branch) - but lacking a nicer way to do it I think this solution can be held back.

dcousens:
Can you add a test case for this @afk11 ?

src/transaction_builder.js
dcousens: @afk11 I think why this confused me for so long is I didn't understand that you meant you added `txIn.value` to the `txIn` in the _transaction_ that you _pass_ to `fromTransaction`.

It isn't a natural usage of the API,  but I can understand your motivation given you want everything prepared before you do `.sign`.

dcousens:
@afk11 4aa741e required a revert to pass tests.
This feature is quite contrived to your particular case where you want to pre-prepare the `input.value`.
Fair enough,  but,  is it worth it?

afk11:
Not sure if contrived is the right word to use. Of course it's natural to sign transactions produced by other libraries. 

I'm pretty sure the detail you reverted is still something we need. fixMSOrder shouldn't deal in terms of redeemScripts, we're talking about segwit.

src/transaction_builder.js
```diff
@@ -192,10 +192,10 @@ function fixMultisigOrder (input, transaction, vin) {
       // TODO: avoid O(n) hashForSignature
       var parsed = ECSignature.parseScriptSignature(signature)
       var hash
-      if (input.witness) {
-        hash = transaction.hashForWitnessV0(vin, input.signScript, input.value, parsed.hashType)
+      if (input.witness && input.value !== undefined) {
+        hash = transaction.hashForWitnessV0(vin, input.redeemScript, input.value, parsed.hashType)
```
afk11: should be witnessScript

src/transaction_builder.js
```diff
@@ -192,10 +192,10 @@ function fixMultisigOrder (input, transaction, vin) {
       // TODO: avoid O(n) hashForSignature
       var parsed = ECSignature.parseScriptSignature(signature)
       var hash
-      if (input.witness) {
-        hash = transaction.hashForWitnessV0(vin, input.signScript, input.value, parsed.hashType)
+      if (input.witness && input.value !== undefined) {
```
afk11: this isn't a reversion, this will skip signatures if the input.value isn't set.

afk11:
I should clarify, I don't think this is an acceptable solution - it's not right to abuse the structures in this way - but it did the trick, and retains BC :) my ideal solution is require the full txout at signing time, so weird shortcuts aren't necessary anymore https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/src/transaction_builder.js#L351

dcousens:
> I should clarify, I don't think this is an acceptable solution - it's not right to abuse the structures in this way - but it did the trick, and retains BC :)

Well,  do we want to merge it or no?
If it was OK for you,  it may be OK for others.

In future our efforts will focus towards a BIP174 [PSBT](https://github.com/bitcoin/bitcoin/pull/12136) compliant implementation,  and the removal of `TransactionBuilder`,  so,  I see no reason why we couldn't support this until then?

dcousens:
@afk11 do we keep this open or?

PR #907, opened 2017-09-26
rm bitcoin.script circular dependencies with templates

dcousens:
Resolves https://github.com/bitcoinjs/bitcoinjs-lib/issues/905 (hopefully)

@maxcloudwgh please test

Issue #901, opened 2017-09-22
TransactionBuilder.fromTransaction (Transaction.fromHex ()) and segwit input signing


dakk:
Version: Bitcoinjs-0.3.2 and also bitcoinjs/bitcoinjs-lib#fixes

I'm trying to integrate segwit in my walleting software, but I'm stuck in one problem. The flow is the following:

1. I create a segwit address from 2of3 p2sh keys
2. I send some testnet coin to this address
3. I get the unspent tx
4. I create the tx with this unspent as input
5. I generate the incomplete hex, passing it to another software
6. I then try to sign the transaction with the following code:

```javascript
var signTxSegwit = function (txhex, pubkeys, privkey, utxos, n) {
var txb = new bitcoinjs.TransactionBuilder.fromTransaction (
	bitcoinjs.Transaction.fromHex (txhex), bitcoinjs.networks.testnet);
var upair = bitcoinjs.ECPair.fromWIF(privkey, bitcoinjs.networks.testnet);
var pubkeys_raw = pubkeys.map(function (hex) { return new Buffer(hex, 'hex'); });
var witnessScript = bitcoinjs.script.multisig.output.encode (n, pubkeys_raw);
var redeemScript = bitcoinjs.script.witnessScriptHash.output.encode (bitcoinjs.crypto.sha256(witnessScript));
var scriptPubKey = bitcoinjs.script.scriptHash.output.encode (bitcoinjs.crypto.hash160(redeemScript));
var address = bitcoinjs.address.fromOutputScript(scriptPubKey, bitcoinjs.networks.testnet);
	
for (var j = 0; j < txb.tx.ins.length; j++) {
	txb.sign (j, upair, redeemScript, null, parseInt (utxos[j].value * 100000000), witnessScript);
}

var tx = txb.build ();
return tx.toHex ();
}
```

The program fail at txb.sign, with this error:
```
   Message:
     Error: Expected property "2" of type Satoshi, got undefined
   Stacktrace:
     Error
    at TfTypeError.Error (native)
    at new TfTypeError (/home/dakk/Repositories/MyRepos/backend/node_modules/typeforce/errors.js:43:24)
    at typeforce (/home/dakk/Repositories/MyRepos/backend/node_modules/typeforce/index.js:193:11)
    at /home/dakk/Repositories/MyRepos/backend/node_modules/typeforce/index.js:152:18
    at Array.every (native)
[...]

PR #910, opened 2017-09-26
Fixes for issue 901 continued

dcousens:
@afk11 maybe this is what you intended in #909 ?

f4a83f8 is a follow up to #907 which was having issues locally

### DIFF
--- a/src/transaction_builder.js
+++ b/src/transaction_builder.js
@@ -2,13 +2,14 @@ var Buffer = require('safe-buffer').Buffer
 var baddress = require('./address')
 var bcrypto = require('./crypto')
 var bscript = require('./script')
+var btemplates = require('./templates')
 var networks = require('./networks')
 var ops = require('bitcoin-ops')
 var typeforce = require('typeforce')
 var types = require('./types')
-var scriptTypes = bscript.types
-var SIGNABLE = [bscript.types.P2PKH, bscript.types.P2PK, bscript.types.MULTISIG]
-var P2SH = SIGNABLE.concat([bscript.types.P2WPKH, bscript.types.P2WSH])
+var scriptTypes = btemplates.types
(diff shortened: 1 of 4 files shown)


src/transaction_builder.js
```diff
@@ -686,7 +682,13 @@ TransactionBuilder.prototype.sign = function (vin, keyPair, redeemScript, hashTy
 
   var kpPubKey = keyPair.getPublicKeyBuffer()
   if (!canSign(input)) {
-    prepareInput(input, kpPubKey, redeemScript, witnessValue, witnessScript)
+    if (witnessValue !== undefined) {
+      if (input.value !== undefined && input.value !== witnessValue) throw new Error('Input didn\'t match witnessValue')
+      typeforce(types.Satoshi, witnessValue)
+      input.value = witnessValue
+    }
```
dcousens: @afk11

afk11: I think this side steps the call to fixMultisigOrder, which would remove any signatures from the array which aren't valid?

dcousens: @afk11 this isn't related to MS order,  it simply ensures that `.value` is added on if missing

afk11:
@dcousens ah, I see the confusion.. I just found the exact same issue as mine recently: https://github.com/bitcoinjs/bitcoinjs-lib/issues/901#event-1266583387 which is indeed related to the input.value not being present, but the fixMsOrder demonstrates how it can be a problem if you leave it until `sign` - invalid signatures remain in the array unless fixMultisigOrder is called. 

So my solution was #909, which means that signatures fail and aren't extracted IF the transaction, which was passed to TransactionBuilder.fromTransaction as inputs which have the input.value for segwit outputs. 

Ping @dakk since we are having identical issues.

dcousens:
As discussed on Slack,  we'll need to incorporate the second half of the complete fix for this problem in #909

dakk:
I confirm it works now with my test case, excellent work

Issue #862, opened 2017-08-23
Add tests for scripts.witness*.input


dcousens:

dcousens:
Closing in favour of #911

PR #911, opened 2017-09-26
Add tests for scripts.witness*.input and fixes

dcousens:
Resolves https://github.com/bitcoinjs/bitcoinjs-lib/issues/862

`witnessScriptHash` was a bit of a cluster... returned `redeemScriptSig`, a script, not witness data.
Could be a breaking change, but I'd consider it a bug tbh.


dcousens:
We don't even use `witnessScriptHash.input` in `TransactionBuilder`...
We build it manually:
``` js
        witness = buildStack(input.witnessScriptType, input.signatures, input.pubKeys, allowIncomplete)
        witness.push(input.witnessScript)
```

test/templates.js
```diff
@@ -347,6 +350,31 @@ describe('script-templates', function () {
     })
   })
 
+  describe('witnessPubKeyHash.input', function () {
+    fixtures.valid.forEach(function (f) {
+      if (f.type !== 'pubkeyhash' && f.type !== 'witnesspubkeyhash') return
+      if (!f.inputStack) return
+
+      var pubKey = Buffer.from(f.pubKey, 'hex')
```
junderw: use fromHex here?

dcousens: @junderw I was only using `fromHex` in the `.map(fromHex)` case,   if we had ES6 lambdas,  I wouldn't have bothered.

dcousens:
Published `v3.2.1`

PR #917, opened 2017-10-26
Correctly decoding op_return data

runn1ng:
As currently written, nulldata.decode doesn't return correct results with longer data.

An example:

[This transaction]( https://blockchain.info/tx/8e533c09bae92c9b47ef9bc26b58707550f2a54a3b9238dc1a7545273523b84f?format=hex) when decoded starts data with 0x50, but that is the amount of data after OP_PUSHDATA1

It was working on smaller nulldata, because it stripped the one OP_PUSHDATA correctly

I will write tests that trigger this issue in old code, but don't in this fix, later

### DIFF
--- a/src/templates/nulldata.js
+++ b/src/templates/nulldata.js
@@ -6,10 +6,8 @@ var typeforce = require('typeforce')
 var OPS = require('bitcoin-ops')
 
 function check (script) {
-  var buffer = bscript.compile(script)
-
-  return buffer.length > 1 &&
-    buffer[0] === OPS.OP_RETURN
+  return script.length === 2 &&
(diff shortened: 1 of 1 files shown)


runn1ng:
Oh, I didn't realize `check` is also used in detection code, I might have broken that

junderw:
Try parsing that with this branch please?
https://github.com/bitcoinjs/bitcoinjs-lib/pull/859

runn1ng:
Didn't notice that branch. From the outside, it does more or less the same.

dcousens:
Close in fovour of #859?

runn1ng:
Yep it looks the same.

PR #919, opened 2017-11-01
allow index of hardned child account xpub for bip44 wallets

Blutchie:


junderw:
This breaks the tests, and understandably so, as the derive function relies on the index being accurate.

https://github.com/Blutchie/bitcoinjs-lib/blob/9e4aec1beb90da5228d2979eff3e02593b43a95b/src/hdnode.js#L211

This is where the index is checked for being higher than that number, and if it is, the hardened derivation will be used.

Why exactly did you feel this change was necessary? Was there an issue? Could you show us your code that experienced the issue?

Closing this PR, but please respond with any issues you had that inspired you to make this pull request.

dcousens:
I think @Blutchie is assuming that `.index` should be re-normalized for public accessibility.
Understandable I suppose,  but we for us that would throw away meaningful information without adding more state.

If you need a `getDerivationIndex` function that normalizes the value,  that could be added,  if you have a use case.

Blutchie:
Hi @junderw  and @dcousens, I'll try to be clear in why I made this change.
I'm kind of new to developing for the blockchain, but have theoretical understanding about how this works.

For reporting purposes I'm making a POC for a read only wallet. This will receive public master keys from a backend, without knowing the private key. Because it is in a very early stage I use coinomi as the main wallet. For the POC I went to the Bitcoin account details to get the xpub. Then I derived the child addresses with bitcore-lib, which worked. But because I'm working with a typescript project, I didn't like that there were no typings available. Then I googled and found this library with available typings, so that was great!

Once I figured out how the library should work I ended up with an "Invalid Index"  error.
image

After some research I figured out that index 2147483648 meant it was a hardened key with derivable child keys. I tested if the addresses of the derived child keys matched the one in the wallet and it did. This in combination with the working bitcore-lib, I thought it was a design fault, so I corrected it.

I just tried it again with a new wallet and got the same issues. The master xpub of Bitcoin is xpub661MyMwKB68aT5DDeBkbbBKAp5T4AMjTTQ3CA9QEoY2C14d74zgWNAmQzQZNY6Et5uiqtkdgCvfSMHQ7Ra2QQUWNnw8xxf9cKp44863CQ3Q.

If this is because of a wrong approach, then I'm sorry for the faulty PR and like to hear what is the correct approach.

dcousens:
If you have an invalid index... it means you're `depth` value has been encoded wrong by some faulty library...

Blutchie:
ow ok, so probably a bug in their library. What I see it's fork of bitcoinj, but it's far behind the master branch: https://github.com/Coinomi/bitcoinj, or they went private by now.

I'll check another wallet later this week, to see if the issue is indeed caused by coinomi.

Blutchie:
I just checked what happens with en xpub exported from the exodus wallet. Also this one had an index of 2147483648, but a depth of 3. So that would indeed make it likely that the encoding of the depth goes wrong in coinomi. Instead of making a fix for this wallet, I will just export the xpub's myself from the seed.

Thnx for pointing me to the right direction!

dcousens:
Happy to help @Blutchie,  always feel free to ask questions :+1:

junderw:
that xpub you pasted is broken.

an extended key of the depth 0 (aka Master) MUST have an index of 0 and can not be hardened.

It looks like their wallet is broken for xpub derivation. Please report the issue to Coinomi.

PR #927, opened 2017-11-17
Support for BIP49 YPUB/UPUB

plasticlobster:
The only thing missing from bitcoinjs-lib to support BIP49 is the network type designator. This simple change allows the HD wallet to support BIP49 YPUB's (or testnet UPUB's). This allows for the import of pure segwit HD wallets.

See also: https://github.com/bitcoin/bips/blob/master/bip-0049.mediawiki

Understanding that BIP49 is draft, you may reject this change, but I've made this change for a project and figured submitting a PR would be an appropriate offer to allow you to implement this simple functionality.

### DIFF
--- a/src/hdnode.js
+++ b/src/hdnode.js
@@ -64,7 +64,9 @@ HDNode.fromBase58 = function (string, networks) {
   if (Array.isArray(networks)) {
     network = networks.filter(function (x) {
       return version === x.bip32.private ||
-             version === x.bip32.public
+             version === x.bip32.public ||
+             version === x.bip49.private ||
+             version === x.bip49.public
     }).pop()
 
     if (!network) throw new Error('Unknown network version')
@@ -75,7 +77,9 @@ HDNode.fromBase58 = function (string, networks) {
   }
 
   if (version !== network.bip32.private &&
-    version !== network.bip32.public) throw new Error('Invalid network version')
+    version !== network.bip32.public &&
+    version !== network.bip49.private &&
+    version !== network.bip49.public) throw new Error('Invalid network version')
 
   // 1 byte: depth: 0x00 for master nodes, 0x01 for level-1 descendants, ...
   var depth = buffer[4]
--- a/src/networks.js
+++ b/src/networks.js
@@ -9,6 +9,10 @@ module.exports = {
       public: 0x0488b21e,
       private: 0x0488ade4
     },
+    bip49: {
+      public: 0x049D7CB2,
+      private: 0x049D7878
+    },
     pubKeyHash: 0x00,
     scriptHash: 0x05,
     wif: 0x80
@@ -19,6 +23,10 @@ module.exports = {
     bip32: {
       public: 0x043587cf,
       private: 0x04358394
+    },
+    bip49: {
+      public: 0x044A5262,
+      private: 0x044A4E28
     },
     pubKeyHash: 0x6f,
     scriptHash: 0xc4,


dcousens:
From reading that BIP,  I don't see the network constants you've specified...

Instead,  the only difference I saw was that the `purpose` constant was `49` not `44`?

> To derive a public key from the root account, this BIP uses the same account-structure as defined in BIP 44, but only uses a different purpose value to indicate the different transaction serialization method.

plasticlobster:
@dcousens 

My address generation happens outside of bitcoinjs-lib, and I'm not sure if there are more implications to fully supporting an implementation of BIP49 as a result. 

The use case was for an electrum wallet. See:
http://docs.electrum.org/en/latest/seedphrase.html#list-of-reserved-numbers
and
https://github.com/spesmilo/electrum/blob/0a648e2b1cdb7852d80c0626c62984ef01c1c7e9/lib/bitcoin.py#L50

The usage of these constants (which translate to ypub and upub) allow for auto-detection of a bip49-style wallet. When using a bip32-style xpub, it's impossible to determine whether the wallet is bip32 or bip49. While the constants do not appear in a BIP, they are the constants currently in use by wallet providers like electrum.

dcousens:
Is this related to https://github.com/bitcoinjs/bitcoinjs-lib/issues/934

src/hdnode.js
```diff
@@ -64,7 +64,9 @@ HDNode.fromBase58 = function (string, networks) {
   if (Array.isArray(networks)) {
     network = networks.filter(function (x) {
       return version === x.bip32.private ||
-             version === x.bip32.public
+             version === x.bip32.public ||
+             version === x.bip49.private ||
```
afk11: I think these semantics should be opt-in. Imagine someone supported Bitcoin Cash, and imported keys. There's not currently a bitcoin-cash network in the library, so they're using the bitcoin network for that. Imagine they didn't read the release notes, and allowed users to import a P2SH(P2WPKH) address, and funded that on Bitcoin Cash.. They're unspendable without talking to a miner.

I think this feature should be opt in, so a new function `bip44TypefromBase58` or something that describes what the function will do (parse a HD key with an embedded script type, constructing a hdnode which produces the right address instead of always P2PKH)

afk11:
In general I think the proposal has its merits, it's nice to support what some of the major wallets are doing. I have some comments on the PR and proposal in general. 

See review, should be opt in instead of in `fromBase58`... Maybe introduce it as `fromBip49ypub` for now, or consider supporting more of the script types suggested by others and make it `fromExtendedBip32Version` or something that deals with all of them?

The feature should allow us to seamlessly generate the correct script and therefore address here:
https://github.com/plasticlobster/bitcoinjs-lib/blob/8f39b516946805c6b4640d1df0fdd33a0ebedba7/src/hdnode.js#L129 so if someone was expecting a P2SH(P2WPKH) address after using this code, they'd be surprised..

Generally, concept ACK once it's opt in.

src/hdnode.js
dcousens: @afk11 I don't understand,  what about these new BIP49 constants would trap users into creating unspendable addresses?

dcousens: Isn't the risk that any P2SH(P2WPKH) scripts are unspendable by BCASH nodes?

dcousens:
Shouldn't `getAddress` be changed too?

dcousens:
@dabura667 @junderw where does `zpub` fit into this?

dabura667:
@dcousens Honestly... I think maybe this is a little bit early.

I would really like to see a BIP from @ecdsa before we merge this.

Also, I think this is insufficient.

Electrum also has extra version bytes for WIF private keys. We should probably include those if we include this.

zpub/prv are for native P2WPKH.

Currently, ECPair and HDNode have a network attribute which refers to a network in networks.

I think they should also have an attribute for "scriptType" which will tell us what to use when using toWIF or toBase58Check respectively. So we can look at the scriptType, and use that to guide us as to what to fetch from the network object.

I also dislike calling this bip49... perhaps keep all the legacy network structure and add a new key:

ie.

```javascript
  bitcoin: {
    messagePrefix: '\x18Bitcoin Signed Message:\n',
    bech32: 'bc',
    bip32: {
      public: 0x0488b21e,
      private: 0x0488ade4
    },
    pubKeyHash: 0x00,
    scriptHash: 0x05,
    wif: 0x80,
    versionBytes: {
      ECPair: {
        'P2PKH': {
          wif: 0x80
        },
        'P2SH(P2WPKH)': {
          wif: 0x82
        },
        'P2WPKH': {
          wif: 0x81
        }
      },
      HDNode: {
        'P2PKH': {
          public: 0x0488b21e,
          private: 0x0488ade4
        },
        'P2SH(P2WPKH)': {
          public: 0x049d7cb2,
          private: 0x049d7878
        },
        'P2WPKH': {
          public: 0x04b24746,
          private: 0x04b2430c
        }
      }
    }
  },
```

(WIF values weren't in wiki, but they are in the bitcoin.py.
https://github.com/spesmilo/electrum/blob/0a648e2b1cdb7852d80c0626c62984ef01c1c7e9/lib/bitcoin.py#L483
The function for encoding WIF keys is right below it and you see they are adding those numbers to 128, which is essentially ORing them.)

dcousens:
What a mess.

junderw:
Yup. It is.

Big old mess.

But I have been thinking for a while that now we have different ways to use private keys, we need a state similar to compressed that can store the script type. The way this is represented when exported or imported can be added later when BIPs are written.

What do you think? @dcousens

dcousens:
@junderw it *feels* like the version bytes are useless as network information,  and instead,  acting as flags for other meta information.

Maybe we should support non-strict `fromBase58` decoding? Sigh.

dcousens:
Can we simply tell these muppets to please stop abusing the BIP32 standard and instead write another BIP that stores this information in a simple flag.

junderw:
IMO we have bech32, why not use it? lol.

junderw:
hrp could be p2shp2wpkh or something

junderw:
xpubp2shp2wpkh1sdfouofgsdfijo9df8ehosf8feojosihfhe8fh

etc etc

dcousens:
@junderw I have an idea,  lets write a BIP that captures the path information *AND* the address type(s).

junderw:
While that comic is true.

Character encoding has a clear winner.

If you don't use UTF-8 you are wrong.

dcousens:
@junderw would `ROOT_NODE | (PATH | ADDRESS_TYPE | IS_PRIMARY | IS_CHANGE | GAP_LIMIT)+` suit most of these cases?
I guess you probably want some extension possibilities for things like multisig...

Why is the distinction necessary anyway?  Who is actually passing around these Base58 strings rather than a mnemonic?

ecdsa:
@dabura667 we will submit a BIP

junderw:
> Why is the distinction necessary anyway?

Mainly for xpubs. A lot of people pass around xpubs for watch-only and merchant services.

dcousens:
> Mainly for xpubs. A lot of people pass around xpubs for watch-only and merchant services.

Everyone read the [**Security and Implications** part of BIP32](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#implications),  right?
Why has this become a thing?

ecdsa:
> @ecdsa are you open to engaging in a discussion about a more generic standard e.g like the above?

can you be more specific?

junderw:
> Why has this become a thing?

Have you ever used Electrum? Electrum made it a thing since before BIP32 was even a thing.

The rationale is that you can't export individual private keys from Electrum, so it's not a problem. (iirc)

mautematico:
>Why is this necessary anyway? Who is actually passing around these base58 strings rather than a mnemonic?
To create watch-only wallets for internal (but secure) use; just like I am trying to do right now.

>The rationale is that you can't export individual private keys from Electrum, so it's not a problem. (iirc)
Yes, you can export individual private keys from electrum.
Wallet >> Private Keys >> Export

dabura667:
> watch-only wallets for internal (but secure) use

Watch-only wallets are easy to hack: Just replace the xpub with a hacker's xpub. Now everyone will send money to the hacker and not you.

mautematico:
>Watch-only wallets are easy to hack: Just replace the xpub with a hacker's xpub. Now everyone will send money to the hacker and not you.

That is pointless; exactly the same can be done regardless if you are using xpub or not. So no "xpub threat" here.

dabura667:
This PR is not complete, that’s why it’s not working.

clarkmoody:
We also have the [SLIP-32](https://github.com/satoshilabs/slips/blob/master/slip-0032.md) proposal out there, which encodes the derivation path within the serialized `xpub` data payload.

plasticlobster:
@mautematico You're right. getAddress would need modification to get the proper address, but you can do it as follows:

```
const Bitcoin = require('bitcoinjs-lib');

Bitcoin.address.fromOutputScript(
  Bitcoin.script.scriptHash.output.encode(
    Bitcoin.crypto.hash160(
      Bitcoin.script.witnessPubKeyHash.output.encode(
        Bitcoin.crypto.hash160(
          Bitcoin.HDNode.fromBase58('ypub6XMTwf6NSvfzYYgVgdNWRNfMTiQt4rSjZbEk8qoCnBGhUD2rsgZ2A8pexgzaGLKgySZxqxrctDpAVU8QtfxqfX8QUAhtFmGFUFx9B51TVg8', Bitcoin.networks.bitcoin)
          .derive(0)
          .derive(27)
          .getPublicKeyBuffer()
        )
      )
    )
  ),
  Bitcoin.networks.bitcoin
)
```

Obviously a change like this would have to be implemented in getAddress to make this easier, but I don't even want to pretend I understand the coding requirements of this project enough to propose that change.

dcousens:
For context,  using the new API in `bitcoinjs-playground` (a testing bed for new ideas),  the above code becomes:

``` js
const bitcoin = require('bitcoinjs-lib')
const pubKey = bitcoin.HDNode.fromBase58('ypub6XMTwf6NSvfzYYgVgdNWRNfMTiQt4rSjZbEk8qoCnBGhUD2rsgZ2A8pexgzaGLKgySZxqxrctDpAVU8QtfxqfX8QUAhtFmGFUFx9B51TVg8')
  .derive(0)
  .derive(27)
  .getPublicKeyBuffer()

const { address } = bitcoin.payments.p2sh({
  redeem: bitcoin.payments.p2wpk({
    pubkey: pubKey
  })
})
```

Thoughts?

dcousens:
ping @afk11

afk11:
I think the PR needs to address generating the scriptPubKey/address for it to be complete, first of all.

It'd really be nice for it to work with the other BIP's that do this custom HD prefix to indicate an address. There's BIP49, and BIP84, each has custom HD prefixes implementing the same logic, which is constructing a HD key capable of producing a P2SH|P2WPKH address, or P2WPKH respectively. Electrum takes this a step further, and defines 5 prefix/scriptTypes for which they do this. 

I've been thinking about the missing address generation part of this PR: could we add an optional parameter to the hdkey constructor, a closure accepting a public or private key, and returns a {scriptPubKey:, rs:, ws:, address: } when called? We can add the optional parameter to the generation methods as well. Derivations could pass on the closure to children as well.

hdkey.getAddress could then check if the closure was set, and call it if so. If it wasn't set, just return P2PKH. A default closure of P2PKH could be how we make this work, actually. Maybe we want to add another function to hdkey, that returns {scriptPubKey:, rs:, ws:, address: }, since getAddress exclusively returns the address. This information would be really handy for signing time!

That way, we can inject different kinds of closures to work with all the types that electrum supports (http://docs.electrum.org/en/latest/seedphrase.html):
 - P2PKH (xprv)
 - P2SH|P2PKH (xprv) 
 - P2SH|P2WPKH (yprv)
 - P2WSH|P2PKH (Yprv)
 - P2WPKH (zprv)
 - P2WSH ???? (this doesn't say which type, just p2wsh)

It'd be nice to support the 5 (or more) serialization formats as well. One issue I have with the PR is that the deserialization will succeed even for people who don't want the feature. It'd be nice if people could control which prefix/scriptTypes are active during a deserialization? Could we find a way to pass in the extra prefixes/scriptTypes/closures for consideration in a call

Maybe the prefixes/scriptTypes could be defined in the network, then a selection of these can be passed into deserialize/serialize methods?

clarkmoody:
We're putting these registered HD prefix types in SLIP-0132, to pull the
reference out of any one implementation's docs.

https://github.com/satoshilabs/slips/blob/master/slip-0132.md



-Clark

afk11:
I took a quick look at getting getAddress to work, but a mocked getAddress test fails, since keyPair.getAddress was overwritten and now the results aren't identical. 

Makes me wonder if keyPair should have getAddress, but anyways. I'll comment the test for now since the addresses will vary going forward!

dcousens:
Related to https://github.com/bitcoin/bips/pull/673 ?

jackylimel:
@mautematico may I have one question?

for this code:

```
Bitcoin.HDNode.fromBase58('ypub6XMTwf6NSvfzYYgVgdNWRNfMTiQt4rSjZbEk8qoCnBGhUD2rsgZ2A8pexgzaGLKgySZxqxrctDpAVU8QtfxqfX8QUAhtFmGFUFx9B51TVg8', Bitcoin.networks.bitcoin)
    .derive(0)
    .derive(27)
    .getAddress()
```

I'd like to know what's the meaning of `derive(0)` and then `derive(27)`, could you please explain a little bit? 

Thank you very much.
Jacky.

mautematico:
@jackylimel sure,

have you read about bip32 derivation paths? They let you generate multiple public-private key pairs from a single initial master public-private pair (ypub & ypriv).

twice `.derive(0).derive(0)` is the same as `.derivePath("m/0/0")`

https://bitcoin.stackexchange.com/questions/56561/bip32-derivation-path
http://bip32.org/

Overtorment:
for anyone waiting for ypub support in bitcoinjs, heres a quick workaround I borrowed from https://github.com/bitcoinjs/bitcoinjs-lib/issues/966

```
    // bitcoinjs does not support ypub yet, so we just convert it from xpub
    const b58 = require('bs58check')
    let data = b58.decode(xpub)
    data = data.slice(4)
    data = Buffer.concat([Buffer.from('049d7cb2','hex'), data])
    return b58.encode(data);
```

sneurlax:
@Overtorment, thanks for that.  This should be the zpub equivalent workaround:

```
let data = b58.decode(xpub)
data = data.slice(4);
data = Buffer.concat([Buffer.from('04b24746','hex'), data]); // see https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2017-September/014907.html for tenative (non-BIP-official) version bytes
return b58.encode(data);
```

dcousens:
What do I do here?
With the removal of `getAddress`...

Thoughts?

youssefgh:
@dcousens i believe that "YPUB/UPUB" is now a part of BIP49. See [Extended Key Version](https://github.com/bitcoin/bips/commit/4413bdc46ffa779987ff2ffb9ac9257744e35316#diff-bdcbea0821b59d69d5e836bdf9a3db24)
What do you think ?

PR #931, opened 2017-11-22
increase max feerate sanity check from 1000 to 2500

Melvillian:
image

Back in Nov 2016 when we first set the maximum fee rate sanity check the average feerate for inclusion in 1-2 blocks was ~70 sat/byte and there was relatively low variability. Nowadays we're seeing an increase in variability, with the fee rates spiking to over ~1000 sat/byte.

We've had multiple customers trigger this "absurd fees" error when trying to build their transactions. 

I think the current fee market warrants increasing this to a higher value. I think 2500 sat/byte is appropriate because it will be handle more than double the current largest spike we've seen, and it's roughly 10x what the average feerate has been for the past 10 months.

### DIFF
--- a/src/transaction_builder.js
+++ b/src/transaction_builder.js
@@ -469,7 +469,7 @@ function TransactionBuilder (network, maximumFeeRate) {
   this.network = network || networks.bitcoin
 
   // WARNING: This is __NOT__ to be relied on, its just another potential safety mechanism (safety in-depth)
-  this.maximumFeeRate = maximumFeeRate || 1000
+  this.maximumFeeRate = maximumFeeRate || 2500
 
   this.inputs = []
   this.tx = new Transaction()


dcousens:
Published as `v3.3.1`

Thanks @Melvillian.

For reference,  users can set the maximum using the 2nd parameter of the `TransactionBuilder` constructor.

dcousens:
> it's roughly 10x what the average feerate has been for the past 10 months.

I think this should be our baseline generally - I'd rather have users submitting complaints then their software bite them due to some fault.

dcousens:
Had to fix the tests with https://github.com/bitcoinjs/bitcoinjs-lib/pull/942/commits/34df503421af0aa41315c47763d7ce90d5cadadb

PR #942, opened 2017-11-30
enforce decodeStack receives a stack array

dcousens:
Burned by this twice, a `Buffer` can pass `check` because of ambiguity in `decompile`,  returning a `.slice` of bytes, not chunks, which makes zero sense.


dcousens:
Added https://github.com/bitcoinjs/bitcoinjs-lib/pull/942/commits/34df503421af0aa41315c47763d7ce90d5cadadb to fix the unit tests

PR #944, opened 2017-12-01
Add BIP68 / BIP112 CSV integration tests 

dcousens:
https://github.com/bitcoin/bips/blob/master/bip-0068.mediawiki
https://github.com/bitcoin/bips/blob/master/bip-0112.mediawiki


test/integration/csv.js
```diff
@@ -0,0 +1,57 @@
...
+
+      // {Alice's signature}
+      var redeemScriptSig = bitcoin.script.scriptHash.input.encode([
+        alice.sign(signatureHash).toScriptSignature(hashType)
+      ], redeemScript)
+
+      txRaw.setInputScript(0, redeemScriptSig)
+
+      // TODO: fix
+      testnetUtils.transactions.propagate(txRaw.toHex(), done)
```
dabura667: ummm... unless we wait for the faucet deposit to confirm, and then wait an additional hour (MBT) after the MBT of the block we were confirmed in... then this won't propagate...

Maybe assert that it throws?... CSV is hard to automate testing...

dcousens: @dabura667 not if you have a reg-test server :+1:

junderw:
ACK

PR #952, opened 2017-12-07
Change package name to bitgo-bitcoinjs-lib

Melvillian:
We need this to distinguish between our package and bitcoinjs-lib


Melvillian:
I branched this off of an existing branch, I'm closing this and redoing it off master

PR #968, opened 2017-12-27
fix bitcoin-gold/bitcoin-cash test error

ysy:


dakk:
Not sure if this is the right place, but is it possible to have an example of using bitcoincash library to send funds from a 2-of-3 multisig address? I'm getting "16: mandatory-script-verify-flag-failed (Script evaluated without error but finished with a false/empty top stack element). Code:-26" but I don't understand if I'm using it right.

junderw:
This doesn't do anything. Closing until further clarification.

PR #970, opened 2017-12-28
crypto.sha256 should receive a Buffer as an input rather than a string

kaminskypavel:
checked and it works with buffers.

if you wanted to use string implicitly, the function parameter should be renamed  from buffer to string.
https://github.com/bitcoinjs/bitcoinjs-lib/blob/f4caaf42e7b58332d74c1540f88bcda7e55b82e6/src/crypto.js#L11

### DIFF
--- a/test/integration/addresses.js
+++ b/test/integration/addresses.js
@@ -17,7 +17,7 @@ describe('bitcoinjs-lib (addresses)', function () {
   })
 
   it('can generate an address from a SHA256 hash', function () {
-    var hash = bitcoin.crypto.sha256('correct horse battery staple')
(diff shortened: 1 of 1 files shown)


dcousens:
Unclear if this requires a `typeforce` pre-amble?

dabura667:
@dcousens tbh there should be a deprecation warning if you do that...

But imo, if the upstream standard crypto.Hash accepts string and implicitly converts to utf8 Buffer, I don't know why we should add extra type enforcement...

dcousens:
@dabura667 that was my thoughts,  but I suppose if that is the idea,  then this PR is mute (and should be reverted).

PR #985, opened 2018-01-16
Fix integration test 3PBP

junderw:
Fixed the integration tests to use smartbit.

### DIFF
--- a/test/integration/addresses.js
+++ b/test/integration/addresses.js
@@ -102,17 +102,13 @@ describe('bitcoinjs-lib (addresses)', function () {
     var address = keyPair.getAddress()
 
     dhttp({
-      method: 'POST',
-      url: 'https://api.ei8ht.com.au/3/addrtxs',
-      body: {
-        addrs: [address],
-        height: 0
-      }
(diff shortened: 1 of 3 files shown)


junderw:
hmmm... ok, I guess the problem wasn't the endpoint but the faucet returns unspents without them actually being propagated...

hmmmm....

dcousens:
I don't mind what endpoint we use.   But,  we want the green tick :smiley:

dcousens:
I'm presently working towards a reg-test server for us to use to avoid these issues...

test/integration/addresses.js
```diff
@@ -102,17 +102,13 @@ describe('bitcoinjs-lib (addresses)', function () {
     var address = keyPair.getAddress()
 
     dhttp({
-      method: 'POST',
```
EasonWang01: This line need to keep or it will result in `Error: Page not found or request method incorrect for this resource (GET)`

dcousens:
Closing in favour of https://github.com/bitcoinjs/bitcoinjs-lib/pull/1001

Issue #995, opened 2018-01-23
Derive public LTC addresses from an extended public LTC address in xpub format


nnymsgthb:
Hello everybody!
I'm very new to JS, nodeJS and bitcoinjs, so please excuse my noobidity.

Im trying to derive public LTC addresses from an extended LTC address in xpub format (created by Jaxx and exported by Exodus). For demonstration purposes, I created (in Jaxx) and exported (in Exodus) a wallet with the following mnemonic:

_quit disagree kiss among move interest orchard trade bullet forward usual neither_

The extended public LTC address is:
_xpub6CAp1aH6mzGrH3hkCuaZPCDrgTMkPm55xhBpNETbPPuhAxxoprvsiHBDipL7ePY6aAiZZThyWCN5qiKcCZxtjCJLtVjAeVryZCtamBKyi3m_

And the first public LTC address is:
_LeUD63jWGkLTohEWq6t4gyqpithtnReXEk_

I ran some tests with [https://iancoleman.io/bip39/](url) and they yielded the info, that the derivation path is: _m/44'/2'/0'/0_

So far, I got the following code:
```var bitcoin = require('bitcoinjs-lib');
var network = bitcoin.networks.litecoinXprv;

var xpub = 'xpub6CAp1aH6mzGrH3hkCuaZPCDrgTMkPm55xhBpNETbPPuhAxxoprvsiHBDipL7ePY6aAiZZThyWCN5qiKcCZxtjCJLtVjAeVryZCtamBKyi3m'

var node = bitcoin.HDNode.fromBase58(xpub,network)

var address = bitcoin.HDNode.fromBase58(xpub,network).derivePath("m/44'/2'/0'/0").derive(0).getAddress();

// Possible alternative?
//var address = node.derive(44).derive(2).derive(0).derive(0).derive(0).getAddress()

console.log("First address: " + address)
```

Thanks for bitcoinjs and any provided support.

EDIT: I added the missing apostrophes.

dabura667:
Missing apostrophes.

nnymsgthb:
Thanks @dabura667. Absolutely right, therefore I added the missing apostrophes. 

It still doesn't work, but I think I might finally have figured out why:
The apostrophes indicate hardened addresses and hardened addresses cannot be derived just from the extended public key but I need to provide the extended private key. Is that right?

[...]

PR #996, opened 2018-01-24
Support for extended LTC-keys in xprv/xpub format

nnymsgthb:
Because LTC is natively supported by bitcoinjs-lib I think it is justifiedd to support extended LTC-keys in xprv/xpub format, especially because this format is used by some common wallets (e.g. Jaxx and I believe Exodus). The network definition was taken from @iancolemn (see [https://github.com/iancoleman/bip39/blob/master/src/js/bitcoinjs-extensions.js](url)).

### DIFF
--- a/src/networks.js
+++ b/src/networks.js
@@ -33,5 +33,15 @@ module.exports = {
     pubKeyHash: 0x30,
     scriptHash: 0x32,
     wif: 0xb0
+  },
+  litecoinXprv: {
+    messagePrefix: '\x19Litecoin Signed Message:\n',
+    bip32: {
+      public: 0x0488b21e,
+      private: 0x0488ade4,
+    },
+    pubKeyHash: 0x30,
+    scriptHash: 0x32,
+    wif: 0xb0
   }
 }


dabura667:
Ugh...

What is the point of standards if no one uses them... *sigh*

@dcousens perhaps we should allow HDNodes to prefer the network object and only soft-warn if it differs from the standard.

Switching to a soft-warn rather than a hard fail could also (in an ugly way) fix the ypub / zpub problem.

dabura667:
NACK on the new network idea, though.

dcousens:
@dabura667 my thoughts was,  only enforce the network object strictly __if__ they provide it.

dcousens:
@nnymsgthb this isn't the solution to this problem.

For your purposes,  you can simply use:
```
bitcoin.HDNode.fromBase58(..., [bitcoin, litecoin,  custom1, custom2])

// or

bitcoin.HDNode.fromBase58(..., custom1)
```

Where `custom` is your network object defined above.

dcousens:
Related https://github.com/bitcoinjs/bitcoinjs-lib/pull/927

nnymsgthb:
Related #995

vincenzourbisaglia:
@dabura667 doesn't the litecoin-project define xprv/xpub?
https://github.com/litecoin-project/litecoin/blob/v0.13.2.1/src/chainparams.cpp#L130

PR #1007, opened 2018-02-19
upgrade test dependencies

hackerrdave:
Upgrading:
Mocha 3 -> 5
NYC 10 -> 11
Sinon 1 -> 4

Adding:
sinon-test (was split out as separate module from sinon)


dcousens:
Thanks @hackerrdave

PR #1014, opened 2018-03-04
Support ypub/zprv, plus custom addresses

afk11:
This PR adds a new parameter to the hdnode constructor. The parameter describes a bip32 prefix, and has the public & private prefix values, plus a scriptFactory field. This exposes a function, `convert(ECPair key)` and produces:
```
{
  scriptPubKey: Buffer,
  signData: {
    redeemScript: x,
    witnessScript: y
  }
}
```
 - also updated fromSeedHex, so you can persistently mark a key and it's children with a prefix
 - updated fromBase58, if no list of networks is provided (in which case, we only compare against `.bip32`, we assume the network is an object (or take bitcoin as the default), and in this case will check the prefix from the key against the `.private` and `.public` members of the list of prefixes. 
 - if no prefix is set, we default to p2pkh.
- getAddress was modified to use the script factory, by doing address.fromOutputScript() on the resulting scriptPubKey - please see the test I commented out, and lets talk about that! https://github.com/bitcoinjs/bitcoinjs-lib/pull/1014/files#diff-4db2c2d53279dc2a5cf3fee520cbb23eR126
 - added the bip49 and bip84 prefixes to networks. 
 - added tests, basically just covering the expected behavior of a key with a prefix, using the fixtures from slip132

Not sure what direction this are going with the new templates/script stuff, so all ears for feedback on the stuff in keytoscript!

### DIFF
--- a/src/keytoscript.js
+++ b/src/keytoscript.js
@@ -1,0 +1,124 @@
+var bcrypto = require('./crypto')
+var btemplates = require('./templates')
+
+function checkAllowedP2sh (keyFactory) {
+  if (!(keyFactory instanceof P2pkhFactory ||
+      keyFactory instanceof P2wpkhFactory ||
+      keyFactory instanceof P2pkFactory
+    )) {
+    throw new Error('Unsupported script factory for P2SH')
+  }
+}
+
+function checkAllowedP2wsh (keyFactory) {
+  if (!(keyFactory instanceof P2pkhFactory ||
+      keyFactory instanceof P2pkFactory
+    )) {
+    throw new Error('Unsupported script factory for P2SH')
+  }
+}
+
+var P2pkFactory = function () {
+
+}
+
+/**
+ * @param {bitcoin.ECPair} key
+ */
+P2pkFactory.prototype.convert = function (key) {
+  return {
+    scriptPubKey: btemplates.pubKey.output.encode(key.getPublicKeyBuffer()),
+    signData: {}
+  }
+}
+
+var P2pkhFactory = function () {
+
+}
+
+/**
+ * @param {bitcoin.ECPair} key
+ */
+P2pkhFactory.prototype.convert = function (key) {
+  var hash160 = bcrypto.hash160(key.getPublicKeyBuffer())
+  return {
+    scriptPubKey: btemplates.pubKeyHash.output.encode(hash160),
+    signData: {}
+  }
+}
+
+var P2wpkhFactory = function () {
+
+}
+
+/**
+ * @param {bitcoin.ECPair} key
+ */
+P2wpkhFactory.prototype.convert = function (key) {
+  var hash160 = bcrypto.hash160(key.getPublicKeyBuffer())
+  return {
+    scriptPubKey: btemplates.witnessPubKeyHash.output.encode(hash160),
+    signData: {}
+  }
+}
+
+var P2shFactory = function (keyFactory) {
+  checkAllowedP2sh(keyFactory)
+  this.factory = keyFactory
+}
+
+P2shFactory.prototype.convert = function (key) {
+  var detail = this.factory.convert(key)
+  var hash160 = bcrypto.hash160(detail.scriptPubKey)
+  return {
+    scriptPubKey: btemplates.scriptHash.output.encode(hash160),
+    signData: {
+      redeemScript: detail.scriptPubKey
+    }
+  }
+}
+
+var P2wshFactory = function (keyFactory) {
+  checkAllowedP2wsh(keyFactory)
+  this.factory = keyFactory
+}
+
+P2wshFactory.prototype.convert = function (key) {
+  var detail = this.factory.convert(key)
+  var hash160 = bcrypto.hash160(detail.scriptPubKey)
+  return {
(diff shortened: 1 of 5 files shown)


src/keytoscript.js
```diff
@@ -0,0 +1,124 @@
+var bcrypto = require('./crypto')
+var btemplates = require('./templates')
+
+function checkAllowedP2sh (keyFactory) {
```
afk11: these two functions suck, doesn't really allow people to bring their own functions.. but wanted run-time prevention against nesting script hash types incorrectly.

src/keytoscript.js
```diff
@@ -0,0 +1,124 @@
...
+  if (!(keyFactory instanceof P2pkhFactory ||
+      keyFactory instanceof P2wpkhFactory ||
+      keyFactory instanceof P2pkFactory
+    )) {
+    throw new Error('Unsupported script factory for P2SH')
+  }
+}
+
+function checkAllowedP2wsh (keyFactory) {
```
afk11: see above

test/hdnode.js
```diff
@@ -124,12 +123,12 @@ describe('HDNode', function () {
...
-          .once().withArgs().returns('foobar')
-
-        assert.strictEqual(hd.getAddress(), 'foobar')
-      }))
+      // it('wraps keyPair.getAddress', setupTest(function () {
+      //   this.mock(keyPair).expects('getAddress')
+      //     .once().withArgs().returns('foobar')
+      //
+      //   assert.strictEqual(hd.getAddress(), 'foobar')
+      // }))
```
afk11: Should discuss whether keys have an associated address type. Maybe following keytoscript, we should just assume _some_ function is responsible for converting keys to their scripts, and then addresses, rather than having getAddress and trying to manage defaults.

afk11:
haven't tested all of the the key-to-script functions, will leave that after I get feedback on the approach being taken. the prefixes/script functions used in tests obviously work :)

dcousens:
Did you look to https://github.com/bitcoinjs/bitcoinjs-playground which should land in `4.0.0` for how these might integrate?
There is definitely a lot of double up :confused:

afk11:
Is it this PR needs refactoring first, or we need to wait until that gets merged?

dcousens:
> It confused me a little though as it has a lot of emphasis on inputs, but yea, all I need is address/scriptPubKey/rs/ws!

Can you elaborate on what you mean by this?

afk11:
> It confused me a little though as it has a lot of emphasis on inputs, but yea, all I need is address/scriptPubKey/rs/ws! 

Yea - with the emphasis on inputs, witnesses, and decoding all the information from the object, it looks less like it's meant to make the data I need. Signature parsing normally happens in a separate step to making the output script you'd request payment on.. 

Though, it means something like this might work:

hdnode.getScriptData -> (p2wsh(p2wpkh()) -> output scripts
hdnode.getScriptData(input) -> (p2wsh(p2wpkh()) -> output scripts + input signatures, etc

dcousens:
That aside,  I  really don't think tagging `keyPair`s or `HDNode` with script information is the right outcome.  It is unfortunate that `ypub`/`zpub` has done that,  but,  it really seems like 1 particular decision is trying to re-organize several code bases...
We wanted to *remove* `getAddress`,  not encode it further...

afk11:
Yea I don't like it on key-pairs much either. Electrum also proposes we start doing script-types with WIF's, which I think is awful, but at least they're the only ones doing it for now :P

The hate for getAddress was because we had no control over the script type, and couldn't change it without consumers being affected. This PR would actually fix that problem for a hdnode, surely?

I've had requests for this feature elsewhere as well, since two wallets (electrum & trezor) have implemented it, I don't see it going away.. 

So instead of scrapping it, we can try carve a useful feature out of it? As it stands, users generally have to plug primitives together (scripts/templates/crypto,etc) to get an address, so I think exposing redeem/witness/spk script generation, and thereby addresses, is a really rapid way to get people signing without worrying about internals (like script types, whether it's p2sh, p2wsh, and the address type!)

dcousens:
> , so I think exposing redeem/witness/spk script generation, and thereby addresses, is a really rapid way to get people generating addresses and signing without worrying about a tonne of internals (like script types, whether it's p2sh, p2wsh, and the address type!)

Isn't that exactly what the `bitcoinjs-playground` code provides?  A safe and complete abstraction for users to replace the clunky-ness of tag-teaming the right incantation for payment type?

afk11:
In this case, I see the useful feature just being being able to produce scriptPubKeys/redeemScripts/witnessScripts/addresses for the HD key.

Instead of shooting straight for pubkey -> address, we can go pubkey -> scripts -> address, and also make it a useful way to get the scripts for signing.

dcousens:
See [BIP178](https://github.com/bitcoin/bips/pull/673#pullrequestreview-111474458),  this is probably what we will do.

dcousens:
@afk11 maybe we should hit the discussion up with how this will be handled with extended keys?

dcousens:
@afk11 what are you're thoughts on this now with the `4.0.0` API?

PR #1025, opened 2018-03-22
Add getter for attribute `compressed` of `ECPair`

vsund:
This PR adds a `getCompressed()` getter for the `compressed` attribute from `ECPair`.
I tried to stick to the `getNetwork()` getter / tests. Not sure if the test is supposed to work this way.

Please let me know if anything is wrong, happy to change things.

---
Related: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/24453

### DIFF
--- a/src/ecpair.js
+++ b/src/ecpair.js
@@ -104,6 +104,10 @@ ECPair.prototype.getAddress = function () {
   return baddress.toBase58Check(bcrypto.hash160(this.getPublicKeyBuffer()), this.getNetwork().pubKeyHash)
 }
 
+ECPair.prototype.getCompressed = function () {
+  return this.compressed
+}
+
 ECPair.prototype.getNetwork = function () {
   return this.network
 }
--- a/test/ecpair.js
+++ b/test/ecpair.js
@@ -217,6 +217,17 @@ describe('ECPair', function () {
     })
   })
 
+  describe('getCompressed', function () {
+    fixtures.valid.forEach(function (f) {
+      it('returns ' + f.compressed + ' for ' + f.WIF, function () {
+        var compressed = f.compressed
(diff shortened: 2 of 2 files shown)


dcousens:
I'm not against this,  but the `.compressed` attribute is specifically public (compared to `__Q` etc).

IMHO,  it might be more consistent to remove `getNetwork()`,  however that has been controversial as `HDNode` wasn't exposing `.network`,  and so to keep the two consistent,  we kept `getNetwork`.

Thoughts?

vsund:
I'd have absolutely no problem with leaving it as it is. Just thought that it's more consistent to add a proper getter for `compressed` as there was one for `network`.

This was just the idea over in https://github.com/DefinitelyTyped/DefinitelyTyped/issues/24453#issuecomment-375169833 to stay consistent. Happy to proceed with using `compressed` (and adding this to the types).

youssefgh:
@dcousens thank you for your informations about `getNetwork()`
In this case i think that adding a readonly attribute only to TS would be simpler
Better adding a readonly attribute also for network or any any other missing attribute in ECPair. I leave the choice to you @vsund 😉

vsund:
Would agree that we leave this as it is, there's no need to re-open an already-made decision about `network` vs `getNetwork()` :)

So closing this PR and adding types for `compressed` and `network` (while keeping `getNetwork()`).
Ok @dcousens @youssefgh?

dcousens:
@vsund :+1: ,  I'll re-review if `getNetwork` is necessary in `4.0.0`

vsund:
Good to hear, closing this now and going on with https://github.com/DefinitelyTyped/DefinitelyTyped/issues/24453

PR #1032, opened 2018-04-06
add coinbase multisig tool to list of utilizing projects

prahaladbelavadi:
Ref: [Coinbase multisig tool](https://github.com/coinbase/multisig-tool)


dcousens:
Quite an old verison though,  but indeed

PR #1036, opened 2018-04-13
change TransactionBuilder to default to version 2

dcousens:
Unclear how 'breaking' this would be to be honest.

I opted to change it in `TransactionBuilder` rather than `Transaction`,  as this is the abstraction most users interact with unless they are dealing with the bare metal anyway.


dcousens:
Rather than re-do all signatures in the fixtures, I opted instead to simply add the version explicitly to the fixtures,   and then add a new test case that demonstrates the default.

PR #1039, opened 2018-04-13
decompile should return Maybe chunks

dcousens:
``` haskell
decompile :: Buffer -> Array
```

becomes

``` haskell
decompile :: Buffer -> Maybe Array
```

Additionally,  we will want a decode function that supports a partial decode and partial result.
Open to naming,  but it'll be added in this PR.

~~Tests failing due to some odd fixtures that were using completely invalid scripts.~~

~~**edit:** tagged no merge for now (until tests pass)~~


dcousens:
Ready to merge after travis passes

youssefgh:
@dcousens the CHANGELOG.md says :

> script.decompile now returns Buffer or null, if decompilation failed

I think it should be :

> script.decompile now returns Buffer Array or null, if decompilation failed

PR #1041, opened 2018-04-13
Change nulldata to return Array

dcousens:
Original PR https://github.com/bitcoinjs/bitcoinjs-lib/pull/859/files

This changes

``` haskell
nulldata.output.encode :: Buffer -> Buffer
nulldata.output.decode :: Buffer -> Buffer
```

to

``` haskell
nulldata.output.decode :: Buffer -> Array
nulldata.output.encode :: Array -> Buffer
```

While necesssary,  this is dangerous and may burn users who aren't doing any type checking.
I suggest we opt- for a different template name other than `nulldata`,  perhaps this is added to the new payments API instead as `P2NULL`?

Either way,  this PR can hold that discussion.

### DIFF
--- a/src/templates/nulldata.js
+++ b/src/templates/nulldata.js
@@ -14,15 +14,15 @@ function check (script) {
 check.toJSON = function () { return 'null data output' }
 
 function encode (data) {
-  typeforce(types.Buffer, data)
+  typeforce([types.Buffer], data)
 
-  return bscript.compile([OPS.OP_RETURN, data])
+  return bscript.compile([OPS.OP_RETURN].concat(data))
 }
 
 function decode (buffer) {
   typeforce(check, buffer)
 
-  return buffer.slice(2)
+  return bscript.decompile(buffer).slice(1)
 }
 
 module.exports = {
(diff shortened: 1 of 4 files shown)


src/templates/nulldata.js
```diff
@@ -14,15 +14,15 @@ function check (script) {
 check.toJSON = function () { return 'null data output' }
 
 function encode (data) {
-  typeforce(types.Buffer, data)
+  typeforce([types.Buffer], data)
```
dcousens: this would protect `encode` users

src/templates/nulldata.js
```diff
@@ -14,15 +14,15 @@ function check (script) {
...
 
-  return bscript.compile([OPS.OP_RETURN, data])
+  return bscript.compile([OPS.OP_RETURN].concat(data))
 }
 
 function decode (buffer) {
   typeforce(check, buffer)
 
-  return buffer.slice(2)
+  return bscript.decompile(buffer).slice(1)
```
dcousens: but this is very dangerous and will likely burn unsuspecting users who upgrade without reading the CHANGELOG closely

PR #1042, opened 2018-04-16
Add CONTRIBUTING.md

dcousens:
See changes

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,71 @@
+# Contributing to bitcoinjs-lib
+
+Firstly in terms of structure, there is no particular concept of "bitcoinjs developers" in the sense of privileged people.
+Open source often naturally revolves around meritocracy where longer term contributors gain more trust from the developer community.
+
+However, for practical purpose, there are repository "maintainers" who are responsible for merging pull requests.
+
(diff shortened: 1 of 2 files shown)


junderw:
We should add:

1. GPG sign your commits.
2. Or else the person reviewing will co-author and sign with their GPG key (this should be done right before merging, after the author rebases / squashes as instructed.) (https://help.github.com/articles/creating-a-commit-with-multiple-authors/) and (https://help.github.com/articles/signing-commits-using-gpg/) respectively.
3. Github signed commits (if you make commit through Web UI currently Github will sign your commit.) are fine.

dcousens:
(@junderw can you add that? I sourced https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md for inspiration at times)

dcousens:
We have branch protection in place... but it'd be nice if we could remove the ability to simply switch-it-off in the settings...

Any ideas?

junderw:
hmmmmmmmm.....

If we restrict master to PR merges only, the master HEAD should always be a merge commit...

I wonder if there's some sort of setting file (ie. .gitignore) that will allow for some sort of pre-hook for git push... not allowing push (locally) if the HEAD commit msg doesn't start with "Merge..." etc.

git is deep, so I wouldn't be surprised if there were such a thing.

PR #1043, opened 2018-04-21
add warning for generating addresses from sha256 hash example

prahaladbelavadi:
Potential fix for #1037 
Needs review from @dcousens


dcousens:
ACK,  but don't include the `package.lock.json`

dcousens:
Its a start,  but,  I'll leave https://github.com/bitcoinjs/bitcoinjs-lib/issues/1037 open,  as this explanation leaves some hand-holding to be desired.

Issue #1044, opened 2018-04-21
Proposal for standalone examples


prahaladbelavadi:
Examples references in the repository point to test cases which could potentially act as a barrier to entry to someone that isn't familiar with how testing works.

I've been manually trying to examine each of these functions in standalone js files [here](https://github.com/prahaladbelavadi/bitcoin-js-testing/tree/master/myTest/transaction-operations)

Examples logging outputs at certain places will help with understanding how the library works.

Would like to hear thoughts on preparing separate sample files for making it easier for people to adopt get started with using bitcoinjs library

afk11:
Somewhat agree! I'd say by now things that need tests are already tested.. we could possibly be a little more verbose in example scripts. could execute them during travis the same way we do tests, to make sure we haven't broken something.

dcousens:
> could execute them during travis the same way we do tests, to make sure we haven't broken something.

That was the biggest benefit of the integration approach.
Is there a simple dependency that will test the code used in a `README.md`?

prahaladbelavadi:
Maybe we could adopt a similar structure from [ExpressJS examples](https://github.com/expressjs/express/tree/master/examples) 

Or 

We could use [this](https://github.com/Widdershin/markdown-doctest) to test the example code if we'll be embedding sample code in markdown

Thoughts on this ? @dcousens @afk11

afk11:
I'd just run nodejs over every file in the directory and link to it from a markdown file. Often when I leave a code snippet in markdown, it goes out of date a week later :)

prahaladbelavadi:
lol that's true 
it makes sense
I'll add example files and make an index to reference them

dcousens:
Closing in faovur of #1051,  and if any other new PR comes,  I'll track it from there

PR #1051, opened 2018-04-24
Atomize examples to their own files

prahaladbelavadi:
Potential fix for #1044
Needs review from @afk11 and @dcousens


dcousens:
I guess the question is,  is this format any clearer than the integration tests as is?

prahaladbelavadi:
I think it would lower the barrier to entry as to those that would be interacting with this library.

I don't think people that are learning how to code would be able to absolve what's happening from the test cases. 
This would be similar to providing good developer resources/documentation for developers that are potentially going to be working on other applications on top of this library. 

By providing individual examples, they can try out their own stuff and see things functioning.

A personal motive for adding these examples is that being self taught, I was unable to grasp the context earlier from the test cases. 

I felt examples would've helped

dcousens:
I'm not against this.
Do we do it for every example?

Could the same result be achieved without using `mocha` but still having `assert`s or some way to verify the outputs against something?

We simply don't want the behaviour of the examples to change.
Maybe even checking as a script target that `$(node examples/* | diff - expected.log) && true` could be helpful here.

dcousens:
With the example links persistently becoming out of sync,  I think this is the only sane solution.

What if we put these examples ~~tests~~ under `/docs` instead?  (and still ran them as integration tests,  but made them distinctly separate)

junderw:
Do we still want to do this? I could take it over.

dcousens:
I think it's probably a good idea,  but emphasis on the examples being "verified"  is important IMHO.

junderw:
Closing stale PR. If you'd like to re-open please let me know.

PR #1062, opened 2018-05-08
ADD getDepth and getIndex getters

Fabsolute:


dcousens:
In this case,  the internal variables can be considered part of the public API.  No need for getter/setters.
I hope to clean this up some in 4.0.0.

dcousens:
See https://github.com/bitcoinjs/bitcoinjs-lib/issues/1069

Issue #1069, opened 2018-05-14
Prefer properties over getters (discuss)


dcousens:
Discussion requested.

@junderw @dabura667 

We typically sprinkle an unhealthy mix of `getNetwork` and `.network` throughout for many different fields.

Should we preference to read-only `Object.defineProperty`'s where possible or maintain `get___`?

`Transaction.prototype.getId` could become `tx.id` ...

Related:
https://github.com/bitcoinjs/bitcoinjs-lib/pull/1025
https://github.com/bitcoinjs/bitcoinjs-lib/pull/1062

junderw:
I prefer:

1. .network, .getNetwork(), .setNetwork() exist for every attribute
2. for internal attributes, .getNetwork() and .setNetwork() throws (I don't think you can make a read of .network throw in JS... can you?
3. for booleans, getNetwork() has an alias isNetwork() (booleans are usually "readable" etc... so it would be isReadable() which is a lot more.......... I'm sorry, unintentional dad-joke...... readable.

It's very java-esque, but I like it.

junderw:
We could also disallow touching .network, then we can put type checking using typeforce in the set___() functions.

junderw:
Object.defineProperty would be fine.

I just think the important part is consistency.

dcousens:
@junderw we don't maintain any of the invariants of our structures if users are modifying them.
We _very_ rarely provide `set__` except in a few rare cases where it does feel clunky.

> I just think the important part is consistency.

Agreed.

> I don't think you can make a read of .network throw in JS... can you?

Yes?

See https://github.com/bitcoinjs/bip32/commit/a037794e2189e55bd3f6850bfdd397034e9a26f7 for an example of how this might apply to `ECPair`...

jprichardson:
With the clean ES6 syntax, modern JavaScript code will move to getters / setters unlike the horrible verbose Java syntax (`getX`, `setX`). Let's use them and move on - especially since they've been with us since ES5.

junderw:
> Yes?

So if I do:

```javascript
let o = {}
// do something to add .network in a way that reading it would throw
let x = o.network
// This throws, while internal functions accessing this.network can read it perfectly fine.
```

What is that "something"?

you could (I guess) have logic in the getter to check the context in which it is being called?...

[...]

PR #1073, opened 2018-05-15
rm HDNode, use bip32 module

dcousens:


dcousens:
The exact version is still `<1.0.0` (note, `0.0.3` in this PR),  the exact API will match depending on https://github.com/bitcoinjs/bitcoinjs-lib/issues/1069.

As is, it uses a property based API.

PR #1076, opened 2018-05-18
Added ECPair.toUncheckedHex() and ECpair.fromUncheckedHex()

woodser:
This PR adds the ability to get the private key as unchecked hex (without network or compression information) and initialize a new ECPair from unchecked hex, network, and compression.

This is helpful to minimize the encoded data for encryption and splitting with Shamir's Secret Sharing which is why other libraries usually offer this in their API.

I'm using this API in the [CryptoStorage tool](https://cryptostorage.com/) [here](https://github.com/cryptostorage/cryptostorage.com/blob/aa9900a2361e9e8871a8b22cfccb6969fa011f66/js/CryptoPlugins.js#L585) and [here](https://github.com/cryptostorage/cryptostorage.com/blob/aa9900a2361e9e8871a8b22cfccb6969fa011f66/js/CryptoPlugins.js#L590).

### DIFF
--- a/src/ecpair.js
+++ b/src/ecpair.js
@@ -84,17 +84,6 @@ ECPair.fromWIF = function (string, network) {
   })
 }
 
-ECPair.fromUncheckedHex = function (hex, network, compressed) {
-	
-	// assume a network object (or default to bitcoin)
-	network = network || NETWORKS.bitcoin
-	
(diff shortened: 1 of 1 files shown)


woodser:
Close and recreate to separate PRs

PR #1070, opened 2018-05-14
rm ecdsa, switch out ECPair secp256k1 impl.

dcousens:
Related to https://github.com/bitcoinjs/bitcoinjs-lib/issues/1045, https://github.com/bitcoinjs/bitcoinjs-lib/issues/1049

PR #1077, opened 2018-05-18
Add ECPair.toUncheckedHex() and ECPair.fromUncheckedHex(hex, network, compressed)

woodser:
This PR adds the ability to get the private key as unchecked hex (without network or compression information) and initialize a new ECPair from unchecked hex, network, and compression.

This is helpful to minimize the encoded data for encryption and splitting with Shamir's Secret Sharing which is why other libraries usually offer this in their API.

I'm using this API in the [CryptoStorage tool](https://cryptostorage.com/) [here](https://github.com/cryptostorage/cryptostorage.com/blob/aa9900a2361e9e8871a8b22cfccb6969fa011f66/js/CryptoPlugins.js#L585) and [here](https://github.com/cryptostorage/cryptostorage.com/blob/aa9900a2361e9e8871a8b22cfccb6969fa011f66/js/CryptoPlugins.js#L590).

### DIFF
--- a/src/ecpair.js
+++ b/src/ecpair.js
@@ -90,9 +90,9 @@ ECPair.fromUncheckedHex = function (hex, network, compressed) {
 	network = network || NETWORKS.bitcoin
 	
 	return new ECPair(BigInteger.fromHex(hex), null, {
-		compressed: compressed,
-		network: network
-	})
+    compressed: compressed,
+    network: network
+  })
 }
 
 ECPair.makeRandom = function (options) {
--- a/src/ecpair.js
+++ b/src/ecpair.js
@@ -84,17 +84,6 @@ ECPair.fromWIF = function (string, network) {
   })
 }
 
-ECPair.fromUncheckedHex = function (hex, network, compressed) {
-	
-	// assume a network object (or default to bitcoin)
-	network = network || NETWORKS.bitcoin
-	
-	return new ECPair(BigInteger.fromHex(hex), null, {
-		compressed: compressed,
-		network: network
-	})
-}
-
 ECPair.makeRandom = function (options) {
   options = options || {}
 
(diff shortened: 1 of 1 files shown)


dcousens:
NACK,  already in https://github.com/bitcoinjs/bitcoinjs-lib/pull/1070

woodser:
The motivation of this PR is to encapsulate initialization from hex.  It looks like #1070 does not permit a caller to initialize an ECPair from a hex string without using BitcoinJS's internal Buffer class.  Please correct me if I'm wrong.  Otherwise I think initializing from unchecked hex should be added and I can make the PR off of #1070 if you agree?

dcousens:
@woodser `Buffer` isn't internal to bitcoinjs,  it is required for many things to interact with this library.

woodser:
I've been reading through #1070.  If I understand correctly, ECPair is going to accept a buffer as its first constructor argument, and the way to get the hexadecimal private key will be `keypair.privateKey.toString('hex')`, correct?

The hex will not be padded so it may contain fewer than 64 characters for some private keys, such as this LTC private key: T34MuTEaRxt3kj6dn2MMHbvSPQHrsBjKrdz6WtwrLe1KzCzYkMup.

Shouldn't it be padded to conform to 64 characters?

Also, is `d.toBuffer(32).toString('hex')` a valid way of creating a padded hex from a BigInteger?

Thanks.



It appears the way to create hexadecimal private keys in #1070 is `keyPair.privateKey.toString('hex')`.

The hex is not padded which can lead to private keys with fewer than 64 hex characters.

Would this be the preferred way to pad?

'keypair.d.toBuffer(32).toString('hex')`



Shouldn't it be padded?


From https://github.com/bitcoinjs/bitcoinjs-lib/pull/1070/files#diff-f4ef36df8339cfcb24466661f69ac75aR97 in #1070, it appears the way to get hexadecimal will become , correct?

It doesn't look like the hex output is padded.  Shouldn't it be since hexdecimal format is defined to be 64 characters

dcousens:
@woodser https://github.com/bitcoinjs/bitcoinjs-lib/blob/d7eb6c8e77c0d73bdd54a2b4dedfa1055d361c33/src/ecpair.js#L51-L57

Note specifically https://github.com/bitcoinjs/bitcoinjs-lib/blob/d7eb6c8e77c0d73bdd54a2b4dedfa1055d361c33/src/ecpair.js#L52

dcousens:
> Also, is d.toBuffer(32).toString('hex') a valid way of creating a padded hex from a BigInteger?

That is what we did previously :+1:

woodser:
Ok, thanks!

PR #1078, opened 2018-05-18
fromBase58Check should use network information for payload length 

woodser:
This PR fixes bugs caused by using BitcoinJS with altcoins like Zcash.

### DIFF
--- a/src/address.js
+++ b/src/address.js
@@ -13,7 +13,7 @@ function fromBase58Check (address, network) {
   
   // TODO: 4.0.0, move to "toOutputScript"
   if (payload.length < 21) throw new TypeError(address + ' is too short')
-  var maxLength = network.pubKeyHash < 256 ? 21 : 22;
+  let maxLength = types.UInt8(network.pubKeyHash) ? 21 : 22;
   if (payload.length > maxLength) throw new TypeError(address + ' is too long')
   
   var version = payload.readUInt8(0)
--- a/src/address.js
+++ b/src/address.js
@@ -7,13 +7,15 @@ var networks = require('./networks')
 var typeforce = require('typeforce')
 var types = require('./types')
 
-function fromBase58Check (address) {
-  var payload = bs58check.decode(address)
-
+function fromBase58Check (address, network) {
+  network = network || networks.bitcoin;
+  var payload = bs58check.decode(address, network.pubKeyHash)
+  
   // TODO: 4.0.0, move to "toOutputScript"
   if (payload.length < 21) throw new TypeError(address + ' is too short')
-  if (payload.length > 22) throw new TypeError(address + ' is too long')
-
+  var maxLength = network.pubKeyHash < 256 ? 21 : 22;
+  if (payload.length > maxLength) throw new TypeError(address + ' is too long')
+  
   var version = payload.readUInt8(0)
   var hash = payload.slice(1)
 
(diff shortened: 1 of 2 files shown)


src/ecpair.js
```diff
@@ -84,6 +84,17 @@ ECPair.fromWIF = function (string, network) {
...
 
+ECPair.fromUncheckedHex = function (hex, network, compressed) {
+	
+	// assume a network object (or default to bitcoin)
+	network = network || NETWORKS.bitcoin
+	
+	return new ECPair(BigInteger.fromHex(hex), null, {
+		compressed: compressed,
+		network: network
+	})
```
afk11: Uses tabs, should use spaces instead

woodser: I've updated the pull request to use spaces instead of tabs.

src/address.js
```diff
@@ -12,7 +12,7 @@ function fromBase58Check (address) {
 
   // TODO: 4.0.0, move to "toOutputScript"
   if (payload.length < 21) throw new TypeError(address + ' is too short')
-  if (payload.length > 21) throw new TypeError(address + ' is too long')
+  if (payload.length > 22) throw new TypeError(address + ' is too long')
```
dcousens: This isn't valid for Bitcoin though.

Perhaps a `network` parameter could be passed to `fromBase58Check` that provides the length paramters?

src/address.js
```diff
@@ -32,13 +32,24 @@ function fromBech32 (address) {
 }
 
 function toBase58Check (hash, version) {
```
dcousens: And here,  again,  `version` would need to be replaced with `network`

woodser: https://github.com/bitcoinjs/bitcoinjs-lib/blob/080675a5cc0a3be4c18a2d07be48f00bbf981d1e/src/address.js#L54 and https://github.com/bitcoinjs/bitcoinjs-lib/blob/080675a5cc0a3be4c18a2d07be48f00bbf981d1e/src/address.js#L55 use output script context to determine the version as network.pubKeyHash or network.scriptHash.  If the network is passed as a parameter to `toBase58Check(hash, network)` instead of version then it will also need that context to determine which is the version to use.  Seems like if all toBase58Check() needs is a version then maybe it should continue to take a version instead of the network?

woodser: I added the network parameter to `toBase58Check()`.

src/address.js
```diff
@@ -7,13 +7,15 @@ var networks = require('./networks')
...
+function fromBase58Check (address, network) {
+  network = network || networks.bitcoin;
+  var payload = bs58check.decode(address, network.pubKeyHash)
+  
   // TODO: 4.0.0, move to "toOutputScript"
   if (payload.length < 21) throw new TypeError(address + ' is too short')
-  if (payload.length > 22) throw new TypeError(address + ' is too long')
-
+  var maxLength = network.pubKeyHash < 256 ? 21 : 22;
+  if (payload.length > maxLength) throw new TypeError(address + ' is too long')
```
dcousens: Nice,  though I'd use `0xff` to be explicit about the significance of `256`.  
You can also do `let maxLength = types.UInt8(network.pubKeyHash) ? 21 : 22`

woodser: Updated per your suggestion.

dcousens:
NACK,  with the way we are headed in `4.0.0`,  this simply doens't really make sense any more.

Issue #1017, opened 2018-03-09
Merge in bitcoinjs-playground payments API


dcousens:

dcousens:
@junderw as I use this more,  and in its lazy-evaluated form,  I think this may simply replace `bitcoin.script.[templates]` and `bitcoin.address` in its entirety.

Can we have a chat this week to walk through the APIs before I [hopefully] get the PR up here?

junderw:
@dcousens sure. Hit me up on Slack.

dcousens:
Added breaking change label,  as this will likely replace all of the `bitcoin.script.[templates]` code when merged.

dcousens:
Merged https://github.com/bitcoinjs/bitcoinjs-lib/pull/1041 to simplify test transitions for this PR

dcousens:
See https://github.com/bitcoinjs/bitcoinjs-lib/pull/1096

PR #1085, opened 2018-05-22
rm getAddress

dcousens:

### DIFF
--- a/src/ecpair.js
+++ b/src/ecpair.js
@@ -1,16 +1,14 @@
-var baddress = require('./address')
-var bcrypto = require('./crypto')
-var ecdsa = require('./ecdsa')
-var randomBytes = require('randombytes')
-var typeforce = require('typeforce')
-var types = require('./types')
-var wif = require('wif')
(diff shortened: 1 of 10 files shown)


dcousens:
Part of https://github.com/bitcoinjs/bitcoinjs-lib/issues/1017

dcousens:
@junderw good to merge,  removal of `getAddress` only for now,  payments PR will be separate

junderw:
Could you change requires to use const instead of let?

regex replace `let(\s+\S+\s+=\s+require\('[^']+'\))` with `const$1`

dcousens:
@junderw I intend to do that in a mass PR after this stage.  For now,  I'm changing to `let` on anything I touch.

takahser:
how can we get testnet addresses after this merge? when using `payments.p2pkh`, a `p2pkh` address is returned, rather than a testnet address (starting with `m` or `n`)

junderw:
@takahser pass the testnet network.

PR #1086, opened 2018-05-22
Move Transaction to ES6 class syntax

junderw:
Edit: Decided to only do the syntax change here, getters and setters etc. should be in a different PR.

Two goals:

1. ES6 usage all throughout Transaction. Including class constructor.
2. change getHash, getId etc. to tx.hash tx.id etc.

One question:

block.js L153 typeforce: What should I change it to?

### DIFF
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -7,13 +7,13 @@ const typeforce = require('typeforce')
 const types = require('./types')
 const varuint = require('varuint-bitcoin')
 
-const varSliceSize = (someScript) => {
+function varSliceSize (someScript) {
   const length = someScript.length
 
   return varuint.encodingLength(length) + length
 }
 
-const vectorSize = (someVector) => {
+function vectorSize (someVector) {
   const length = someVector.length
 
   return varuint.encodingLength(length) + someVector.reduce((sum, witness) => {
@@ -241,24 +241,24 @@ class Transaction {
 
     let tbuffer, toffset
 
-    const writeSlice = (slice) => {
+    function writeSlice (slice) {
       toffset += slice.copy(tbuffer, toffset)
     }
 
-    const writeUInt32 = (i) => {
+    function writeUInt32 (i) {
       toffset = tbuffer.writeUInt32LE(i, toffset)
     }
 
-    const writeUInt64 = (i) => {
+    function writeUInt64 (i) {
       toffset = bufferutils.writeUInt64LE(tbuffer, i, toffset)
     }
 
-    const writeVarInt = (i) => {
+    function writeVarInt (i) {
       varuint.encode(i, tbuffer, toffset)
       toffset += varuint.encode.bytes
     }
 
-    const writeVarSlice = (slice) => {
+    function writeVarSlice (slice) {
       writeVarInt(slice.length)
       writeSlice(slice)
     }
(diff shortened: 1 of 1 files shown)


dcousens:
> block.js no L153 no typeforce: What should I change it to?

I thought issues like this might prop,  esp. around `getHash` etc.

https://github.com/bitcoinjs/bitcoinjs-lib/blob/0d76f00917d8654377b4434ef94762fcef2402c1/src/block.js#L152-L161

`(t) => t.hash` would simply replace the inner-loop.  If we type-check it,  we double-down the on the cost.
It would have to either be a constant structure that we can lazily evaluate to be sane on repeated `.hash` access (like we do in `ECPair`),  or we should probably leave `getHash` as is.

`.id` falls to the same thing.

A constant structure approach would break _a lot_.

dcousens:
d85c4c9 ... so many dirty flags :cry:

dcousens:
@junderw we can still do this,  but you need to drop the

> change getHash, getId etc. to tx.hash tx.id etc.

changes.

junderw:
Hmm, to be honest, rebasing and re-thinking is kind of hairy, so I am thinking of just closing this and starting all over with a different approach maybe...?

dcousens:
OK,  if you think it'd be easier.

junderw:
Decided to only do Syntax for now, much easier. unit tests and standard tests passed locally.

src/transaction.js
```diff
@@ -7,35 +7,20 @@ const typeforce = require('typeforce')
 const types = require('./types')
 const varuint = require('varuint-bitcoin')
 
-function varSliceSize (someScript) {
+const varSliceSize = (someScript) => {
   const length = someScript.length
```
dcousens: why?

src/transaction.js
```diff
@@ -7,35 +7,20 @@ const typeforce = require('typeforce')
...
 
   return varuint.encodingLength(length) + length
 }
 
-function vectorSize (someVector) {
+const vectorSize = (someVector) => {
   const length = someVector.length
 
-  return varuint.encodingLength(length) + someVector.reduce(function (sum, witness) {
```
dcousens: OK, I do prefer the shorter style here,  but,  only as a matter of consistency

src/transaction.js
```diff
@@ -7,35 +7,20 @@ const typeforce = require('typeforce')
...
 const varuint = require('varuint-bitcoin')
 
-function varSliceSize (someScript) {
+const varSliceSize = (someScript) => {
   const length = someScript.length
 
   return varuint.encodingLength(length) + length
 }
 
-function vectorSize (someVector) {
```
dcousens: this seems silly,  searching for `function vectorSize` is immensely useful compared to the above syntax, which, although constant, makes me think of this `function` as somehow variable...

src/transaction.js
```diff
@@ -46,447 +31,525 @@ const BLANK_OUTPUT = {
...
+    this.ins = []
+    this.outs = []
   }
 
-  function readUInt32 () {
-    const i = buffer.readUInt32LE(offset)
-    offset += 4
-    return i
+  static get DEFAULT_SEQUENCE () {
+    return 0xffffffff
```
dcousens: is this _the_ way to do constants?

src/transaction.js
junderw: Personal preference. hoisted functions make me feel dirty inside nowadays... I'm not religious about it tho.

src/transaction.js
junderw: I love regex, so I'd just search `const xxxxx = \S+ => ` where xxxxx is the function I'm looking for... in fact that query is in my favorites actually...

Again, preference, and I'm not too set on it.

src/transaction.js
junderw: This is the best way I found. Open to other suggestions. (I'm guessing you want one constant per line?)

junderw:
You can add a commit tweaking the preference stuff if you'd like, I'm not too set on it either way.

src/transaction.js
dcousens: for now, lets leave this as it was,  and then we can discuss the new style separately?

junderw:
Reverted

dcousens:
LGTM, shame the diff is a mess.

I wonder if we can force it to 'align'?

junderw:
idk lol... refactors are usually messy...

junderw:
Forgot to merge this lol

dcousens:
Can we revert this and merge it into an `es6` branch,  then we can retarget each PR to that branch

PR #1088, opened 2018-05-24
add litecoin testnet support

mmsqe:
noted that litecoin uses SCRIPT_ADDRESS2 for scriptHash based on [params]
[params]: https://github.com/litecoin-project/litecoin/blob/master/src/base58.cpp#L290

### DIFF
--- a/src/networks.js
+++ b/src/networks.js
@@ -33,5 +33,16 @@ module.exports = {
     pubKeyHash: 0x30,
     scriptHash: 0x32,
     wif: 0xb0
-  }
+  },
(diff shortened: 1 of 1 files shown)


dcousens:
We should probably remove `litecoin` entirely.

mmsqe:
noted, thx

fanatid:
@mmsqe you can use [coininfo](https://github.com/cryptocoinjs/coininfo) for bitcoinjs-lib network parameters

Daniel, maybe you will `litecoin` for 4.0.0?

PR #1090, opened 2018-05-29
fix standard 11 issues

dcousens:
As follow up to b2d34c5 (which I pushed mistakenly... branch protection fail).
@junderw if only we could make this an autonomous organizations that didn't allow admins to push without review`

### DIFF
--- a/src/script_number.js
+++ b/src/script_number.js
@@ -35,11 +35,11 @@ function decode (buffer, maxLength, minimal) {
 
 function scriptNumSize (i) {
   return i > 0x7fffffff ? 5
-  : i > 0x7fffff ? 4
-  : i > 0x7fff ? 3
-  : i > 0x7f ? 2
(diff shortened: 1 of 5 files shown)


dcousens:
Good to merge after ACK

src/script_number.js
```diff
@@ -35,11 +35,11 @@ function decode (buffer, maxLength, minimal) {
...
-  : i > 0x7fffff ? 4
-  : i > 0x7fff ? 3
-  : i > 0x7f ? 2
-  : i > 0x00 ? 1
-  : 0
+    : i > 0x7fffff ? 4
+      : i > 0x7fff ? 3
+        : i > 0x7f ? 2
+          : i > 0x00 ? 1
+            : 0
```
dcousens: @feross was this intentional in standard 11?

feross: This was not intentional.

PR #1091, opened 2018-05-30
force update MTP for when tests not recently run

dcousens:
Median time past was unchanging unless several blocks were found recently.

This was failing tests that rely on MTP (CLTV, CSV) as their expiries were in the past (-3 hours),  but not _actually in the past in terms of MTP_.


dcousens:
Noticed tests were failing in https://github.com/bitcoinjs/bitcoinjs-lib/pull/1085,  this PR fixes that

PR #1094, opened 2018-06-05
script isCanonical* refactor/renames

dcousens:
Misleading otherwise


dcousens:
Alternatively,  we could probably remove `isCanonicalPubkey` in favour of users using the ECC library?

dcousens:
@junderw post-review ACK appreciated

PR #1111, opened 2018-06-25
s/var/let and s/var/const

dcousens:


dcousens:
As intended,  you can now assume any instance of `(^let )|(  let )` is a mutating variable.

:fireworks:

dcousens:
Ready to merge on 1 ACK

junderw:
@dcousens regtest server TLS cert expired :-P

`Error: certificate has expired`

PR #1113, opened 2018-06-26
fix getAddress function in integration test

ak1103dev:


dcousens:
@ak1103dev this has changed in master. :smile:

PR #1114, opened 2018-06-27
payments: dont always require output data

dcousens:
Previously we upheld that there was `enough data` for atleast `.output`,  however,  in many use cases,  you might only have the data [and only want] for `.input`.

I started loosening this in https://github.com/bitcoinjs/bitcoinjs-lib/commit/8197f9d7040671375f250f7656974a1ee28709e7,  but this follows through a bit more for P2SH/P2WSH.


dcousens:
Additionally, it uses the new payments API for integration tests

PR #1132, opened 2018-07-17
Remove Node 6, update to Node LTS

dcousens:
Restricts the library to Node 8 and above.

Thoughts? Queries? Qualms?

Does this go in `4.0.0`? (https://github.com/bitcoinjs/bitcoinjs-lib/pull/1131)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,14 @@
 sudo: false
 language: node_js
 node_js:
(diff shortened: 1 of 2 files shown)


junderw:
ACK 0bdf6be

Reference:
https://github.com/nodejs/Release

April 2019 will come sooner than we think.

PR #1135, opened 2018-07-20
Add additional payments integration tests

dcousens:


test/integration/payments.js
```diff
@@ -0,0 +1,153 @@
...
+          })
+        } else if (depends.signature) {
+          txb.sign(0, keyPairs[0], p2sh.redeem.output, null, unspent.value)
+        }
+
+        regtestUtils.broadcast(txb.build().toHex(), done)
+      })
+    })
+
+    it.only('can (wrapped in P2WSH) broadcast as an output, and be spent as an input', (done) => {
```
dcousens: Something blows up on `P2WSH(P2WPKH)` ... is that a problematic combo?

test/integration/payments.js
```diff
@@ -0,0 +1,157 @@
...
+        } else if (depends.signature) {
+          txb.sign(0, keyPairs[0], p2sh.redeem.output, null, unspent.value)
+        }
+
+        regtestUtils.broadcast(txb.build().toHex(), done)
+      })
+    })
+
+    it('can (wrapped in P2WSH) broadcast as an output, and be spent as an input', (done) => {
+      if (k === 'p2wpkh') return done() // skip P2WSH(P2WPKH)
```
dcousens: is `P2WSH(P2WPKH)` possible?
Are we simply lacking support? (it fails)

test/integration/payments.js
```diff
@@ -0,0 +1,157 @@
...
+        } else if (depends.signature) {
+          txb.sign(0, keyPairs[0], null, null, unspent.value, p2wsh.redeem.output)
+        }
+
+        regtestUtils.broadcast(txb.build().toHex(), done)
+      })
+    })
+
+    it('can (wrapped in P2WSH, then wrapped in P2SH) broadcast as an output, and be spent as an input', (done) => {
+      if (k === 'p2wpkh') return done() // skip P2SH(P2WSH(P2WPKH))
```
dcousens: same here, is `P2SH(P2WSH(P2WPKH))` possible?
Are we simply lacking support? (it fails)

dcousens:
I'm happy to release after this passes

test/integration/payments.js
```diff
@@ -0,0 +1,71 @@
...
+      buildAndSign(depends, p2wsh.output, null, p2wsh.redeem.output, done)
+    })
+
+    it('can (as P2SH(P2WSH(' + k + '))) broadcast as an output, and be spent as an input', (done) => {
+      if (k === 'p2wpkh') return done() // skip P2SH(P2WSH(P2WPKH))
+
+      const p2wsh = bitcoin.payments.p2wsh({ redeem: { output }, network: NETWORK })
+      const p2sh = bitcoin.payments.p2sh({ redeem: { output: p2wsh.output }, network: NETWORK })
+
+      buildAndSign(depends, p2sh.output, p2sh.redeem.output, p2wsh.redeem.output, done)
```
dcousens: that this works so simply... is awesome

dcousens:
Waiting to see if `P2WSH(P2WPKH)` should be possible... if it is, I'll fix the tests, if it isn't, I'll enforce that `TransactionBuilder` refuses to sign it.

src/transaction_builder.js
```diff
@@ -232,6 +232,11 @@ function prepareInput (input, ourPubKey, redeemScript, witnessValue, witnessScri
       expanded.signatures = input.signatures
     }
 
+    let signScript = witnessScript
+    if (expanded.type === SCRIPT_TYPES.P2WPKH) {
+      signScript = payments.p2pkh({ pubkey: expanded.pubkeys[0] }).output
+    }
```
dcousens: I thought this would resolve the signing issues for `P2WSH(P2WPKH)`... but, it still fails.  Investigating.

dcousens:
ok,  @sipa has mentioned that `P2WSH(P2WPKH)` would fail,  so,  we need to prevent `TransactionBuilder` from even trying (or supporting users in their attempt!)

dcousens:
Any thoughts about whether we should try and block it via the payments scripts?
Can be done post `4.0.0` though.

PR #1139, opened 2018-07-23
4.0.1 | bump dependencies to 1.0.0 versions (and simplify Travis)

dcousens:


dcousens:
Added 94f0ae8 as the previous version of `tiny-secp256k1` did have `ecurve`,  which it was happily using (by mistake)

PR #1149, opened 2018-07-26
Add in missing Classify export (fixes #1148)

### LINKED ISSUE #1148: Classify is not Exported after Templates Export was removed
Hello!

Inside of my library I was using the `templates` classes that were being exported/attached onto the `scripts` export, however in a recent commit, `templates` stopped being exported in favor of moving to a new `classify` class (https://github.com/bitcoinjs/bitcoinjs-lib/commit/6cacea6f31aa18cce61bed562d730c55ac3709d3#diff-1fdf421c05c1140f6d71444ea2b27638L2).

It would be great if `classify` could be exported so that I could access/use the new `classify.output` and `classify.input` functions.

I will make a pull request that fixes this issue by adding back in the export that was removed.

Thanks!
Sky Young

OstlerDev:
Hello!

In a recent commit, the `templates` class was refactored into the `classify` class, and in the process, the `templates` class was removed as being an export (previously attached under the `scripts` export). This issue is documented at #1148.

This pull request fixes the issue of the removed export by adding the `classify` class as an export, to allow (similar) access to the `templates` export that was removed.

Please let me know if any changes are required, I can happily make them.

Thanks!
Sky Young

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -12,5 +12,6 @@ module.exports = {
   networks: require('./networks'),
   opcodes: require('bitcoin-ops'),
   payments: require('./payments'),
+  classify: require('./classify'),
   script: script
 }


dcousens:
Please see explanation in https://github.com/bitcoinjs/bitcoinjs-lib/issues/1148

PR #1151, opened 2018-07-27
Enforce Transaction has outputs

dcousens:


dcousens:
Probably should be enforced if using `build()` too

dcousens:
Simplified to throw in the SIGHASH_NONE flow for now.

test/transaction_builder.js
```diff
@@ -340,6 +340,7 @@ describe('TransactionBuilder', function () {
       it('throws ' + f.exception + (f.description ? ' (' + f.description + ')' : ''), function () {
         const txb = construct(f, true)
 
+        let threw = false
```
dcousens: if nothing throws, it isn't an invalid test is it!

dcousens:
Opened https://github.com/bitcoinjs/bitcoinjs-lib/issues/1230 for the remaining solo `SIGHASH_ALL` which should fail if the number of `outputs.length === 0`... the reason for this PR!

dcousens:
Good to merge, if tests pass

PR #1181, opened 2018-08-21
default redeemScript to what is previously provided

dcousens:
Is this a regression? Or did we decide against this somewhere?

### DIFF
--- a/src/transaction_builder.js
+++ b/src/transaction_builder.js
@@ -216,6 +216,8 @@ function expandOutput (script, ourPubKey) {
 }
 
 function prepareInput (input, ourPubKey, redeemScript, witnessValue, witnessScript) {
+  redeemScript = redeemScript || input.redeemScript
+
   if (redeemScript && witnessScript) {
     const p2wsh = payments.p2wsh({ redeem: { output: witnessScript } })
     const p2wshAlt = payments.p2wsh({ output: redeemScript })
--- a/test/fixtures/transaction_builder.json
+++ b/test/fixtures/transaction_builder.json
@@ -1561,8 +1561,7 @@
             "scriptSigAfter": "OP_0 OP_0 OP_0 30440220793d87f2a8afeb856816efa38984418c692c15170e99ca371f547454079c0dd3022074ae95e438fee1f37619fabe0ce1083c3be0d65c3defb5337833d50fdc694b1301 52210258db1bb3801f1ecde47602143beaeb9cac93251724b8e589fae5c08c1a399a9121038e803e3d84cfc821cc8bf46233a9c2bb359d529db0bcdd3f1a4f38678dd02d7f2103b83e59d848407d7f62a82c99905f5ca3e8e8f5d6400eb78a0b4b067aea0720d953ae",
             "signs": [
               {
-                "keyPair": "cTkcnMZoFYH1UgumzCFHv2veLMNN1PaJyHHUxFT127zhNGBqqEZ2",
-                "redeemScript": "OP_2 0258db1bb3801f1ecde47602143beaeb9cac93251724b8e589fae5c08c1a399a91 038e803e3d84cfc821cc8bf46233a9c2bb359d529db0bcdd3f1a4f38678dd02d7f 03b83e59d848407d7f62a82c99905f5ca3e8e8f5d6400eb78a0b4b067aea0720d9 OP_3 OP_CHECKMULTISIG"
+                "keyPair": "cTkcnMZoFYH1UgumzCFHv2veLMNN1PaJyHHUxFT127zhNGBqqEZ2"
               }
             ]
           }


src/transaction_builder.js
```diff
@@ -216,6 +216,8 @@ function expandOutput (script, ourPubKey) {
 }
 
 function prepareInput (input, ourPubKey, redeemScript, witnessValue, witnessScript) {
+  redeemScript = redeemScript || input.redeemScript
```
dcousens: Not much to it,  but, did we decide to *enforce* users providing the redeem script rather than signing it without them knowing?

test/fixtures/transaction_builder.json
```diff
@@ -1547,6 +1547,27 @@
         ]
       }
     ],
+    "fromTransactionSequential": [
+      {
+        "description": "Transaction w/ P2SH(P2MS 2/3) -> ?",
+        "network": "testnet",
+        "txHex": "0100000001b033b2214568b49fda417371aba0634b0303a2b6a19884c25d03d0b91bdbe231000000006f000000004c6952210258db1bb3801f1ecde47602143beaeb9cac93251724b8e589fae5c08c1a399a9121038e803e3d84cfc821cc8bf46233a9c2bb359d529db0bcdd3f1a4f38678dd02d7f2103b83e59d848407d7f62a82c99905f5ca3e8e8f5d6400eb78a0b4b067aea0720d953aeffffffff0200e1f5050000000017a914a9974100aeee974a20cda9a2f545704a0ab54fdc87c72831010000000017a9149f57a6712ef023f85ffac631ed4263b977b2d0678700000000",
```
dcousens: @omrip30 I assume it is OK for me to add this (your transaction) as a test fixture?

omrip30: No problem, I created this especially for this example :)

junderw:
If we do this, the security check of input.redeemScript === redeemScript is useless...

If you assume signer A signed a certain input with a certain redeemScript, but it turned out they tried to trick you into signing another input, there are only a few ways we can catch that... one way is to explicitly sign the utxo amount, which is great that segwit added it, but not all utxos are segwit.

So if I made two P2SH addresses, one that is a 2 of 3 with myself, A, and B... and one that is a 4 of 6 with myself, Z, Y, X, W, A... and we had some rules for which funds are used for which...

I should be required to explicitly say "this is the redeemScript I am signing for" when signing... I should not say "ahhh whatever is in that transaction should be cool, I trust A."

dcousens:
That was my concern. 
We already provide tooling to extract these scripts quite simply.   The point is that they need to be matching it up against something pre existing, retained from when it was created.

junderw:
> If it was witness, you could specify the value and sign without concern?

Not really "without" concern, but "with less" concern... since there is still a small chance that two different UTXOs I want to manage happen to have the same amounts.

omrip30:
I basically agree with what you said but, I still need access to the inputs (to access the redeem script).
Otherwise, I won't be able to pass them to the 'sign' method (again, I use a multisig scheme) For some reason,  both input arrays are undefined in 'TransactionBuilder' and 'Transaction' after I create a transaction from an existing hex...

junderw:
> (again, I use a multisig scheme where each user receives a templated transaction to sign)

You can give the user a JSON string that is something like:

```
// For a 5 input tx.
{"inputRedeemScripts": ["xxxx","yyyyy","zzzzz","1111","2222"], "rawtx": "aaaaaaa" }
```

omrip30:
Possible, I'll think about it. Thanks

afk11:
I think this might be OK if the prevOutScript has been provided already, allowing us to check the redeemScript. I suspect if it always worked, people would appreciate the convenience and not consider the issue any further.

@omrip30 - @junderw raises a good point about saving redeemScripts elsewhere.. Is there some limitation in your app that prevents this? Thinking about crafting the unsigned transaction for the first time, one only adds outpoints funded by a scriptPubKey owned by the participants, and that's derived from the redeemScript.. Why wouldn't they be aware of the redeemScript if they somehow had the address to pay to in the first place?

junderw:
Idea: If you include the xpubs of all participants in each piece of data, your app can derive the keys to generate redeemScript on-site... and you can sort+concat the strings, take a hash, and convert the hash to a short phrase via BIP39 etc... and have your signers remember the phrase... that way they know if the xpubs have been tampered with.

however, this involves carrying around all xpubs on some disk that might not be secure... so it might be best to pre-generate the redeemscripts and add them with unsigned data.

dcousens:
@junderw or you could checksum the redeemscript,  with the length of the checksum matching your trust model?

omrip30:
@afk11  Sounds reasonable, I'll think about it :)

junderw:
Yes. Let's go ahead and close this.

PR #1191, opened 2018-08-24
Update networks.js

Amirh24:
Add dogecoin network to networks.js


dcousens:
Hi @Amirh24 ,  we aren't accepting alternate networks at this time.
You can provide these network details yourself if you decide the library is suitable to use with Dogecoin.

PR #1192, opened 2018-08-24
Add lowR grinding

junderw:
Requires https://github.com/bitcoinjs/tiny-secp256k1/pull/23
Once it is merged and published, I'll bump the version of the dependency here.

- [ ] Decide whether we actually want to allow this yet.


junderw:
Memo: this also needs to be implemented for bip32 library as well. (It uses an independent sign function)

junderw:
Closed in favor of #1385

Issue #1194, opened 2018-08-27
Problem creating transaction from Segwit multisig input


coinflippa:
I followed this example to create a multisig bech32 address (my address is 2/3 though).

https://github.com/bitcoinjs/bitcoinjs-lib/blob/20551fd38070cf68d078c0644c7586a2607f57fa/test/integration/addresses.js#L76-L88

I spent to the address and then I am attempting to create a transaction, mostly based on this example: https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/test/integration/transactions.js#L223-L260

```
const network = bitcoin.networks.testnet

const uxtos = [
  {
    "txid": "...",
    "vout": 1
  }
]

const destinations = [
  {
    "address": "...",
    "amount": "..."
  }
]

const publicKeys = [
  "03421bdf5825c26fa96fe1402a49df9901f31973dc9fab189d9ff4d8af853d0c4c",
  "022b1e2fcfe441df558097e7300e4c23a5f459f3ea45c8828962c589fc88d43543",
  "03749ba41055a85371460a7b01c5b74eab55af7c0de7c490f981cdccd5c2f0f286"
  ]

const pubkeys = publicKeys.map((pubkey) => new Buffer(pubkey, 'hex'))
const p2ms = bitcoin.payments.p2ms({ m: 2, pubkeys, network })
const p2wsh = bitcoin.payments.p2wsh({ redeem: p2ms, network })
const p2sh = bitcoin.payments.p2sh({ redeem: p2wsh, network })

const wifStrings = [
  "cUfMQu8f97JsSCYMUtr3sNpKSqhFifdaoy5GAfQv94atAZH2Sg7e",
  "cW3nPhsqEhCztNF3AqGrKpNaF...",
  "cV2tBrP4VSni3nCYzAdZDfX7K..."
]

const wifs = wifStrings.map((wif) => bitcoin.ECPair.fromWIF(wif, network))
const txb = new bitcoin.TransactionBuilder(network)

utxos.map((utxo) => {
  txb.addInput(utxo.txid, utxo.vout, null, p2sh.output)
})

destinations.map((destination) => {
  txb.addOutput(destination.address, destination.amount)
})

txb.__inputs.forEach((_ele, index) => {
  wifs.map((wif) => {
    txb.sign(index, wif, p2sh.redeem.output, null, req.body.total, p2wsh.redeem.output)
   })
})

const incomplete = txb.build().toHex()
```

[...]

PR #1195, opened 2018-08-27
Fix P2MS not checking during build

junderw:
Related: #1194

### DIFF
--- a/src/transaction_builder.js
+++ b/src/transaction_builder.js
@@ -399,7 +399,7 @@ function build (type, input, allowIncomplete) {
         signatures = signatures.filter(x => x)
       }
 
-      return payments.p2ms({ signatures }, { allowIncomplete })
+      return payments.p2ms({ signatures }, { allowIncomplete, validate: true })
     }
     case SCRIPT_TYPES.P2SH: {
       const redeem = build(input.redeemScriptType, input, allowIncomplete)


test/fixtures/p2pk.json
```diff
@@ -5,6 +5,7 @@
       "arguments": {
         "output": "030000000000000000000000000000000000000000000000000000000000000001 OP_CHECKSIG"
       },
+      "options": {},
       "expected": {
```
dcousens: did you merely add this to every vector?

junderw: Yes.

junderw:
I could. Would you like me to?

Any preference as to which one?

dcousens:
99% undefined,  as that is the default for most.  Then explicitly test the `{}` and other possible options.

test/transaction_builder.js
```diff
@@ -477,7 +477,7 @@ describe('TransactionBuilder', function () {
                   redeem: payments.p2ms({
                     output: redeemScript,
                     signatures
-                  }, { allowIncomplete: true })
+                  }, { allowIncomplete: true, validate: false })
```
dcousens: wait, shouldn't `allowIncomplete` allow for `validate: true`,  but in an acceptable way?
They aren't exclusive?

Issue #404, opened 2015-04-29
Redundant blank signatures are added when signing incomplete multisig transaction


nixoid:
My goal is to sign pre-generated 3-of-5 transaction with 1 signature and transmit it to other 2 signers.

Transaction is somewhat custom (omnilayer extension to be specific) so i cannot build it from scratch using bitcoinjs. So i'm getting pre-constructed unsigned transaction from another place.

Pre-constructed transaction comes with 3 OP_FALSE codes, supposed to be replaced by signatures, but after signing with bitcoinjs i'm there is a signature and 4 to-be-replaced OP_FALSE codes are there.

Steps to reproduce:

    privateKeyWIF = 'L1NCkTfLgduMYtLkXgVGgskYfPYkN23MCHDSukStcV8d1i1EwWjz'
    // transaction with no signatures
    txHex = '01000000019c1c878eff34fb9262a02b8490fd0efb59bbd8b083170dc24ff10e99c742ffda03000000b3000000004cad532103832abbe9737befa67ea259fe06541c2585ba29462bc86f4fc67ac8235600984a2103dd0cc537865c5f04eadd58812b63a2fcf39fcb8158673ef794608af57ebd58c721029db58dbf054452422f0b51d68f93b8ff4160f2f7a44edc2a7f3103ec3d872e222102d89857b54f59b58726560c6de7db15c787fbd2b2bb0c0b615351366de96b91712103fc035bc6f1336c82ac1cb6d6928f3d21cf3c0610f577551b8d565a001297ab4b55aeffffffff04ac02000000000000475121030c28fe5f18c699a30b62b7d67de92b7493658f2bab2a1806fbf5f0ea7f28e8db21022ac1c2aa5600b2e8e68126945bf9ac83aa663357c8018fdea271a4886e2e2b6852ae22020000000000001976a914946cb2e08075bcbaf157e47bcb67eb2b2339d24288ac22020000000000001976a9149e0070a1aaa842e95f5fb4639d26802526c6304a88ac601100000000000017a9146d73e76d90cfdf438f6b394fb6ff8f08b6bf2d198700000000'

    key = BitcoinJs.ECKey.fromWIF(privateKeyWIF)
    tx = BitcoinJs.Transaction.fromHex(txHex) 
    var txb = BitcoinJs.TransactionBuilder.fromTransaction(tx)
    txb.inputs[0].hashType = BitcoinJs.Transaction.SIGHASH_ALL
[...]

PR #1201, opened 2018-09-03
tests: add extra P2MS testing fixtures

dcousens:


src/classify.js
```diff
@@ -9,7 +9,7 @@ const witnessScriptHash = require('./templates/witnessscripthash')
 const witnessCommitment = require('./templates/witnesscommitment')
 
 const types = {
-  MULTISIG: 'multisig',
```
dcousens: This is a breaking change from `3.0.0`,  but,  we don't expose `classify` at this time (https://github.com/bitcoinjs/bitcoinjs-lib/issues/1148)

src/transaction_builder.js
```diff
@@ -392,14 +399,15 @@ function build (type, input, allowIncomplete) {
...
+      const m = input.maxSignatures
       if (allowIncomplete) {
         signatures = signatures.map(x => x || ops.OP_0)
       } else {
         signatures = signatures.filter(x => x)
       }
-
-      return payments.p2ms({ signatures }, { allowIncomplete })
+      const validate = !allowIncomplete || (m === signatures.length)
+      return payments.p2ms({ m, pubkeys, signatures }, { allowIncomplete, validate })
```
dcousens: If the transaction is "not not complete === complete",  or if `m === signatures.length` validate!
Otherwise the `OP_0` counts being different to `m` causes this to explode...

Maybe this comment should be an inline comment in the code...

dcousens: Related https://www.google.com/search?q=OP_0+multisig+incomplete

https://github.com/bitcoinjs/bitcoinjs-lib/issues/404

dcousens:
Rebased, with tests passing locally, if OK, merge :+1:

src/transaction_builder.js
dcousens: Added comment in https://github.com/bitcoinjs/bitcoinjs-lib/pull/1201/commits/7c3d1f292d6914c355e5b3e8cbd5ea732bb735cd

dcousens:
alternative to c8ae86b is to add `maxSignatures` to every type... but, it isn't needed by anything else so...

src/transaction_builder.js
dcousens: Looking back at this,  maybe `payments.p2ms` with `{ allowIncomplete: true, ... }` should be what is doing the special casing, not `TransactionBuilder`... thoughts @junderw?

Shall we make an issue to track this?

PR #1228, opened 2018-09-26
README: add usage notes

dcousens:
@junderw as discussed.

(work in progress)


README.md
```diff
@@ -42,12 +43,33 @@ If in doubt, see the [.travis.yml](.travis.yml) for what versions are used by ou
...
+
+Problems in [`Buffer (UInt8Array)`](https://github.com/feross/buffer), for example, can trivially result in catastrophic fund loss without any warning.
+It can do this through undermining your random number generation, accidentally producing a duplicate `k` value, sending Bitcoin to a malformed output script, or any of a million different ways.
+Running tests in your target environment is important and a recommended step to verify continuously.
+
+Finally, **adhere to best practice**.   We aren't an authorative source for best practice, but, at the very least:
+
+* Don't re-use addresses.  Privacy is important, but, .... TODO
+* Don't share BIP32 extended public keys. They are a liability, and [as shown in our examples](https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/test/integration/crypto.js#L68), it only takes 1 mistake until **catastrophic failure**.
+* TODO, anythign else of importance here?
```
dcousens: * Use `Typescript`?

Unfortunately, without up-to-date `typescript` bindings, that may seem hypocritical.

dcousens: * Enforce users manually verify a **freshly-decoded** human-readable version of your intended transaction before broadcast.

Even a stray neutrino could mess up your output script.

junderw: Reducing reliance on RNG:

1. Use RFC6979 (done)
2. Use BIP32 (this should be explicit. Every individual key you generate with RNG is a roll of the dice. If we can reduce that even more by just deterministically generating addresses it would be better...)
 (However, not sure if we should recommend BIP39/BIP32/BIP44 as a combo.......imo BIP39 is not a good standard, but it is standard... sigh)

dcousens: @junderw provided you did that 1RNG call, really well.  Yes.

dcousens:
Added suggestions, and removed old examples.

junderw:
rm stealth and rm bad sha256 brainwallet examples ACK

rm recover from reused r and recover from xpub and child privkey NACK

Nit: I think we should explicitly suggest using BIP32 wherever possible. (either after the don't reuse addresses part or after the RNG part (BIP32 is deterministic so less dependence on RNG))

PR #1233, opened 2018-09-28
Hk custom address

rufelo:


junderw:
hdnode was moved to a separate repo:

bip32

PR #1239, opened 2018-10-06
Add example for using in production

ak1103dev:


dcousens:
Maybe we can force `rng` to throw if it isn't run within a tape or mocha env?

ak1103dev:
The address.js is test script. And it is example for use the library. I mean remove "rng" function for using it in real code.

junderw:
@dcousens I agree.

Adding more comments probably won't help a person who copy pastes without understanding what's going on.

Adding logic to prevent rng setting would be good.

IMO, this should be something like:
(Though not sure how we'd check if running in test. Dumb idea below)

```javascript
} while (!ecc.isPrivate(d))

if (typeof global.it !== 'function' &&
    d.equals(Buffer.from('zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz'))) {
  throw new Error('DO NOT COPY PASTE THE EXAMPLE RNG FROM TESTS!!!')
}

return fromPrivateKey(d, options)
```

dcousens:
We put it in the `rng` provided in the tests, not `ECPair`.
PR accepted :+1:

dcousens:
Why not assert, don't encourage copy pasting

junderw:
Because I have seen copy-pasters in action, and they will:

1. Notice the throw was from code they pasted.
2. See a simple if that just throws the error.
3. "Hmm, what happens if I delete this if statement 3 lines.
4. "Yay it worked!"

That said, maybe there's a better way...

But sitting around thinking of the ways someone can shoot themselves in the foot seems less productive then making sure the gun they use only shoots nerf foam darts...

I'll make another PR with another suggestion that:

1. Will throw if 100% copy-pasted
2. Will be hard to tease apart and reason through. Making it hard to remove the throw.
3. If the throw is somehow removed, randombytes will be used.

Issue #1260, opened 2018-11-14
Can't create regtest p2wpkh address


Janaka-Steph:
Hello,

I am trying to run the [P2WPKH test](https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/test/integration/transactions.js#L165) on my machine, using Bitcoin Core on regtest mode to fund the p2wpkh address manually. 
The address starts with `tb`, which is not what I want because  Core is using `bcrt` address prefix on regtest.

Bitcoinjs doesn't have a regtest object in `bitcoin.networks`, so what am I suppose to do? 
Is bitcoinjs able to produce a regtest p2wpkh address?
Thank you.

junderw:
> Is bitcoinjs able to produce a regtest p2wpkh address?

If you create a custom network, yes.

The network object is not that complicated, you can make a custom one to meet your needs.

dcousens:
Check the integration tests.

`regtest` has the same constants as `testnet`.
We simply use `bitcoin.networks.testnet` for our `regtest` tests.

If you want to put up a PR which aliases `testnet` with `regtest`... I think that would be OK?

junderw:
@dcousens He is saying that regtest has a different bech32 value for segwit.

`bech32: 'bcrt'`

dcousens:
Make a PR please that adds `regtest` :+1:

junderw:
@dcousens https://github.com/bitcoinjs/bitcoinjs-lib/pull/1261

dcousens:
@Janaka-Steph please see #1261,  thanks @junderw

PR #1261, opened 2018-11-15
Add regtest network

junderw:
ECPair.fromWIF uses .pop() to grab the networks from the list in tests...

If we added regtest AFTER testnet the tests fail........

### DIFF
--- a/src/networks.js
+++ b/src/networks.js
@@ -13,6 +13,17 @@ module.exports = {
     scriptHash: 0x05,
     wif: 0x80
   },
+  regtest: {
+    messagePrefix: '\x18Bitcoin Signed Message:\n',
+    bech32: 'bcrt',
+    bip32: {
+      public: 0x043587cf,
+      private: 0x04358394
+    },
+    pubKeyHash: 0x6f,
+    scriptHash: 0xc4,
+    wif: 0xef
+  },
   testnet: {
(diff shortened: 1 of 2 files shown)


junderw:
Fixes #1260

dcousens:
OK... so if anyone expected `testnet` using that API, they will still get `testnet`?

src/networks.js
```diff
@@ -13,6 +13,17 @@ module.exports = {
     scriptHash: 0x05,
     wif: 0x80
   },
+  regtest: {
+    messagePrefix: '\x18Bitcoin Signed Message:\n',
+    bech32: 'bcrt',
```
dcousens: Any reference for this by the way?

Janaka-Steph: It is not mentioned in BIP173  because it is Bitcoin Core specific. The only reference I can find is the code itself https://github.com/bitcoin/bitcoin/blob/master/src/chainparams.cpp#L341

Janaka-Steph: It's also mentioned here https://github.com/satoshilabs/slips/blob/master/slip-0173.md

dcousens: Thanks @Janaka-Steph

Janaka-Steph:
I see that it has still not been added to the npm version. When do you plan to release a new version?

Janaka-Steph:
Ok thank you.

PR #1268, opened 2018-11-27
README: upgrade iOS warning

dcousens:
Hopefully this is better.
Is there any definitive source for what these issues were?
This is the best I could ascertain from what is reported thus far.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ npm install -g flow-typed
 flow-typed install -f 0.27 bitcoinjs-lib@2.2.0
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -74,7 +74,7 @@ npm install -g flow-typed
 flow-typed install -f 0.27 bitcoinjs-lib@2.2.0
 ```
 
-These definitions are maintained by [@runn1ng](https://github.com/runn1ng).
```
dcousens: This user is completely unrelated to who it was previously...

Issue #1142, opened 2018-07-25
Adding signature data to transaction with external signature


patwhite:
I'm trying to use bitcoin.js with a hardware wallet, where the wallet would provide the signature, but bitcoin.js would produce the final transaction hex (so I don't have to manually construct it, though that is an option). This is a 1 of 2 multi-sig transaction. The transaction ids are on testnet if you'd like to examine them.

Excuse the somewhat poor code here:

    const sig = getHardwareSig()
    const redeem = "512103bb9dad581f8942409140bb6082c39e08d286b8d841064ad2535a3e9743ff192e21031eb52a32cfa8da5dc1d66caab04037c57caaf2d8fc36796c7dfd485af0b1ad4e52ae"
    var buf = new Buffer(redeem)
    txb.addInput('89141edf593b88908f14192b6b6d782b5a7b128530f5ad49085f0851b604b757', 0, null, buf)
    txb.addOutput('2N6GwVhv9G4rq8VpiCbQQ6Vged5hxZVDXRy', 9000)
    txb.inputs[0].signatures = []
    txb.inputs[0].signatures[0] = sig
    const tx = txb.build()
    var fullTx = tx.toHex()

This results in an error: 

    Uncaught Error: nonstandard not supported
    at eval (transaction_builder.js?c914:638)
    at Array.forEach (<anonymous>)
    at TransactionBuilder.__build (transaction_builder.js?c914:630)
    at TransactionBuilder.build (transaction_builder.js?c914:616)

Is this a supported scenario? Is it possible to get this code to work?

patwhite:
I saw this in the docs:
![image](https://user-images.githubusercontent.com/5015863/43177461-4422bd58-8f7d-11e8-83ce-2eea57b819b2.png)

Failing to provide the redeem script caused a separate issue around not being able to build the transaction.

patwhite:
Realized I was misusing the Buffer(rs) - I fixed that issue and was actually able to get a transaction out. I'm verifying it now, but I'd love some feedback if this is the best way to handle this problem.

Thanks in advance!

Marxpark:
I'm having a similar problem when trying to add a nano ledger signature for a transaction. 
I prepare the transaction online and sign it with my ledger offline. Then I take the signature back online, and try to add it to the transaction. I've managed to get this process working with signature generated with seed, but not for signatures generated with nano ledger. 

I.E. take this transaction: 
[...]

PR #1271, opened 2018-11-29
add hardware wallet signing example

dcousens:


test/integration/transactions.js
```diff
@@ -258,6 +259,27 @@ describe('bitcoinjs-lib (transactions)', function () {
...
+    txb.setVersion(1)
+    txb.addInput('61d520ccb74288c96bc1a2b20ea1c0d5a704776dd0164a396efec3ea7040349d', 0) // Alice's previous transaction output, has 15000 satoshis
+    txb.addOutput('1cMh228HTCiwS8ZsaakH8A8wze1JR5ZsP', 12000)
+    // (in)15000 - (out)12000 = (fee)3000, this is the miner fee
+
+    txb.sign(0, {
+      publicKey: Buffer.from('029f50f51d63b345039a290c94bffd3180c99ed659ff6ea6b1242bca47eb93b59f', 'hex'),
+      sign: function (hash) {
+        const privateKey = Buffer.from('8c112cf628362ecf4d482f68af2dbb50c8a2cb90d226215de925417aa9336a48', 'hex')
+        return tinysecp.sign(hash, privateKey)
```
junderw: You should add a comment here saying what format this sign function expects.

```javascript
// sign expects a 64 byte Buffer where the first and last 32 bytes correspond with the r and s values respectively of the ECDSA signature
```

etc.

junderw: expects meaning "expects as a return value"

sign referring to line 273, the function we are declaring here.

dcousens: @junderw can you add it? :+1:

test/integration/transactions.js
```diff
@@ -258,6 +259,36 @@ describe('bitcoinjs-lib (transactions)', function () {
...
+    txb.addOutput('1cMh228HTCiwS8ZsaakH8A8wze1JR5ZsP', 12000)
+    // (in)15000 - (out)12000 = (fee)3000, this is the miner fee
+
+    const fakeECPair = {
+      // publicKey should be a Buffer with the DER encoded public key
+      publicKey: Buffer.from('029f50f51d63b345039a290c94bffd3180c99ed659ff6ea6b1242bca47eb93b59f', 'hex'),
+      // sign is a function that takes a 32 byte Buffer, signs the 32 byte Buffer directly (no extra hashing)
+      // And returns a 64 byte Buffer:
+      //   The first 32 bytes is the signature r value in big endian
+      //    The last 32 bytes is the signature s value in big endian
```
dcousens: spacing is odd

test/integration/transactions.js
```diff
@@ -264,17 +264,26 @@ describe('bitcoinjs-lib (transactions)', function () {
...
     txb.addOutput('1cMh228HTCiwS8ZsaakH8A8wze1JR5ZsP', 12000)
     // (in)15000 - (out)12000 = (fee)3000, this is the miner fee
 
-    txb.sign(0, {
+    const fakeECPair = {
+      // publicKey should be a Buffer with the DER encoded public key
       publicKey: Buffer.from('029f50f51d63b345039a290c94bffd3180c99ed659ff6ea6b1242bca47eb93b59f', 'hex'),
+      // sign is a function that takes a 32 byte Buffer, signs the 32 byte Buffer directly (no extra hashing)
+      // And returns a 64 byte Buffer:
+      //   The 1st 32 bytes is the signature r value in big endian
```
dcousens: `first` , `second`? :smiley:

dcousens: Maybe capitalize R/S too?

dcousens: Do we need to mention they are zero padded?

junderw: lol... if we're going to start a grammar review on every comment in the repo we're gonna be here a while... lol

`head` and `tail`? The head 32 bytes and The tail 32 bytes have the same letter count.

dcousens: > Do we need to mention they are zero padded?

junderw: true... I'll add that.

junderw: done, should be fine.

patwhite:
So, I would rather be commenting on #1142 but it's been locked for some reason - I don't think this is just a documentation issue, I think there should be a better model for doing this than overloading the ECPair. For instance, an easy solution would be an "applySignature" method on the transaction builder that would apply an external signature and would modify the internals of the transaction correctly. If that is an ok solution, I'm happy to code it up, but would like feedback from the developers about if that's a good strategy.

dcousens:
@patwhite the thread was going in circles.  I have used this solution many times,  and I personally don't see it as hacky at all.
We aren't recommending overloading at all,   this is merely using the same `interface` as ECPair.

patwhite:
Well, I'd say the hackiness comes from the fact that it's completely undiscoverable except from documentation, that's my big thing. Even for someone relatively familiar with this library, you would never be able to figure this out without going pretty deep into the internals. Something like .applySignature on the transaction builder object would be discoverable and you'd ultimately be interacting with the transaction in a consistent manner across all changes to the txn.

But, agreed, this is going around in circles, if this is the recommendations from the devs, all good.

patwhite:
Oh interesting, my thinking was that, as a user, you shouldn't really mutate the transaction object directly, but rather only through the transaction builder. I was thinking TB -> incomplete transaction, get signature from there, load incomplete back into TB, then applySig and finish.

If there was something similar on the transaction, applySignature or whatever, that would definitely also work. Basically, abstracting the interaction rather than having to go deep into the transaction itself to change array level things. Put another way, the API for signing with a PK is actually nice, you identify the input, etc. Something similar to that to apply the sig is probably nice.

junderw:
If ever there is a point where the best way to implement something is to txb.buildIncomplete, do something with the Transaction, then turn it back into a Transaction builder, then something is wrong.

IMO, Transaction does not have the concept of signatures, only scripts, so applySignature would not make sense on Transaction.

Meanwhile, TransactionBuilder only understands signatures within the context of specific paradigms we have set out for it.

If we were to add an API, I would suggest something like txb.getSigningInfoForInput. Which would take the same info as sign and return:

1. The preimage for the signature. (hash256 this Buffer to get the arg for ECPair. sign)
2. The amount of the input (if available)
3. The pubkey that needs to verify the signature.
4. isSegwit bool
5. index of the input
6. Transaction from buildIncomplete

The only other info needed would be the previous Transaction (for non-segwit only) so that you can verify the input amount by checking the hash, vout and amount of the output... however, we have no way of knowing this without the user providing it.

Then the user could get the signature while verifying all the info. Not to mention most hardware wallets should require this info as a prerequisite for spitting out the sig.

Then adding a setSignatureInfo API that basically takes the same info with a signature added. And we triple check that the sig verifies, the preimage matches what we think we should be, and then applies the signature.

Which is the bare minimum we would need to do...

However, outsourcing the signing is in general a dangerous proposition, so there will be tons of overhead checking the validity of the signature before committing.

I think it might be worth exploring, but it increases the room for error, so in general I am with @dcousens on this.

patwhite:
Well, it's you guy's code, so up to you if it's not something you want in there, I won't spend time on it. But, I don't think it's reasonable to say that rehydrating a transaction builder doesn't make sense since hardware signing is more or less necessarily an asynchronous act, so there's going to be some amount of dance there. I'd also offer, better support for hardware signing is a net positive on security overall (not just for this library, but for the ecosystem in general), so I don't think it's totally fair to say outsourcing signing is a dangerous proposition - hopefully it ends up increasing security overall!

But, I like the API you proposed, it's something I'd happily kick if there is interest from the dev team.

dcousens:
> outsourcing the signing is in general a dangerous proposition, so there will be tons of overhead checking the validity of the signature before committing.

Basically this.
Javascript is terrifying,  Bitcoin with Javascript is a nightmare.
We really try to not add API's that may become a loaded foot gun down the track.

> Well, it's you guy's code

This library is open source!
@patwhite where are you getting the `hash` and `signature` from in your `applySignature` scenario?
Would an asynchronous `sign` option resolve your problem?

dcousens:
Maybe we should push for BIP174 module/implementation, rather than try to extend this abstraction to support hardware wallets.

Thoughts @junderw? NACK this?

PR #1279, opened 2018-12-06
add P2SH(P2WPKH) signature verification example

dcousens:
This was actually trickier than expected,  thanks to @ngmiller for pointing that out.

@junderw maybe we should add `.outputForSigning` or something (suggestions for name please!) to `payments.p2wpkh`?

We could use it in https://github.com/bitcoinjs/bitcoinjs-lib/blob/9a8552acda9c1881c43b3b6a9b42fb6b46ebd424/src/transaction_builder.js#L277-L280 (and in this case)

### DIFF
--- a/test/integration/transactions.js
+++ b/test/integration/transactions.js
@@ -258,7 +258,7 @@ describe('bitcoinjs-lib (transactions)', function () {
     })
   })
 
-  it('can verify Transaction signatures', function () {
+  it('can verify Transaction (P2PKH) signatures', function () {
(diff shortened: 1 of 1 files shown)


test/integration/transactions.js
```diff
@@ -281,4 +281,39 @@ describe('bitcoinjs-lib (transactions)', function () {
...
+      'f72d1d83ac40fcedd01415751556a905844ab5f44bbb7728565ebb91b1590109:0': {
+        value: 50000
+      }
+    }
+
+    const txHex = '02000000000101090159b191bb5e562877bb4bf4b54a8405a95615751514d0edfc40ac831d2df7000000001716001435a179e5516947a39ae9c8a25e9fe62c0fc598edffffffff01204e0000000000001976a91431d43308d3c886d53e9ae8a45728370571ff456988ac0247304402206ec41f685b997a51f325b07ee852e82a535f6b52ef54485cc133e05168aa052a022070bafa86108acb51c77b2b259ae8fb7fd1efa10fef804fcfe9b13c2db719acf5012103fb03e9d0a9af86cbed94225dbb8bb70f6b82109bce0a61ddcf41dab6cbb4871100000000'
+    const tx = bitcoin.Transaction.fromHex(txHex)
+
+    tx.ins.forEach(function (input, i) {
+      const txId = Buffer.from(input.hash).reverse().toString('hex')
```
dcousens: This is lame.
Since we have accepted using Object properties... maybe it is time to add `.txId` as a lazy property?

ngmiller:
🙌 Thank you @dcousens!

test/integration/transactions.js
```diff
@@ -281,4 +281,34 @@ describe('bitcoinjs-lib (transactions)', function () {
...
+      const utxo = utxos[`${txId}:${i}`]
+      if (!utxo) throw new Error('Missing utxo')
+
+      const p2sh = bitcoin.payments.p2sh({
+        input: input.script,
+        witness: input.witness
+      })
+      const p2wpkh = bitcoin.payments.p2wpkh(p2sh.redeem)
+      const p2pkh = bitcoin.payments.p2pkh({ pubkey: p2wpkh.pubkey }) // because P2WPKH is annoying
```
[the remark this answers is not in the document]
junderw: all should add prevOut as an alias for output.

p2wpkh being special and actually adding some calculations.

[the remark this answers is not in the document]
junderw: or outputForSigning?

just an alias for all but p2wpkh

test/integration/transactions.js
junderw: maybe...

I agree with the lameness.

junderw:
I like outputForSigning.

but we should make it an alias for output for all other types as well imo

test/integration/transactions.js
donza9Pk: May be

PR #1285, opened 2018-12-12
Fix Block to allow regtest target (easiest possible target)

junderw:
when exponent was 29 (32 - 3) the write was being performed in index -1 which caused an error, so Block could not parse blocks made on regtest (bits `0x207fffff`)

Added a test to cover it.


dcousens:
How is this not truncated?

junderw:
because we're only writing 3 bytes.

junderw:
bits is 4 bytes.

1 byte for exponent,

3 bytes for mantissa.

we never write more than 3 bytes.

PR #1289, opened 2018-12-18
Refactor Block to ES6 Class

junderw:
In preparation for a possible rewrite into TypeScript. I will continue forward with making use of ES6 Classes where possible. (it will be easier to port to ts)

PR #1292, opened 2018-12-18
Add Witness Commit checking for Block

junderw:
This PR is based on #1289 and the merge branch is pointed at the ES6 Block branch.

Need to be able to verify witness commits for block verification moving forward.


src/block.js
```diff
@@ -165,6 +193,14 @@ class Block {
     return this.merkleRoot.compare(actualMerkleRoot) === 0
   }
 
+  checkWitnessCommit () {
+    if (!this.transactions) return false
+    if (!this.isWitnessBlock()) return false
```
junderw: I wonder if we should return true here...

Since the no transactions check is saying "Having no transactions is an invalid state therefore false"

But a non-witness-block not having a witness commit to check is the expected valid state.

So if I run checkWitnessCommit on a block with no witness, maybe we should return true???

junderw: If we don't do this, every consumer of this API will need to dance around checking "if isWitnessBlock, then checkWitnessCommit, else don't" etc...

Maybe we want them to do this?

dcousens: I don't think `true` is acceptable.
This is a misunderstanding of the situation no matter how you look at it.
The result is nothing was done,  so maybe `null`,  but honestly,  `Error` seems appropriate.

Like calling `.privateKey` on `ECPair` with only `.publicKey`.

junderw: Yeah.

Maybe I should make checkMerkleRoot also check the witnessCommit when the block has witnesses.

or make a new function called checkTransactionTrees or something, so that people don't need to worry much about whether the block is witness or not.

dcousens:
``` js
    if (!this.transactions) throw new TypeError('Missing transactions')
```

?

dcousens: As for `if (!this.isWitnessBlock()) return false`,  again,  I don't know.
Ideally we'd not have this situation of ambiguity.

Should we "lift" the type somehow?

dcousens: `Transaction` has `hasWitnesses`.

Maybe `hasWitnessCommit`,  force user to check that,  throwing if `checkWitnessCommit` is called and `hasWitnessCommit` is false...

What is the usecase? Merely validation?

dcousens: We do https://github.com/bitcoinjs/bitcoinjs-lib/blob/293116b20ff63473e760039793e5432ffb4a96f2/src/block.js#L164 :-1:

junderw: Thinking of Lightning Network.

We want to eventually support client side block filtering, which means we only download blocks which match the filter, but we download the full block...

So if we download the full block, and LN uses segwit, we would like a way to verify the witness root as well.

So yeah, merely validation.

junderw:
@dcousens Since the point of the previous checkMerkleRoot is "to check that all transactions are represented in the block properly"... it might make sense to offer a new function or replace the old function with it's meaning in the SegWit context... thoughts?

Perhaps we could do one of two things:

1. Make a new function encapsulating the tx root checking: ie.
```javascript
checkTxRoots () {
  return this.checkMerkleRoot() &&
    (this.hasWitnessCommit() ? this.checkWitnessCommit() : true)
}
```
2. Make checkMerkleRoot implicitly check the WitnessCommit when it has one.
```javascript
__checkMerkleRoot () {
  if (!this.transactions) {
    throw new TypeError('Cannot compute merkle root for zero transactions')
  }

  const actualMerkleRoot = Block.calculateMerkleRoot(this.transactions)
  return this.merkleRoot.compare(actualMerkleRoot) === 0
}

checkWitnessCommit () {
  if (!this.transactions) {
    throw new TypeError('Cannot compute merkle root for zero transactions')
  }
  if (!this.hasWitnessCommit()) {
    throw new TypeError('Cannot compute witness commit for non-segwit block')
  }

  const actualWitnessCommit = Block.calculateMerkleRoot(this.transactions, true)
  return this.witnessCommit.compare(actualWitnessCommit) === 0
}

checkMerkleRoot () {
  return this.__checkMerkleRoot() &&
    (this.hasWitnessCommit() ? this.checkWitnessCommit() : true)
}
```

junderw:
Maybe even give the option:

```javascript
checkMerkleRoot (checkWitness) {
  if (checkWitness === undefined) checkWitness = true
  return this.__checkMerkleRoot() &&
    (checkWitness && this.hasWitnessCommit() ? this.checkWitnessCommit() : true)
}
```

mandelmonkey:
ACK

junderw:
@mandelmonkey Thanks for the review!

Any thoughts on the method structure? Should we automatically check the witness root if the block has a witness hash?

mandelmonkey:
@junderw I think as you outlined it makes sense to make a new function, and deprecate the old  checkMerkleRoot with an appropriate warning message, anybody using this function would likely be aware of the structure of new segwit blocks compared to legacy blocks (although thats not always a safe bet). Anyway I think it also makes sense in the new function to automatically check the witness root if there is a witness hash, providing the name of the function is checkTxRoots as it makes sense that such a function would do that.

dcousens:
It isn't deprecated no? Merely 'legacy'?

junderw:
@dcousens By deprecation I mean that as a library we are moving it to a "private" function.

The reasoning being anyone who uses checkMerkleRoot as an alias in their mind for "check that all the Transactions are represented in the block header" should be aware that this assumption is no longer true for segwit blocks.

to snap anyone out of that assumption, a deprecation warning to use a more generic function that supports any root checking is ideal imo.

Then if down the line a new witness commitment is added, we can add that check to checkTxRoots without any problems, whereas if we add witness checking to checkMerkleRoot, the function name doesn't match what it's doing.

src/block.js
```diff
@@ -68,6 +84,20 @@ class Block {
...
+      // The merkle root for the witness data is in an OP_RETURN output.
+      // There is no rule for the index of the output, so use filter to find it.
+      // The root is prepended with 0xaa21a9ed so check for 0x6a24aa21a9ed
+      // If multiple commits are found, the output with highest index is assumed.
+      let witnessCommits = block.transactions[0].outs
+        .filter(out => out.script.slice(0, 6).equals(Buffer.from('6a24aa21a9ed', 'hex')))
+        .map(out => out.script.slice(6, 38))
+
+      // Use the commit with the highest output (should only be one though)
+      block.witnessCommit = witnessCommits[witnessCommits.length - 1]
```
junderw: I should probably check that witnessCommits.length > 0 here and throw an Error if it is.

Since having a witness for the first Transaction but not having the witness commit anywhere in the outputs would mean an invalid block.

junderw:
closed in favor of #1319

PR #1236, opened 2018-09-29
[WIP] Integrated TS types + type tests

torkelrogstad:
Work in progress, not ready for merge. 

I'm opening this PR to get some feedback if this is a nice way to do the TS definitions. I took the types from DefinitelyTyped, and tweaked them a bit. The type checking tests only checks for return types. Is this sufficient, or should parameters be tested as well? Would be a lot of work...

PR #1314, opened 2019-01-01
Add TypeScript types

junderw:
I got this from compiling my attempt at converting to TypeScript. (https://github.com/junderw/bitcoinjs-lib/tree/typeScript)

Seems to work, but not exactly sure how one would test a TypeScript definition since it's all static analysis...


junderw:
Also this is an alternative to #1236 by @torkelrogstad

junderw:
closed in favor of #1319

Issue #1200, opened 2018-08-30
Integrated TypeScript definitions


torkelrogstad:
How do you feel about having TypeScript definitions included in this library? Currently, it's managed through DefinitelyTyped. One downside to this is that types have to be installed separately, increasing the risk of user error. It's also more difficult to keep changes JS/TS in sync, instead of being in control of the entire release process.

Another solution would also be to port the JS itself to TS. By adding a build step _(which might even be done currently, for backwards JS compatability?)_, the library could be distributed to JS users as well. It is possible to port to parts of libraries to TS, it doesn't have to be in one big step.

Seeing as this library deals with monetary value, strict and safe programming is paramount. Static typing makes this a great deal safer.

I'd be glad to put in some work to either write definitions or port (parts) of the library, if that's desirable.

dcousens:
I am not against this anymore,  but,  I am now using typescript myself.

If you could add the type definitions such that they are verified against the existing API in an independent test,  I'd be happy to consider them under a `/types` directory.

dcousens:
@torkelrogstad still interested in helping here?

jprichardson:
> I am now using typescript myself.

I figured you would eventually 😛I've also got bit by the TypeScript bug as well recently.

torkelrogstad:
@dcousens I'm interested! I'll try to get some done tomorrow:)

zone117x:
This can be closed now.
Great work on the typescript port!

PR #1319, opened 2019-01-17
Initial TypeScript implementation

junderw:
In order to add static typing while working on BitcoinJS-lib and automatically generate the type files for use in other TypeScript libraries, I have re-written the library using TypeScript.

A few caveats:

1. Pull requests should modify the ts files, compile, and commit all ts, js, and d.ts files.
2. Before reviewing the JS, a reviewer should build locally and see if a diff occurs.
3. To try and prevent diffs from occurring, setting typescript to a specific version, and only bumping up the version with a re-build and review of the newly compiled JS files (in case of any diff)

ToDos before merging:

1. Thorough review of the TS and JS.
2. Discuss which dependencies we should try and move to TS (at least with definitions, but if this TS methodology above is fine, we could mimic it with other dependencies, ie bip32)
3. Update the dependencies to reflect versions with TS in them. (Currently I have a dependency pointing to the bip32 TS pull request branch)

Any and all feedback is appreciated, and I will try to keep things up to date with master functionality as things change.

I currently have the ES6 PRs and the WitnessBlock checking PRs merged into this branch.

Working on this gave me a new found appreciation for TypeScript.

Any comments are appreciated.

Edit: At this point, diffs kind of lose meaning... so a review would probably consist of:

1. checkout branch
2. npm install
3. npm test (this will perform a build in the process)
4. check to make sure the build didn't cause a diff. (it should be deterministic iirc)
5. check all the files. in src, ts_src, and types
6. check `git diff master -- test/` and see that it's only the new block stuff and a few things to accommodate the new TypeScript structure.


junderw:
At this point, diffs kind of lose meaning... so a review would probably consist of:

1. checkout branch
2. npm install
3. npm test (this will perform a build in the process)
4. check to make sure the build didn't cause a diff. (it should be deterministic iirc)
5. check all the files. in src, ts_src, and types
6. check `git diff master -- test/` and see that it's only the new block stuff and a few things to accommodate the new TypeScript structure.

junderw:
Related issues / PRs that will be completed if this is merged:
#1314 
#1289 
#1236 
#1200 

Feel free to reference other issues that might be interested in this PR.

package.json
```diff
@@ -14,24 +15,34 @@
...
     "type": "git",
     "url": "https://github.com/bitcoinjs/bitcoinjs-lib.git"
   },
   "files": [
-    "src"
+    "src",
+    "types"
   ],
   "dependencies": {
+    "@types/node": "10.12.18",
```
johnta0: I _prefer_ `@types/node` to be included in devDependencies, or you might have the solid reason to add it to dependencies?

junderw: From my understanding by issues like this:
https://stackoverflow.com/a/46011417

1. Applications should set all types to devDependencies.
2. Libraries should set any types that are exposed to the consumer via themselves in dependencies, and devDependencies should only be for types that are solely used in testing.

That being said... I think most apps will have `@types/node`... so maybe it is ok to move to devDependencies.

ts_src/address.ts
```diff
@@ -0,0 +1,102 @@
+import * as Networks from './networks'
```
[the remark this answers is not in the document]
junderw: Thanks. Could you make a pull request with typeScript branch as base branch?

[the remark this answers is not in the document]
johnta0: I did.
https://github.com/bitcoinjs/bitcoinjs-lib/pull/1323

package.json
dcousens: > I think most apps will have @types/node... so maybe it is ok to move to devDependencies.

:+1:

d-yokoi:
@junderw 
Hi, I'm considering adding a formatter and a linter with some basic configurations.
What do you think?

junderw:
great! please make a pr to the typescript branch.

d-yokoi:
please review this pr for using prettier, which is one of the most popular formatters.
https://github.com/bitcoinjs/bitcoinjs-lib/pull/1343

i'll work on a linter after this pr is fixed.

junderw:
@d-yokoi Thanks for the PR, this brings the TypeScript formatting more in line with TypeScript norms.

package.json
```diff
@@ -14,24 +15,35 @@
...
     "bech32": "^1.1.2",
-    "bip32": "^1.0.0",
+    "bip32": "git+https://github.com/junderw/bip32.git#typeScript",
```
d-yokoi: Why don't you push your work to original repository and create a pr so anyone can collaborate on it?

junderw: https://github.com/bitcoinjs/bip32/pull/10

d-yokoi: Sorry, I have missed it.

junderw:
1. After a little more review, we'll merge the BIP32 TS branch and push out a new major version (just in case)
2. Update the dependency here to point to that version.
3. Merge here and bump to v5 just in case.

Probably some time in March.

Also, after merging this I will make a new PR with a concrete upgrade guide from v3 since I still see many people creating issues that amount to "read the new examples."

junderw:
Currently, d-yokoi and myself believe this is ready to merge.
I have pinged some of the devs working on type definitions in Definitely-typed to also get their opinion.

"Like to haves" left:
- TypeScript support for more dependencies.
- Rewrite tests in TypeScript as well.
- `ts_src` folder name change...

d-yokoi:
> ts_src folder name change...

how about `src` and `dist`? this is a more popular naming convention.

junderw:
> how about src and dist? this is a more popular naming convention.

dist implies that the files are only for "distribution" and are not "source" (src) files.

Since we are committing JS to git, JS files should always be checked and verified along side TS files.

dist is more common, but that is because most TS projects do not commit (or even look at) the JS code.

This project is not like that... so a new naming convention is needed.......... but I don't like ts_src... haha

youssefgh:
JS files generated with a specific typescript version + configs should be practically the same, maybe same spaces/tabs/EOL diffs but that shouldn't alter the behavior of the Lib whether a developer use the JS files or the TS ones
I believe that including JS files instead of only generating them when needed would make the review process a bit harder than usual without real added benefits
I couldn't find a reason to include JS in the above comments but maybe am missing some points here
Thank you for the big effort you're putting in this amazing Lib

junderw:
> without real added benefits

Verifiability.

The JS files are committed because people should be given the option to not trust TypeScript/Microsoft if they wish.

TypeScript is only here for two reasons:

1. Makes developing the library more secure, adding static type analysis to the runtime type analysis given by typeforce.
2. Auto-generates type definitions.

An application has the luxury of deciding whether it wants to trust MS.

However, this library is used by people who may want to just verify the JS and not use TS at all. Having git diffs helps that verification.

d-yokoi:
It's common to commit only `.ts` files but publish only `.js` files, which is achieved by using `.gitignore` and `.npmignore` together.

By doing so, users don't necessarily trust TypeScript/Microsoft.

youssefgh:
Now i see the point of keeping the JS files
That make sense for Libs that deal with Bitcoin

junderw:
@d-yokoi That is true, but keeping JS files on git makes it easier for anyone forking us to verify their changes to the ts files are all that is added to the js files. (via git diff)

Plus, that would mean everytime I publish I need to manually verify the whole JS folder.

d-yokoi:
I understand your thoughts.

Then, why don't we leave `.gitignore` as it is and exclude `.ts` files and some config files used only for development from the published package?

Making the size of the package smaller helps users.

d-yokoi:
I have finally noticed `.ts` files and other config files are not to be published from the beginning.

So, please ignore my comment above. Everything looks good to me now :)

Issue #1320, opened 2019-01-17
How can I deal with package-lock.json?


johnta0:
Since npm v5, `package-lock.json` is auto-generated after you run `npm install`. 
I suggest that you add `package-lock.json` to .gitignore, or commit it. 

What do you guys think? Thanks.

junderw:
I'm all for committing it. If you want to make a pull request I'll review it.

PR #1325, opened 2019-01-23
Add package-lock.json

johnta0:
I generated `package-lock.json` by running `npm install`.
Please check if the versions of dependencies are appropriate.

This pull request closes https://github.com/bitcoinjs/bitcoinjs-lib/issues/1320

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,4745 @@
+{
+  "name": "bitcoinjs-lib",
+  "version": "4.0.2",
+  "lockfileVersion": 1,
+  "requires": true,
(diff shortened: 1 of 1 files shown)


junderw:
LGTM. Thanks for the PR.

I re-ran npm install on npm 6.7.0 and it added some things, so I included that.

dcousens:
I don't think this needs to be committed?
`npm` ignores it **unless** it is at the top level of your dependency tree,  aka,  unless you run `npm install` inside the `bitcoins-lib` repository.

Wouldn't it be better to `.gitignore` it?

PR #1343, opened 2019-03-03
replace standard with prettier and apply format

d-yokoi:
I replaced standard with prettier and applied format.
This pr includes only changes that do not affect the meaning of the code.


junderw:
I'll have to look into this before merging.

Since it creates a pretty large diff it might take a while.

junderw:
Looks like when I merged other PRs from you it caused a conflict, could you merge the typeScript HEAD into your branch please?

package.json
```diff
@@ -19,16 +19,17 @@
...
     "nobuild:coverage-html": "nyc report --reporter=html",
     "nobuild:coverage": "nyc --check-coverage --branches 90 --functions 90 --lines 90 mocha",
     "nobuild:integration": "mocha --timeout 50000 test/integration/",
-    "nobuild:standard": "standard ts_src/**/*.ts",
     "nobuild:unit": "mocha",
     "prepare": "npm run build",
-    "standard": "npm run build && npm run nobuild:standard",
-    "test": "npm run build && npm run nobuild:standard && npm run nobuild:coverage",
+    "prettier": "prettier ts_src/*.ts ts_src/**/*.ts --ignore-path ./.prettierignore",
+    "test": "npm run build && npm run nobuild:coverage",
```
junderw: `npm test` should also warn the user if their code is non-standard, so we should add `format:ci` here as well.

d-yokoi: done
https://github.com/bitcoinjs/bitcoinjs-lib/pull/1343/commits/763cb8b92312b817b657489b34d1dc3398932502

junderw:
Thanks for the PR. I went ahead and removed eslint comments for the standard library.

I have checked the full diff, and attempted running the same steps on the typeScript branch to verify it gives the same results. It did.

I will go ahead and merge this into the typeScript branch since it seems like prettier is widely used among typeScript projects.

Issue #1363, opened 2019-03-19
Hash name for ripemd160 is changed for LibreSSL


liangzan:
Node.js's crypto module is dependent on the version of OpenSSL on the platform. On OSX, it has switched to LibreSSL which has a different name for ripemd160. It changed from `rmd160` to `ripemd160`, and it affects line https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/src/crypto.js#L4.

junderw:
It looks like this part would need to be modified to account for both patterns. Please look into writing a pull request here:

https://github.com/crypto-browserify/createHash/blob/master/index.js

Once it is merged upstream we will consider updating the dependency.

junderw:
LibreSSL 2.6.5 was able to work perfectly fine on the latest Macbook.

Can you provide repro steps for your problem?

I can not reproduce.

junderw:
Please try using a more recent version of node:

It looks like crypto.createHash in more recent versions of node has rmd160 and ripemd160 both work. The only explanation would be that your node version is old.

Closing this issue.

```
$ node -v
v10.15.3
$ npm -v
6.9.0
```

PR #1364, opened 2019-03-19
Caters for a different RIPEMD160 name for LibreSSL

liangzan:
Solution for #1363. It catches the error and attempts to create a hash using the new name.

### DIFF
--- a/src/crypto.js
+++ b/src/crypto.js
@@ -1,7 +1,13 @@
 const createHash = require('create-hash')
 
 function ripemd160 (buffer) {
-  return createHash('rmd160').update(buffer).digest()
+  let hash
+  try {
+    hash = createHash('rmd160')
+  } catch (err) {
+    hash = createHash('ripemd160')
+  }
+  return hash.update(buffer).digest()
 }
 
 function sha1 (buffer) {


junderw:
do all tests pass under your broken environment?

liangzan:
if you run `openssl list-message-digest-commands`, it shows the list of hashes as
```
gost-mac
md4
md5
md_gost94
ripemd160
sha1
sha224
sha256
sha384
sha512
streebog256
streebog512
whirlpool
```
with the openssl version 
```
LibreSSL 2.6.5
```

For those with openssl version "OpenSSL 1.x", it will show as "rmd160". This is verified for our engineers. This update fixes the builds with LibreSSL.

junderw:
LibreSSL 2.6.5 was able to work perfectly fine on the latest Macbook.

Can you provide repro steps for your problem?

I can not reproduce.

junderw:
Please try using a more recent version of node:

It looks like crypto.createHash in more recent versions of node has rmd160 and ripemd160 both work. The only explanation would be that your node version is old.

Closing this issue.

```
$ node -v
v10.15.3
$ npm -v
6.9.0
```

PR #1373, opened 2019-04-01
Change rmd160 -> ripemd160 for Electron v4

RyanZim:
Electron v4 only supports ripemd160; no support for the rmd160 alias.


RyanZim:
Looks like tests are failing due to a network error

junderw:
This could break someone's app.

Just like Electron making the assumption that no one uses the rmd160 algorithm, us making the assumption that all of our users will have "ripemd160" could break something.

Please put in a try catch.

see the BIP32 PR for details.

also, we need to bump the BIP32 dependency to use the new patch.

NACK until these changes are made.

jprichardson:
> This could break someone's app.

Absolutely. What if we bring it in with the next major of bitcoinjs-lib?

junderw:
Yeah that's what I was thinking... Perhaps for the typescript v5 bump.

RyanZim:
@junderw What were the breaking changes in bip32 v2.0.0?

junderw:
TypeScript support changed some of the internal variable names.

While we had them named __d etc. to denote "don't rely on this" we can't be sure, so bumped major.

junderw:
close in favor of #1319

junderw:
... You know what... after thinking about it, we are basically just passing the bip32 library as-is as a sub-attribute of this library... and we're bumping major for it in order to fix this problem... which could cause problems for apps if they rely on the internals of BIP32 (which they shouldn't, but still)

I think we'll wait until the next major version ( TypeScript #1319 ) I have gone ahead and added this.

For now, just use Electron v3, or patch your bundle before deploying.

junderw:
backported the patch, this should be fine for release.

junderw:
v4.0.4 is published.

PR #1377, opened 2019-04-06
Export script classification utilities

### LINKED ISSUE #1276: bitcoin.script.classifyOutput is not a function
I have been trying to find it's alternative for hours now but lack of official documentation makes it difficult to find proper method 

after going through number of issues i know that this function was deprecated or removed but what is its alternative ?

kklash:
Fixes #1276. This is so people don't have to use a hacky import statement to get access to these utility functions. IDK if usage of these utils need to be added to the docs or examples as well.


junderw:
Thanks for the pull request.

As discussed in the issue you linked to:

The classify module is used internally in TransactionBuilder, and has a few problems.

1. Classification is not deterministic, and a script that was crafted specifically to trick the classify module could cause people to lose money depending on the usage.
2. It makes broad assumptions about the intent of the script and has a catch-all of 'unknown'.
3. Moving forward with Schnorr etc. could cause overlap. A simple protocol change or two later and "don't use bitcoinjs-lib vX and below's classify!" might be an announcement we'd need to put out.

I could see two possible ways of moving forward:

1. Add an `unsafe` module similar to how golang has the unsafe package. This could be used to house any internal modules deemed useful to some who understand the possible consequences of using them. There should be a part added to README saying that unsafe package has no guarantees for non-breaking changes, so if you rely on the unsafe package, do not use `^` for your version... also we won't offer support for unsafe modules on github. "You're on your own" etc.... but tbh at that point I'm sure the person would just fork it anyways. As your Pull Request shows, forking and adding the module to the exports is not a hard patch to maintain.
2. Try to migrate away from TransactionBuilder and remove the dependency on classify module. (This has been discussed in the past, but a good replacement for TransactionBuilder has yet to be made.)

I'll close this PR for now. But please let me know if you have any insights or ideas on how we might move forward.

If that idea involves a PR you can reopen this after pushing some changes or make a new PR.

Thanks again for your time! Appreciate it.

junderw:
I'm going to close this... since TransactionBuilder is the dependent, and in v6 we will remove it in favor for Psbt class.

Psbt has something similar implemented as below:

https://github.com/bitcoinjs/bitcoinjs-lib/blob/af8f41a1e54c2e5bb562b2e365bf41081fcc3ef0/ts_src/psbt.ts#L1404-L1410

https://github.com/bitcoinjs/bitcoinjs-lib/blob/af8f41a1e54c2e5bb562b2e365bf41081fcc3ef0/ts_src/psbt.ts#L753-L766

kklash:
Possible future solution would be to abstract script classification to a separate NPM package

bomzj:
A replacement for missing `bitcoin.script.classifyOutput`:
```js
const classifyOutputScript = script => {
  const isOutput = paymentFn => {
    try { 
      return paymentFn({ output: script }) 
    } catch (e) {
    }
  }

  if (isOutput(bitcoin.payments.p2pk)) return 'pubkey'
  else if (isOutput(bitcoin.payments.p2pkh)) return 'pubkeyhash'
  else if (isOutput(bitcoin.payments.p2ms)) return 'multisig'  
  else if (isOutput(bitcoin.payments.p2wpkh)) return 'witnesspubkeyhash'
  else if (isOutput(bitcoin.payments.p2sh)) return 'scripthash'
  
  return 'nonstandard'
}
```

PR #1378, opened 2019-04-07
Fix / improve tests

junderw:
- [x] Move _regtest.js to use async await + promises
- [x] Fix faucet to loop in the case that "Missing Unspent" error happens. up to 5 times. (lowers chance of races causing errors)
- [x] Fix error where regtest server isn't mined up to 432 blocks. (Returns "Bad Request" error for faucet)
- [x] Add APIURL env var for altering the regtest server endpoint. (Leaving the server up and public so people don't need docker to run integration tests)
- [x] Built a docker image with regtest server in it, synced to 432 blocks. Travis now uses that for each build, so no more races for Travis!!!!! Time of builds for integration is barely affected at all.
- [x] Switched to arrow functions throughout
- [x] Switched to strict versions of assert

----
Also moved to async/await + promises to make the fix easier (allow a do while loop)

### DIFF
--- a/test/integration/_regtest.js
+++ b/test/integration/_regtest.js
@@ -71,7 +71,7 @@ async function faucet (address, value) {
         }
       )
 
-    await sleep(randInt(50, 150))
+    await sleep(randInt(10, 40))
 
     const results = await unspents(address)
 
--- a/test/integration/_regtest.js
+++ b/test/integration/_regtest.js
@@ -10,7 +10,7 @@ const dhttp = options => new Promise((resolve, reject) => {
 })
 
 const APIPASS = process.env.APIPASS || 'satoshi'
-const APIURL = 'https://regtest.bitbank.cc/1'
+const APIURL = process.env.APIURL || 'https://regtest.bitbank.cc/1'
 const NETWORK = bitcoin.networks.testnet
 
 function broadcast (txHex) {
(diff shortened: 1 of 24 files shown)


junderw:
This should put an end to the "Why are my tests failing!? oh, integration server again :-P" problems with Travis.

test/integration/cltv.js
```diff
@@ -10,8 +10,8 @@ const bob = bitcoin.ECPair.fromWIF('cMkopUXKWsEzAjfa1zApksGRwjVpJRB3831qM9W4gKZs
 
 describe('bitcoinjs-lib (transactions w/ CLTV)', function () {
   // force update MTP
-  before(function (done) {
```
d-yokoi: How about migrating to arrow functions?

test/integration/_regtest.js
```diff
@@ -1,99 +1,129 @@
...
   })
 }
 
-function faucetComplex (output, value, callback) {
+async function faucet (address, value) {
+  let count = 0
+  let _unspents = []
+  const sleep = ms => new Promise((resolve, reject) => setTimeout(resolve, ms))
+  const randInt = (min, max) => min + Math.floor((max - min + 1) * Math.random())
+  do {
```
d-yokoi: I think you can do the same thing with a simple `while` statement.

PR #1382, opened 2019-04-11
Moving tests to use TS as well

junderw:
Goals:

- [x] Move tests to use TS without changing them too drastically. (-w diff of the JS files will show most of the changes are superficial)
- [x] Convert imports of bitcoinjs modules into imports that understand Types so that opening the tests will warn about incorrect arguments / return values.

This has already given insight into some improperly typed modules in bitcoinjs and BIP39 libraries.

Moving forward, having these tests being in TS will help catch type bugs when adding new modules.

@d-yokoi I basically turned off everything in the tsconfig. Let me know if you see any problems... the main goal is so that tests won't build if the tests are written under the wrong assumptions of what args are needed etc. (one issue I keep having is I forget to make later args optional)


junderw:
Also, one issue that may need fixing:

- Since the point of this is to be able to verify args and return values for the various functions in typescript... the data imported from the fixtures should probably be typed...

TODO: Create an interface for each type of fixture, and cast the imported JSON against it so that functions can see what they're taking in.

junderw:
@trungst368 Help with what?

junderw:
`sigh` I will have to retry this later

PR #1385, opened 2019-04-15
Add low R value signing as option to ECPair.sign

junderw:
tiny-secp256k1 now exposes the extra data for nonce generation.

This allows us to grind for a low R value similar to Bitcoin Core.

Defaulting to off for now.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -731,9 +731,9 @@
       "dev": true
     },
     "nan": {
-      "version": "2.12.1",
-      "resolved": "https://registry.npmjs.org/nan/-/nan-2.12.1.tgz",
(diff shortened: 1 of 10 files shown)


junderw:
Also adding `setLowR` to TransactionBuilder to allow people to enable this feature with BIP32 and ECPair objects that support it.

ECPair support is in this PR, and BIP32 support is here: https://github.com/bitcoinjs/bip32/pull/20

junderw:
Which means maybe we should merge BIP32 first and update the dependency here.

junderw:
BIP32 merged and published in 2.0.3, have updated the dependency here too.

PR #1387, opened 2019-04-21
Use Prettier to make JS easier to read/audit

junderw:
Since we will be verifying the JS, this seems like a good step to perform.


package.json
```diff
@@ -15,12 +15,13 @@
...
   "scripts": {
-    "build": "npm run clean && tsc -p ./tsconfig.json",
+    "build": "npm run clean && tsc -p ./tsconfig.json && npm run formatjs",
     "clean": "rimraf src",
     "coverage-report": "npm run build && npm run nobuild:coverage-report",
     "coverage-html": "npm run build && npm run nobuild:coverage-html",
     "coverage": "npm run build && npm run nobuild:coverage",
     "format": "npm run prettier -- --write",
+    "formatjs": "npm run prettierjs -- --write > /dev/null 2>&1",
     "format:ci": "npm run prettier -- --check",
```
d-yokoi: What about CI?

junderw: good point...

junderw: Done.

PR #1388, opened 2019-04-23
fix bug:ECPair.verify should return boolean

jolestar:
fECPair.verify should return boolean, and js implements indeed return boolean.


junderw:
patched in 5.0.3

PR #1392, opened 2019-04-28
Export classify.ts

visvirial:
`classify.ts` is useful for classifying what types does sciript signature or script pubkey have.
However `classify.ts` is not exported in `index.ts` and we cannot use `classify.*` methods from external libraries or applications.
In this commit you will be able to use `bitcoin.classify.*` methods to identify your transaction script:
```
$ ts-node
> import * as bitcoin from 'bitcoinjs-lib';
> bitcoin.classify.output(new Buffer('410496b538e853519c726a2c91e61ec11600ae1390813a627c66fb8be7947be63c52da7589379515d4e0a604f8141781e62294721166bf621e73a82cbf2342c858eeac', 'hex'));
'pubkey'
```


junderw:
Duplicate of #1377

Please respond to the open pull request.

junderw:
also note you can use:

```typescript
import * as classify from 'bitcoinjs-lib/src/classify';
```

and it should work, but breaking changes might occur on patches, so don't use karrat versioning `^`

PR #1408, opened 2019-06-04
[WIP] Add support for BIP174 PSBT import/export

lukechilds:
Just opening this for feedback and review.

A few points:

**1. Fixture data may be unreliable**

Now I've got something working I've just copied the raw PSBT output from the fixture inputs and added them to the fixtures. They may well be inaccurate or missing data. I'm aware of this and will go back and check them all carefully later. I just want something to test against for now so I'll get notified by a failing test if I break something really badly.

**2. I think adding `fromPsbt()` to `TransactionBuilder` was a mistake.**

I think both `fromPsbt()` and `toPsbt()` belong as methods of the `Transaction` class. We could expose a convenience `fromPsbt()` method on `TransactionBuilder` too that just creates a `Transaction` internally and then returns a `TransactionBuilder` instance with `fromTransaction()`.

**3. Theres some missing functionality**

I'm aware the following functionality isn't working yet:

- Signature/Witness data when exporting to PSBT
- Importing a PSBT with data like BIP39 derivation paths and then exporting to PSBT will lose the BIP39 data

## Example

```js
const unsignedTxHex = '0200000001268171371edff285e937adeea4b37b78000c0566cbb3ad64641713ca42171bf60000000000feffffff02d3dff505000000001976a914d0c59903c5bac2868760e90fd521a4665aa7652088ac00e1f5050000000017a9143545e6e33b832c47050f24d3eeb93c9c03948bc787b32e1300';
const unsignedTx = Transaction.fromHex(unsignedTxHex);
const psbt = unsignedTx.toPsbt();
console.log(psbt);
// '70736274ff0100750200000001268171371edff285e937adeea4b37b78000c0566cbb3ad64641713ca42171bf60000000000feffffff02d3dff505000000001976a914d0c59903c5bac2868760e90fd521a4665aa7652088ac00e1f5050000000017a9143545e6e33b832c47050f24d3eeb93c9c03948bc787b32e130000000000'

const unsignedTxFromPsbt = TransactionBuilder.fromPsbt(psbt).buildIncomplete();
console.log(unsignedTx.toHex() == unsignedTxFromPsbt.toHex());
// true
```

@junderw Let me know if you've got any feedback. Looking ok so far? Do you agree that it would make more sense to move all the PSBT logic to the `Transaction` class?


junderw:
Re: 2

A Transaction class should not have extraneous data that is not a part of a Bitcoin Transaction.

You don't have to copy paste TransactionBuilder, but there needs to be an extending class and/or encapsulating class, and since PSBT is made for maintaining state between multiple signatures of multisig / signing offline using HSMs that need more data to verify the data being signed, it makes sense that whatever class represents the PSBT concept in the library should also have "add input" "add output" "sign" etc. methods...

Currently, the closest thing to what I have described is TransactionBuilder.

You can build something from scratch if you really want, but imo copy pasting TXB and adding some extra fields to hold the extra PSBT data (so that fromPSBT -> toPSBT looping will never lose data) is sufficient for now.

The only point where I would think that we might need to propose modifying the BIP would be if we decide we REEEEALLY need some extra data for some use of our TXBv2 case that not currently included in the PSBT spec

But yeah, as for mixing PSBT with the Transaction class, that doesn't seem wise.

lukechilds:
After our discussion on Keybase:

- [ ] Use Base64 for PSBT input/ouput
- [ ] Clone TXB to TXBV2 and do all new work there
- [ ] All PSBT methods should be on TXBV2, not TX
- [ ] `buildIncomplete()` deprecated in favour of PSBT serialisation
- [ ] `isComplete()` method to return if all inputs have signatures
- [ ] Add `inputCount` getter method

lukechilds:
Opened a new PR over at https://github.com/bitcoinjs/bitcoinjs-lib/pull/1425 as this is gonna be totally different with the new BIP174 API and this branch now has lots of conflicts with master.

PR #1414, opened 2019-06-10
split sign method into two separate api to support hardware wallet

Ykid:


ts_src/transaction_builder.ts
```diff
@@ -264,7 +264,52 @@ export class TransactionBuilder {
...
       );
     }
+    return signatureHash;
+  }
+
+  attachSignature(
+    vin: number,
+    keyPair: ECPairInterface,
+    signature: Buffer,
+    hashType?: number,
```
junderw: `signatureHash` that was returned from `createSignHash` already has the hashType embedded into it.

If you use hashTypeA for `createSignHash` and hashTypeB for `attachSignature` the signature will not verify properly.

There should be a verify check here to prevent such a case.

ts_src/transaction_builder.ts
```diff
@@ -264,7 +264,52 @@ export class TransactionBuilder {
...
+  }
 
+  private __sign(
+    vin: number,
+    keyPair: ECPairInterface,
+    signatureResolver: () => Buffer,
+    hashType?: number,
+  ): void {
+    hashType = hashType || Transaction.SIGHASH_ALL;
+    if (!this.__INPUTS[vin]) throw new Error('No input at index: ' + vin);
```
junderw: All these checks should be in `createSignHash`...

ts_src/transaction_builder.ts
```diff
@@ -264,7 +264,52 @@ export class TransactionBuilder {
...
+    const signatureResolver = (): Buffer =>
+      keyPair.sign(hashToSign, this.__USE_LOW_R);
+    this.__sign(vin, keyPair, signatureResolver, hashType);
+  }
 
+  private __sign(
+    vin: number,
+    keyPair: ECPairInterface,
+    signatureResolver: () => Buffer,
+    hashType?: number,
```
junderw: private methods don't need possibly undefined params.

We should always explicitly define these params.

ts_src/transaction_builder.ts
```diff
@@ -196,14 +196,14 @@ export class TransactionBuilder {
     return this.__build(true);
   }
 
-  sign(
+  createSignHash(
     vin: number,
     keyPair: ECPairInterface,
     redeemScript?: Buffer,
     hashType?: number,
```
junderw: You should define a default hashType for `createSignHash` as well.

junderw:
@Ykid Actually, I'm going to go ahead and close this.

In the next minor version update we plan to add async versions of all signing operations, including txb.sign.

This will allow you to pass in an interface that looks like below instead of the ECPair:

```ts
interface Signer {
  publicKey: Buffer; // 33 byte compressed DER pubkey Buffer
  sign(hash: Buffer): Promise<Buffer>; // async function to sign 32 byte hash
  // with the private key associated with publicKey and return 64 byte R+S sig data
}
```

Then your sign method could just contain the hardware sign function (assuming it is async)

Ykid:
@junderw may I have a rough idea about when the release will be out ? Knowing that OSS project might not have enough resource and contributors maintains the project in their leisure time, it seems a bit inappropriate to ask so. I just want some more information in order to judge whether it is better to wait or use something else for the time being instead. This is not to urge you guys. thanks !

PR #1416, opened 2019-06-12
Migrate to stricter type checks during sign

### LINKED ISSUE #1412: TransactionBuilder sign method should require explicitly stating the type of script
Ping @lukechilds (re: the new PSBT thing)

TransactionBuilder's biggest FAQ is "I signed, but it didn't work" which ends up being that they didn't pass the correct data, and the sign method "guesses" their script type based on the data passed to sign.

TODO:

1. Make sign args use a one-object paradigm.
2. Require a `type` string which is one of :
```
// Raw
'p2pkh', 'p2pk', 'p2wpkh', p2ms', 

// P2SH wrapped
'p2sh-p2wpkh', 'p2sh-p2ms',

// P2WSH wrapped
'p2wsh-p2ms',  

// P2SH-P2WSH wrapper
'p2sh-p2wsh-p2ms'
```
3. Depending on the type, do a runtime check of the arg attributes to make sure the proper information is included, and it passes sanity checks.

Currently, people looking at an example for P2PKH signing will think it can be used for P2SH-P2WPKH signing etc etc...

It is confusing, granted.

This is more of a long term goal. Hopefully we can add it in a minor version and deprecate the old way of doing things next major version.

junderw:
Fixes #1412 

Need to add tests.

### DIFF
--- a/ts_src/transaction_builder.ts
+++ b/ts_src/transaction_builder.ts
@@ -16,6 +16,21 @@ const typeforce = require('typeforce');
 
 const SCRIPT_TYPES = classify.types;
 
+const PREVOUT_TYPES: Set<string> = new Set([
+  // Raw
+  'p2pkh',
+  'p2pk',
+  'p2wpkh',
+  'p2ms',
+  // P2SH wrapped
+  'p2sh-p2wpkh',
+  'p2sh-p2ms',
+  // P2WSH wrapped
+  'p2wsh-p2ms',
+  // P2SH-P2WSH wrapper
+  'p2sh-p2wsh-p2ms',
+]);
+
 type MaybeBuffer = Buffer | undefined;
 type TxbSignatures = Buffer[] | MaybeBuffer[];
 type TxbPubkeys = MaybeBuffer[];
@@ -48,6 +63,16 @@ interface TxbOutput {
   pubkeys?: TxbPubkeys;
   signatures?: TxbSignatures;
   maxSignatures?: number;
+}
+
+interface TxbSignArg {
+  prevOutScriptType: string;
+  vin: number;
+  keyPair: ECPairInterface;
+  redeemScript?: Buffer;
+  hashType?: number;
+  witnessValue?: number;
+  witnessScript?: Buffer;
 }
 
 function txIsString(tx: Buffer | string | Transaction): tx is string {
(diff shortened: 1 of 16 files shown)


junderw:
ToDo: add `prevOutScriptType` to all signs in the fixtures (ughhhh)

junderw:
@lukechilds @jprichardson @d-yokoi 

This is done.

However, I think some points for review/discussion:

1. Implementation details.
2. I changed some tests because certain errors were caught earlier by my additions. Should I create a way to bypass checks just so that we can test the failsafes later down the line? I wonder if there's a way to prevent consumers from using that bypass?
3. console.warn... yes or no? (I'd just stick with yes)
4. Documentation... how do we relay the new usage paradigm? README? CHANGELOG? COMMENTS IN TESTS? all of the above?

Questions, comments, anything is appreciated.

ts_src/transaction_builder.ts
```diff
@@ -197,13 +236,36 @@ export class TransactionBuilder {
...
+      );
+      vin = signParams;
+    } else if (typeof signParams === 'object') {
+      checkSignArgs(this, signParams);
+      vin = signParams.vin;
+      keyPair = signParams.keyPair;
+      redeemScript = signParams.redeemScript;
+      hashType = signParams.hashType;
+      witnessValue = signParams.witnessValue;
+      witnessScript = signParams.witnessScript;
```
lukechilds: This block:

```js
keyPair = signParams.keyPair;
redeemScript = signParams.redeemScript;
hashType = signParams.hashType;
witnessValue = signParams.witnessValue;
witnessScript = signParams.witnessScript;
```

Could be written as:

```js
({
    keyPair,
    redeemScript,
    hashType,
    witnessValue,
    witnessScript,
} = signParams);
```

to reduce code duplication.

junderw: I thought that only worked with let or const, and these were already declared... let me try it out and if it works I'll push it

junderw: yup, it worked. TIL! Thanks.

lukechilds: It works with with predefined variables if you wrap it in brackets.

e.g:

No brackets for declaring new variables:

```js
const {
	vin,
	keyPair,
	redeemScript,
	hashType,
	witnessValue,
	witnessScript,
} = signParams;
```

Variables are predefined so we need to wrap the expression in brackets.

```js
let vin;
let keyPair;
let redeemScript;
let hashType;
let witnessValue;
let witnessScript;

({
	vin,
	keyPair,
	redeemScript,
	hashType,
	witnessValue,
	witnessScript,
} = signParams);
```

lukechilds:
>console.warn... yes or no? (I'd just stick with yes)

I think `console.warn` is fine. There's not really any other way to soft deprecate without official docs.

>Documentation... how do we relay the new usage paradigm? README? CHANGELOG? COMMENTS IN TESTS? all of the above?

README is pretty bare, I don't think there's much benefit putting it there. Definitely would in the CHANGELOG.

I think most people use the tests as documentation so I'd go with comments in tests explaining when the old behaviour will be deprecated.

Just to clarify, this PR allows passing a list of args **or** an object, right? So it will accept either? And just give a deprecation warning if an arg list is used. Is that correct?

Also, it looks like you've replaced all the arg list tests to only use an object as an arg.

I'd personally have both argument types in the tests.

Have the old args list style in the tests with clear comments that this pattern is deprecated and will be removed soon, then reference the object style tests as the correct usage.

We really need an official docs site though...

lukechilds:
And man, the object as an arg really does make the code so much more readable...

junderw:
> I'd personally have both argument types in the tests.

Maybe not all, but since most instances of txb.sign are in some forEach of fixtures... maybe I'll run over one of the fixtures twice (once with the old way)... `test/transaction_builder.js` seems fine candidate.

payments and integrations... meh, I'll leave as is.

lukechilds:
Yeah, that should be fine.

I just mean if we're treating tests as docs it's not good to completely remove stuff.

If a user had referenced the tests as documentation previously and then came back to the docs (tests) in the future and it's all changed but their code still works, it's gonna cause confusion. If I can see the syntax I'm familiar with annotated with clear comments saying it's deprecated and will be removed, and to use this other syntax, it's much clearer.

But yeah, it is kind of a pain having to do weird test stuff just because there aren't any official docs.

I know nobody likes writing docs but I really think a proper docs site should be a high priority.

junderw:
@lukechilds I just decided to run TransactionBuilder unit test twice, once with new args and once with old args.

I also refactored sign to make it smaller, and easier to create a new async version down the road without copy pasting a huge amount. DRY.

ts_src/transaction_builder.ts
```diff
@@ -197,92 +236,28 @@ export class TransactionBuilder {
   }
 
   sign(
-    vin: number,
-    keyPair: ECPairInterface,
+    signParams: number | TxbSignArg,
```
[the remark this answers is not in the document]
junderw: We'll go ahead and change this in another PR.

ts_src/transaction_builder.ts
```diff
@@ -976,3 +951,354 @@ function canSign(input: TxbInput): boolean {
...
+  ourPubKey: Buffer;
+  keyPair: Signer;
+  signatureHash: Buffer;
+  hashType: number;
+  useLowR: boolean;
+}
+
+type HashTypeCheck = (hashType: number) => boolean;
+
+function getSigningData(
```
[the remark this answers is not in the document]
junderw: We'll refactor in another PR

PR #1421, opened 2019-06-15
Use regtest-client

junderw:
Separating it out since it was useful for other projects.


junderw:
Github:
https://github.com/bitcoinjs/regtest-client

NPM:
https://www.npmjs.com/package/regtest-client

PR #1424, opened 2019-06-24
Create async sign interface for TransactionBuilder

junderw:
Separated all the ECC operations out into a single file to make it easier to swap in other methods (WASM, hardware wallets, etc.)

Having this new signAsync method will allow people to create SignerAsync interfaces that use HW wallets etc. to sign.

Changed the target to ES2017 to allow for async await in the JS.
async await has been supported since Node v8.x.x and in browsers for quite a while...

This is just an idea. Extended discussion expected.

### DIFF
--- a/src/ecc.js
+++ b/src/ecc.js
@@ -1,0 +1,4 @@
+'use strict';
+Object.defineProperty(exports, '__esModule', { value: true });
+const ecc = require('tiny-secp256k1');
+exports.ecc = ecc;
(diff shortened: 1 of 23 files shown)


junderw:
Creating a new interface for ECC operations to help people fork this library seems like a meh---OK ish idea...

But adding new interfaces to TXB when we have a Psbt class which supports async and is better in every way just seems like a waste of time.

Closing.

PR #1425, opened 2019-06-24
[WIP] Add PSBT support

lukechilds:
Replaces: #1408 

Todo:

- [ ] Implement Input Finaliser
- [ ] Implement Signer
- [ ] Implement Transaction Extractor


lukechilds:
Failing builds are just from low coverage (89.9% instead of 90% 🤦‍♂️), all unit tests are passing.

lukechilds:
All BIP174 test cases are passing 🎉 

Still failing due to low coverage, I'll add some more tests for the helper methods to get coverage up.

junderw:
Looking through TransactionBuilder for features that would port over well:

A maximum sat per byte fee rate just set as a stop-gap. TransactionBuilder has it set at 2500 sat/byte which even in Jan. 2018 it didn't go THAT high... but tbh it's only there to stop people from accidentally sending 500 BTC to a miner etc.

In which case 5k or even 50k sat/byte would be a good stop-gap.

The question is:

1. TransactionBuilder only had one way to export, and that was through build -> Transaction.toHex() etc.
2. The fee check only happened during build because of this.
3. PSBT is much more lenient and could be exported from our fee-checking area and parsed/signed in an environment without fee checking.

Off the top of my head:

1. Set default to 10k sat/byte
2. Check when doing: toBuffer, sign, finalize, extract
3. Offer a way to change the barrier. And in the Error message say the actual fee rate, so they know how high to raise it to ignore the error.

This also got me thinking: There's no way to encode the network or maxSatPerByte values...

Although I guess that could be just hard coded into the apps that use it. We shouldn't really NEED to have the info encoded, since the only time they are used is when adding an output with an address, and checking the max satoshi per byte...

junderw:
The integration tests are awesome. very clean compared to Transaction Builder.

A couple ideas I had:

1. a sign function that searches all inputs for your public key and signs all inputs that apply.
2. a verify input function that will take an inputIndex and a public key buffer, it will search the input for a partialSig of that pubkey, and if present will proceed to verify the signature returning true or false.

lukechilds:
Woah, those integration tests are great!

>1. a sign function that searches all inputs for your public key and signs all inputs that apply.

Yeah, I like this, the pseudo code in the BIP works like this and I've seen other implementations work this way too.

>2. a verify input function that will take an inputIndex and a public key buffer, it will search the input for a partialSig of that pubkey, and if present will proceed to verify the signature returning true or false.

This also sounds very handy.

junderw:
Coverage passing! (barely)

junderw:
A few more checks we need: (I keep adding more coverage branches!!!)

1. Make sure the sighashtype for signing is within a whitelist. Allow that whitelist to be passed as an arg to sign methods. (see below &1)
2. When finalizing, fail to finalize if the partialsig(s) are not the same sighash as the sighash given for the input.

&1

1. I am a signer, I get a PSBT that is ready to sign. I parse. I sign according to the PSBT as-is.
2. I notice my UTXO was stolen by a hacker because they changed my PSBT input's sighashtype to SIGHASH_ANYONECANPAY | SIGHASH_NONE and after the fact they changed the outputs to send to themselves, and added an input they signed with SIGHASH_ALL.
3. I lose the BTC in my UTXO.

We can mitigate by making default whitelist be `[ SIGHASH_ALL ]` only, and maybe we can offer a couple whitelist patterns as a static class getter etc. (like Transaction.SIGHASH_ALL in transaction.ts etc)

ts_src/psbt.ts
```diff
@@ -0,0 +1,1160 @@
...
+import * as payments from './payments';
+import * as bscript from './script';
+import { Output, Transaction } from './transaction';
+
+const DEFAULT_OPTS: PsbtOpts = {
+  network: btcNetwork,
+  maximumFeeRate: 5000, // satoshi per byte
+};
+
+export class Psbt extends PsbtBase {
```
d-yokoi: I think `Psbt` does not have to inherit `PsbtBase`.
Having a `PsbtBase` instance as a property and calling methods of it is enough to manage PSBT format, isn't it?
That also complies with the [Composition over inheritance](https://en.wikipedia.org/wiki/Composition_over_inheritance) principle.

junderw: I'll try it out and see.

junderw: I tried it out... it's kind of long... :-/

junderw:
Ok. I think this is good:

Todo:

1. Merge https://github.com/bitcoinjs/bip174/pull/11
2. Publish bip174 as 1.0.0
3. Update this branch to use 1.0.0

Then we can merge this.

Once we merge this and a few other PRs, let's bump minor version.

Issue #1431, opened 2019-07-08
Dust transaction error or warning


YegorZaremba:
- [x] Bug
- [x] Scope: Transaction builder

When I build my transaction to send small amount of satoshi (< 5460) `bitcoinjs-lib` generates me transaction raw without any errors and warnings. But when I try to push generated transaction raw via [blockchain](https://www.blockchain.com/btc/pushtx) or [blockchyper](https://live.blockcypher.com/btc/pushtx/) I get `dust transaction error`

I found [this PR of bitcore](https://github.com/bitpay/bitcore/pull/916/files#diff-3ff19b4e31a68503f7ddd89615e606d7R146). It doesn't allow to send transactions which value is smaller than 5460 Satoshi (a.k.a. dust)

So, I think it will be better for thousand of users to show warning or throw error about dust transaction (which won't be pushed anyway). Because now it's not obvious

Related issues in `bitcoinjs-lib`: [#1123](https://github.com/bitcoinjs/bitcoinjs-lib/issues/1123), [#951](https://github.com/bitcoinjs/bitcoinjs-lib/issues/951)

Related links: [link1](https://bitcoin.stackexchange.com/questions/14244/why-is-a-transaction-with-outputs-of-less-than-%E0%B8%BF0-00005460-rejected), [link2](https://bitcoin.stackexchange.com/questions/10986/what-is-meant-by-bitcoin-dust)

junderw:
4 year old pull request?...

Dust is depending on network. Also in Bitcoin since over a year ago dust is dynamic.

BlockCypher made a new test network, are you trying to use that?

That is not supported.

junderw:
https://bitcoin.stackexchange.com/a/78097/87485

Here's the most recent "base level"

so 444 satoshis should send.

It is possible blockchain and others are using 4 year old software, but that is not our problem.

junderw:
IMO, dust should be a network parameter, but since it changes so frequently, it is not good to add it there.

So wallet makers should set their own dust limit based on what limit the API they use will allow.

Closing. Thanks for the report.

junderw:
https://github.com/bitcoin/bitcoin/blob/v0.18.0/src/policy/policy.cpp#L18

huh, so:

non-segwit: 546 satoshis

segwit p2wpkh: 294 satoshis

segwit p2wsh: 330 satoshis

PR #1432, opened 2019-07-08
fix(txBuilder): dust error

YegorZaremba:
Fixes: [#1431](https://github.com/bitcoinjs/bitcoinjs-lib/issues/1431)

There are some failed unit tests, if you interesting in these changes, I can fix these failed tests


junderw:
see #1431

PR #1433, opened 2019-07-11
Add name attribute to Payments

junderw:


junderw:
This will help with testing, logging etc.

PR #1455, opened 2019-08-13
fix(transaction_builder.js): change incorrect object compare

MehrdadKhnzd:
change keyPair.network and network compare to JSON.stringify() compare

### DIFF
--- a/src/transaction_builder.js
+++ b/src/transaction_builder.js
@@ -1007,7 +1007,7 @@ function getSigningData(
     throw new Error('sign requires keypair');
   }
   // TODO: remove keyPair.network matching in 4.0.0
-  if (keyPair.network && keyPair.network !== network)
+  if (keyPair.network && JSON.stringify(keyPair.network) !== JSON.stringify(network))
     throw new TypeError('Inconsistent network');
   if (!inputs[vin]) throw new Error('No input at index: ' + vin);
   hashType = hashType || transaction_1.Transaction.SIGHASH_ALL;


lukechilds:
I'm assuming this is to fix the case where the `TransactionBuilder` network object and `ECPair` network object aren't strictly equal but have the same values? For example if the objects were manually created or returned from a function before being passed in to each of the constructors, is that correct?

If so, using `JSON.stringify` for this is not a good solution. When following the [ECMAScript spec](https://tc39.es/ecma262/#sec-ordinaryownpropertykeys) (which not all engines do) each key that is a string will be listed in order of property creation. This means that two objects with identical values but with properties that were added in different orders will not create identical JSON output.

e.g:

```js
JSON.stringify({messagePrefix: 'foo', bech32: 'bar'})
// "{"messagePrefix":"foo","bech32":"bar"}"

JSON.stringify({bech32: 'bar', messagePrefix: 'foo'})
// "{"bech32":"bar","messagePrefix":"foo"}"
```

Those two values are deeply equal to each other but do not produce the same JSON output. If you want to accurately compare the value of the two objects you should use a deep equal comparison function.

Also, the failing tests are due to failed formatting rules. You can run `npm run formatjs` to automatically fix this.

junderw:
1. Use the same network for both. It will be the same. Why would you use two different references to the same data?
2. To pass tests, modify the TypeScript, run `npm run format` then `npm run lint` then `npm run build` which will give you the modified JS. commit both.

Regardless, TransactionBuilder is deprecated, and the workaround is easy (use the same network object reference)

NACK

Thanks for the contribution though.

Try out Psbt as it is much more versatile.

MehrdadKhnzd:
The problem is that I extracted bitcoin js networks into other classes with extra information, and when I used them they didn't have the same references. ‌By the way, I guess I have to try the new library. So, you say the fix is not required? Or you suggest me to consider fixing it in the correct way?

junderw:
You can use your same extended class for all ECPair functions and TransactionBuilder.

I am guessing you are specifying your extended class for one and not the other?

Can you show me your extended class and what's in it?

Also, the network check for the sign method is almost useless, but removing it requires a major version bump and currently we are planning on removing TransactionBuilder in the next major version.

You could maintain a patch removing the check.

Or you can tell me what your network object contains and we can figure out what's going on.

Or you can switch from TransactionBuilder to the much better Psbt class.

Let me know what you are up to and I'll answer any questions.

MehrdadKhnzd:
Thanks for your support! I decided to switch to Psbt class, but I have a problem with parameters, due to my lack of knowledge. How can I get all of the required parameters for `nonWitnessUtxo` from insights like `https://test-insight.bitpay.com/api/`?

junderw:
nonWitnessUtxo is just the full raw transaction that holds the output you are spending.

1. get the unspents from insight (txid, output_index, amount, script_pubkey etc.)
2. if the output you are spending is not segwit, also grab the full raw hex of the unspent's transaction (from the unspent txid)
3. if the output is segwit, create a witnessUtxo object with the scriptPubkey and value.

iirc the API endpoint is `GET /api/rawtx/TXIDHERE`

be sure to convert hex to buffer

MehrdadKhnzd:
I'm stuck. I don't know if I have to open another issue for my questions or not, but here's my problem. When I was using the `TransactionBuilder`, the results look inconsistent to me. Sometimes I simply couldn't broadcast the transaction with errors like `The transaction is already spent` or 'Missing input` or others. It seemed like I was missing something. Then I tried to switch to the `Psbt` class, and I got another problem. With `TransactionBuilder`, sometimes I couldn't broadcast my transaction, but with `Psbt`,  I can't even sign it. It gives me `Cannot sign for input with key` error. I don't know what I'm missing. Based on your advice, here's my approach:
1. I get all of my unspent transactions from all of my addresses.
2. I select which unspent transactions fit my desired amount.
3. I add those transactions with their related address wif to an array
4. I add inputs like this:
`
const psbt = new BitcoinLib.Psbt({
  network: currentNetwork
})
for (let k in transactions) {
  const rawTx = await getRawTransaction(transactions[k].txid)
  psbt.addInput({
    hash: transactions[k].txid,
    index: Number(k),
    nonWitnessUtxo: Buffer.from(
      rawTx,
      'hex'
    )
  })
}
`
5. And then I try to sign it this way:
`
for (let k in transactions) {
  const key = BitcoinLib.ECPair.fromWIF(transactions[k].wif, currentNetwork)
  psbt.signInput(Number(k), key)
  psbt.validateSignaturesOfInput(Number(k))
}
`
And it gives me sign error. My old approach with `TransactionBuilder` was also similar to this, although I couldn't get desired results but it was fine in signing inputs. This code is simplified a little bit, but the general idea is the same. What's my problem?

junderw:
Number(k) is not the index.

transactions[k].vout is the index.

MehrdadKhnzd:
At first, I did so, but as it gave me `No input at index 1` with both classes, I changed it to `Number(k)` and it worked in most situations (which now I know was just some sort of luck), but now I double checked everything and I found out where my problem really was, and with `Psbt` class and `transactions[k].vout` it now works really fine! Thanks again for your help, It's really appreciated! I guess that the problem's completely solved. I let you know if I have any other problems. Thanks again!

PR #1461, opened 2019-08-23
[PSBT] Allow nonWitnessUtxo with segwit

junderw:
Also fixes a bug.

One fixture for finalizeAllInputs only had a finalScriptSig when it should have also had a finalWitnessScript...

strangely, we test the same script in the integration tests, but the integration tests passed.... ????

This fails currently, and needs a bip174 fix upstream to allow for this.


junderw:
Also, the reason why we should allow this is that segwit needs less info, but nonWitnessUtxo has more info, so why reject it when we can sign it?

This way wallet apps could decide to code it so that all inputs have nonWitnessUtxo just in case... or if they don't know what the redeemscript of a P2SH prevoutScript is they can just err on the side of caution and stick the nonWitnessUtxo in it.

PR #1465, opened 2019-08-30
WIP: Move tests to TypeScript

junderw:
Once we are done... I think we should not commit test JS files...

But for now, just to check that we don't break the tests I will commit JS files along the way, and then once we're all done, I'll just squash all the commits.

@d-yokoi ...... does this look right so far?


d-yokoi:
I prefer `build:test` as a npm script name, but everything else looks good to me 👍

junderw:
Done... but coverage isn't working... someone help please?

junderw:
I collapsed to 2 commits:

1. name change to .spec.ts
2. Everything else.

If you look at the second commit it is easy to see the diff of each test file from js to ts.

.mochaTsMixin.js
```diff
@@ -0,0 +1,3 @@
+require("ts-node").register({
+  project: "test/tsconfig.json",
+});
```
d-yokoi: I cannot see why the name starts with `.` and the file is placed in the root directory.
`test/ts-node-register.js` makes more sense to me.

junderw: I wonder if it needs the js extension...

I have gitignore and clean set to test/**/*.js

d-yokoi: I see.
How about adding `!test/ts-node-register.js` to .gitignore or create a `test/dist` directory for build output?

junderw: It works without any extensions, so I have renamed it.

test/integration/_regtest.spec.ts
```diff
@@ -0,0 +1,6 @@
+import { RegtestUtils } from 'regtest-client';
+
+const APIPASS = process.env.APIPASS || 'satoshi';
+const APIURL = process.env.APIURL || 'https://regtest.bitbank.cc/1';
+
+export const regtestUtils = new RegtestUtils({ APIPASS, APIURL });
```
d-yokoi: The file name should not be suffixed with `.spec` since this file is not supposed to be executed by a test runner.

test/integration/payments.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+  bitcoin.ECPair.makeRandom({ network: NETWORK }),
+  bitcoin.ECPair.makeRandom({ network: NETWORK }),
+];
+console.warn = () => {}; // Silence the Deprecation Warning
+
+async function buildAndSign(
+  depends: any,
+  prevOutput: any,
+  redeemScript: any,
+  witnessScript: any,
```
d-yokoi: This type definition must be improved in the near future.

test/payments.utils.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+      // recurse
+      r = r[k];
+      object = object[k];
+    } else {
+      r[k] = object[k];
+    }
+  });
+
+  return result;
+}
```
d-yokoi: This file is just an util file, so `.spec` is not appropriate.

test/transaction_builder.spec.ts
```diff
@@ -0,0 +1,947 @@
...
+    });
+  });
+
+  return txb;
+}
+
+function construct(
+  f: any,
+  dontSign?: any,
+  useOldSignArgs?: any,
```
d-yokoi: Must be improved sometime.

test/tsconfig.json
```diff
@@ -0,0 +1,39 @@
...
+    "module": "commonjs",
+    "outDir": "../",
+    "declaration": false,
+    "rootDir": "../",
+    "rootDirs": [
+      "../src",
+      "../types"
+    ],
+    "types": [
+      "node"
```
d-yokoi: It is common practice to add `mocha` here so you don't have to import functions like `describe` or `it` from mocha in every spec file.

.mochaTsMixin.js
d-yokoi: It may work, but I don't think it's a good solution because newcomers will wonder why this file does not have an extension.

junderw:
I have fixed some of the main maintenance issues.
For type defs of helper functions etc. This can be done later.

The primary reason for moving to TypeScript for tests is so that when we write a new test we can make sure it runs correctly with TypeScript beforehand.

If we ever add tests to TxB (we probably won't since it will be removed soon) or payments (very possible) we can look at re-typing the helper functions.

dhruvbansal:
This commit broke many of the links in the README?

junderw:
d'oh!

Sorry about that! One second!

you21979:
#1474 
this pull request fix readme.

PR #1476, opened 2019-09-12
Remove TransactionBuilder from tests (besides transaction_builder)

### LINKED ISSUE #1475: TODO: Tests: Replace TransactionBuilder with Psbt
For every instance of TransactionBuilder *OUTSIDE OF transaction_builder.ts* we should replace it with PSBT.

junderw:
Fixes #1475


test/psbt.spec.ts
```diff
@@ -14,7 +14,7 @@ const initBuffers = (object: any): typeof preFixtures =>
     const data = result[1];
     const encoding = result[2];
 
-    return Buffer.from(data, encoding);
+    return Buffer.from(data, encoding as any);
```
d-yokoi: Really?

junderw: yes. `Buffer.from` now only accepts hardcoded strings... so we can not take it dynamically.

PR #1487, opened 2019-10-01
Add sanity check for arguments

### LINKED ISSUE #1486: "Error adding output." - uninformative error logs with psbt structure.
I'm trying to create a raw bitcoin tx using this library, a simple CTLV p2sh where I fund the p2sh address. Yet, I am getting uninformative errors about why it can't add this final output.
```
  const { address } = bitcoin.payments.p2sh({
    redeem: {
      output: redeemScript,
      network: network
    },
    network: network
  });

  const psbt = new bitcoin.Psbt();
  unspentOutputs.forEach(output => {
    if (output.length > 0) {
      let splitOutput = output.split('-');
      psbt.addInput(idToHash(splitOutput[0]), Number(splitOutput[1]), 0xfffffffe);
    }
  });
  // Send amount of satoshis to the P2SH time lock transaction
  psbt.addOutput(address, Number(amount));
```

The error I get is:
```
(node:74229) UnhandledPromiseRejectionWarning: Error: Error adding output.
    at PsbtTransaction.addOutput (/Users/drewstone/code/commonwealth/supernova-lockdrop/node_modules/bitcoinjs-lib/src/psbt.js:555:13)
    at Psbt.addOutput (/Users/drewstone/code/commonwealth/supernova-lockdrop/node_modules/bip174/src/lib/psbt.js:103:31)
    at Psbt.addOutput (/Users/drewstone/code/commonwealth/supernova-lockdrop/node_modules/bitcoinjs-lib/src/psbt.js:174:15)
```

junderw:
Fixes #1486

### DIFF
--- a/src/psbt.js
+++ b/src/psbt.js
@@ -143,6 +143,17 @@ class Psbt {
     return this;
   }
   addInput(inputData) {
+    if (
+      arguments.length > 1 ||
+      !inputData ||
+      inputData.hash === undefined ||
+      inputData.index === undefined
+    ) {
+      throw new Error(
+        `Invalid arguments for Psbt.addInput. ` +
+          `Requires single object with at least [hash] and [index]`,
+      );
+    }
     checkInputsForPartialSig(this.data.inputs, 'addInput');
     const c = this.__CACHE;
     this.data.addInput(inputData);
@@ -163,6 +174,17 @@ class Psbt {
     return this;
   }
   addOutput(outputData) {
+    if (
+      arguments.length > 1 ||
+      !outputData ||
+      outputData.value === undefined ||
+      (outputData.address === undefined && outputData.script === undefined)
+    ) {
+      throw new Error(
+        `Invalid arguments for Psbt.addOutput. ` +
+          `Requires single object with at least [script or address] and [value]`,
+      );
+    }
     checkInputsForPartialSig(this.data.inputs, 'addOutput');
     const { address } = outputData;
     if (typeof address === 'string') {
--- a/ts_src/psbt.ts
+++ b/ts_src/psbt.ts
@@ -182,6 +182,17 @@ export class Psbt {
   }
 
   addInput(inputData: PsbtInputExtended): this {
+    if (
+      arguments.length > 1 ||
+      !inputData ||
+      inputData.hash === undefined ||
+      inputData.index === undefined
+    ) {
+      throw new Error(
+        `Invalid arguments for Psbt.addInput. ` +
+          `Requires single object with at least [hash] and [index]`,
+      );
+    }
     checkInputsForPartialSig(this.data.inputs, 'addInput');
     const c = this.__CACHE;
     this.data.addInput(inputData);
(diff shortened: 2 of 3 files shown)


junderw:
I guess we should use typeforce...

This is just a quick fix.

ts_src/psbt.ts
```diff
@@ -182,6 +182,17 @@ export class Psbt {
   }
 
   addInput(inputData: PsbtInputExtended): this {
+    if (
```
lukechilds: Why is this check required?

Shouldn't TypeScript throw for us if `inputData` doesn't match the `PsbtInputExtended` interface?

junderw: TypeScript is only a compile-time check (transpile-time actually).

Some users don't use TypeScript, so we should do runtime checks as well.

lukechilds: Ah, I'm with you.

I wonder why TypeScript don't add this error checking automatically for non TypeScript consumers.

They could just represent all interfaces as JavaScript objects and then inject a checking function as the fist line of the body of all functions.

e.g:

```js
addInput(inputData) {
  checkTsInterface(inputData, PsbtInputExtended);
```

src/psbt.js
```diff
@@ -163,6 +174,17 @@ class Psbt {
     return this;
   }
   addOutput(outputData) {
+    if (
+      arguments.length > 1 ||
+      !outputData ||
```
lukechilds: Why truthy check here but undefined check below?

src/psbt.js
```diff
@@ -163,6 +174,17 @@ class Psbt {
...
+      outputData.value === undefined ||
```
lukechilds: Just a personal preference but I'd always opt for `typeof foo === 'undefined'` over `foo === undefined`.

Two reasons off the top of my head:

1. If variable is not defined typeof will still work, strict comparison will throw. (not an issue here with object properties but nice to be consistent)
2. `undefined` value can be redefined as something else causing this check to fail.

Both pretty pedantic reasons that clearly won't affect this code so totally understand if you wanna stick with a strict equality check.

Just my 2c as to why I'd always avoid `=== undefined`.

src/psbt.js
junderw: Nearly everything in JavaScript is an object, an object merely needs to be checked if it exists, so truthy check is fine.

ts_src/psbt.ts
junderw: That's what typeforce does... but the problem is that errors like that are extremely unhelpful.

If you look at the original issue, it amounts to "what does this error mean?" and it was basically the same, "Expected string, got Buffer" etc...

Saying a specifically tailored "hey, this method requires this type for this argument, and you didn't provide that." for each and every method is hard.

src/psbt.js
d-yokoi: That makes sense.

d-yokoi: Maybe it's good to consider migrating the way of checking undefined sometime.

Issue #1505, opened 2019-11-19
Can't get value of output


longhoangwkm:
```
"bitcoinjs-lib": "^5.1.6"
```

```js
// Testnet
import * as BitcoinJS from 'bitcoinjs-lib';
const txHex = '0100000003003de762bd2fe1f88c9c139c9f6dd41167adb69bc8f4a115d4d26952a689ff9a010000006a47304402205616e7d48ac0020c1306674288419f4ce0badade72c33e5c32bca6184a3d7f0d022060878f8d5deae80aeb9a80b0ba9f5a906230bd75b2ab83d8163a97c942045a67012103d67b1c931f6a7e6a9a821d23cf656b5e9baac8c1c24ab8428f9703a94807164affffffffc51d7b0bf751456978687cdb740c84da205eaa54dc06f7bd7fa974ee46bbc199000000006a47304402205dfc7592a51e603eab656e0b3efa7e1563923793af27062689270961a730e028022027d4633afdac16c5fa9f0a72f980ee6cfb5f6ecf85d9a5a9775680cee90e70db0121029cdcbcd4abb2b234c9b3c4060f88571a6992fda29099651fb12bb892dbced5e5ffffffff93d3d84e94b471e33a98a65b79f71f195e5ba3af08df7d45ffebbffdb51ac402010000006a47304402201058a8644f5260c9185f8314d0c549f27f87a4c7ad7306acddf879b4d248d512022069d1f3e4f67ef2febbf0b99f5622909acc66dc49b03b5cfa34f3985a897b5a4e012102f52c08169709f31cf0bd30ea69887bbc92a2728440e58516159d882422ee95aaffffffff0230dcfa020000000017a9144e269b90bfdc52dbd46b13eaaefaa55dcffda157876a4603060000000017a9145186e0b0a54fbf599a956a02f02ea15aee7aedbe8700000000';
const prevTx = BitcoinJS.Transaction.fromHex(txHex);
console.log(prevTx.outs[0].value);

```
I got error 
```
error TS2339: Property 'value' does not exist on type 'OpenOutput'.
  Property 'value' does not exist on type 'BlankOutput'
```

If i change `.value` to `.valueBuffer`.
I got error:
```
error TS2339: Property 'valueBuffer' does not exist on type 'OpenOutput'.
  Property 'valueBuffer' does not exist on type 'Output'.
```

lukechilds:
Your code example is working fine for me:

[...]

PR #1506, opened 2019-11-20
Fix error TS2339: Property valueBuffer does not exist on type OpenOutput

longhoangwkm:
- Fix https://github.com/bitcoinjs/bitcoinjs-lib/issues/1505
Accessing property in union of object types fails for properties not defined on all union members.

- Solution:
  Add optional types for each union member https://github.com/microsoft/TypeScript/issues/12815


rained23:
When will this be available in stable release?

junderw:
published as 5.1.7

PR #1508, opened 2019-11-22
Make less confusing and save time when finding examples

longhoangwkm:
I took a bit time to find the right example and I think other people got same trouble. So i would like to add lines blocks to easy to find right example


longhoangwkm:
cc @junderw, how do you think

junderw:
We used to do this.

1. If we do we must not use the master link, we must link to a specific commit blob.
2. Someone needs to update the links all the time when we change tests.

2 is hard, and difficult to keep up...

Sorry, I don't think this is good from a maintainability standpoint.

longhoangwkm:
Ok, got it. Bitcoin scripts actually are sometimes make me confusing

Issue #1509, opened 2019-11-23
Is there example which get block information in Bitcoin?


hyunsooda:
In the REAME.md file, it is not described how get block information in Bitcoin.

Where can I get the code or information?

junderw:
Block information from what?

If you have a block's data as Buffer, you can just do Block.fromBuffer()

hyunsooda:
Hi @junderw :)

The block information what I want to get is block size(WU or bytes), timestamp, height or # of txs, etc..

You said, `Block.fromBuffer()` function for block data query but there is no document in README.md. where is the entire document that describes data structure, function, etc?

junderw:
The Block class is not documented outside of the source code itself.

ts_src/block.ts

It is also not well maintained, I recently added the ability to verify witness commit against the segwit transactions in the block, but I have yet to update byteLength to account for segwit yet.

Pull requests are welcome.

hyunsooda:
Thanks for friendly replying @junderw :) have a nice day.

PR #1515, opened 2019-11-29
Add weight and ability to get strippedsize

junderw:
The issue #1509 brought to light that we had no method to get the weight of a Block.

Some questions to ponder:

1. Using Transaction private method is bad.
2. Maybe we should also create a private __byteLength in Block too?
3. Should we (and if so, how?) add examples about gathering info about blocks using our Block class? (The integration test only has one example, and it's not too useful.)

### DIFF
--- a/ts_src/block.ts
+++ b/ts_src/block.ts
@@ -148,13 +148,20 @@ export class Block {
     return anyTxHasWitness(this.transactions!);
   }
 
-  byteLength(headersOnly?: boolean): number {
+  weight(): number {
+    const base = this.byteLength(false, false);
+    const total = this.byteLength(false, true);
+    return base * 3 + total;
+  }
+
+  byteLength(headersOnly?: boolean, allowWitness: boolean = true): number {
     if (headersOnly || !this.transactions) return 80;
 
     return (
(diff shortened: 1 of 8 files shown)


ts_src/block.ts
```diff
@@ -148,13 +148,20 @@ export class Block {
...
+
+  byteLength(headersOnly?: boolean, allowWitness: boolean = true): number {
     if (headersOnly || !this.transactions) return 80;
 
     return (
       80 +
       varuint.encodingLength(this.transactions.length) +
-      this.transactions.reduce((a, x) => a + x.byteLength(), 0)
+      // @ts-ignore using the __byteLength private method on Transaction
+      this.transactions.reduce((a, x) => a + x.__byteLength(allowWitness), 0)
```
d-yokoi: Why don't you make public the `byteLength` method on transaction which accepts segwit option.

junderw: I was just trying to not change existing APIs... but I guess we could make allowWitness default true.

junderw: Check the new commit I just made.

Issue #1502, opened 2019-11-03
Used same pub key twice to create in p2ms


amantoor:
I used the same pubkey twice to generate a multi signature address, now i am not able to spend, I can sign transaction fine but its give error :: 64: non-mandatory-script-verify-flag (Dummy CHECKMULTISIG argument must be zero) while broadcasting. I got funds locked, how can i spend

let pubkeys = [
    '02180d73b99486aa591398f970fa64ca56385bca28ad78b9c38daecfde454a57b1',
    '02180d73b99486aa591398f970fa64ca56385bca28ad78b9c38daecfde454a57b1'
].map(hex => Buffer.from(hex, 'hex'));

let p2ms = bitcoin.payments.p2ms({ m: 1, pubkeys });
let p2sh = bitcoin.payments.p2sh({ redeem: p2ms });

junderw:
show the code you used to spend your coins. There's likely a bug.

amantoor:
let pubkeys = [
'02180d73b99486aa591398f970fa64ca56385bca28ad78b9c38daecfde454a57b1',
'02180d73b99486aa591398f970fa64ca56385bca28ad78b9c38daecfde454a57b1'
].map(hex => Buffer.from(hex, 'hex'));

let p2ms = bitcoin.payments.p2ms({ m: 1, pubkeys });
let p2sh = bitcoin.payments.p2sh({ redeem: p2ms });    
let txb = new bitcoin.TransactionBuilder(bitcoin.networks.bitcoin);
    txb.addInput(TXID, 1, null, p2sh.pubkey);
    txb.addInput(TXID, 0, null, p2sh.pubkey);
    txb.addInput(TXID, 0, null, p2sh.pubkey);
    txb.addOutput(ADDRESS, 2382000);
    tx = txb.buildIncomplete();
    for (var i = 0; i < tx.ins.length; i++) {
        tx.ins[i].script = p2sh.redeem.output;
    }
    tx = tx.toHex();

amantoor:
the same method worked for me in past when i used 2 of 3 and 2 were same pubkey and 3rd and different, but now only 2 same keys its not working

amantoor:
It is indeed a bug the script is signing the transaction twice where it should check and ignore second time. I managed to spend the funds using python bitcoinutils.

junderw:
you aren't signing anywhere.

This code by itself can never create a valid transaction.

junderw:
[...]

PR #1519, opened 2019-12-02
PSBT Bugfix for multiple of same pubkey in p2ms

junderw:
Issue #1502 made me realize that the logic for checking if p2ms if finalizable was not accounting for multiple instances of the same pubkey within the p2ms script.

This fixed that.

### DIFF
--- a/ts_src/psbt.ts
+++ b/ts_src/psbt.ts
@@ -747,16 +747,32 @@ function canFinalize(
       return hasSigs(1, input.partialSig);
     case 'multisig':
       const p2ms = payments.p2ms({ output: script });
-      return hasSigs(p2ms.m!, input.partialSig);
+      return hasSigs(p2ms.m!, input.partialSig, p2ms.pubkeys);
     default:
       return false;
   }
 }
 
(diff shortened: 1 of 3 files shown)


ts_src/psbt.ts
```diff
@@ -747,16 +747,32 @@ function canFinalize(
...
   if (!partialSig) return false;
-  if (partialSig.length > neededSigs) throw new Error('Too many signatures');
-  return partialSig.length === neededSigs;
+  let sigs: any;
+  if (pubkeys) {
+    sigs = pubkeys
+      .map(pkey => {
+        const pubkey = ecPairFromPublicKey(pkey, { compressed: true })
+          .publicKey;
+        return partialSig.filter(pSig => pSig.pubkey.equals(pubkey))[0];
```
d-yokoi: I think `find` is a bit more appropriate in this use case.
```suggestion
        return partialSig.find(pSig => pSig.pubkey.equals(pubkey));
```

junderw:
committed the suggested change

junderw:
Really glad for the gitdiff script on Travis... I would have committed it without the JS mofication... *phew*

PR #1533, opened 2020-01-14
Add BufferWriter and BufferReader class

OttoAllmendinger:
Move various write methods to a class `BufferWriter`.

This allows increased code reuse for libraries that want to implement
different serialization formats.

Also de-duplicates some code paths and increases test coverage.

Original concept by idea by https://github.com/argjv:
https://github.com/BitGo/bitgo-utxo-lib/blob/master/src/bufferWriter.js


src/buffer_writer.js
```diff
@@ -0,0 +1,44 @@
...
+  writeUInt64(i) {
+    this.offset = bufferutils.writeUInt64LE(this.buffer, i, this.offset);
+  }
+  writeVarInt(i) {
+    varuint.encode(i, this.buffer, this.offset);
+    this.offset += varuint.encode.bytes;
+  }
+  writeSlice(slice) {
+    this.offset += slice.copy(this.buffer, this.offset);
+  }
```
OttoAllmendinger: possible improvement: unlike the preceding methods, this does not throw an error if we try to write out of bounds

test/buffer_writer.spec.ts
```diff
@@ -0,0 +1,231 @@
+import * as assert from 'assert';
+import { describe, it } from 'mocha';
+import { BufferWriter } from '../src/buffer_writer';
+
+const varuint = require('varuint-bitcoin');
+
+describe('BufferWriter', () => {
+  function concatToBuffer(values: number[][]): Buffer {
+    return Buffer.concat(values.map(data => Buffer.from(data)));
+  }
```
OttoAllmendinger: this is basically a workaround to `prettier` removing my formatting of `expectedBuffer` as groups

junderw:
A few comments:

1. BufferReader would also be useful. (we use it in Block and Transaction fromBuffer)
2. Perhaps these should be included in bufferutils.ts itself.
3. This can also be used in the Block class.

I will work on it a bit and submit a commit with some changes and let me know what you think.

junderw:
Ok, tests are passing. (I added new tests for BufferReader as well)

@OttoAllmendinger If you could double check my work, I think we can merge this.

src/buffer_writer.js
junderw: Yeah maybe checking would help. I'll add it in.

OttoAllmendinger:
looks solid!

junderw:
Ok, hopefully this makes it easier for you to maintain your fork.

OttoAllmendinger:
it certainly will, although there is much more to do :)

PR #1573, opened 2020-05-21
Throw errors when p2wsh or p2wpkh contain uncompressed pubkeys.

### LINKED ISSUE #1572: P2WPKH and P2WSH Payments should check for compressed pubkeys
Segwit forbids uncompressed pubkeys, but currently we can use p2wpkh and p2wsh Payments to create addresses for scripts with uncompressed pubkeys.

This should be checked in the `validate: true` clauses.

junderw:
This will enforce BIP143 compressed pubkey rules on an address generation level.

Fixes #1572

### DIFF
--- a/src/payments/p2wsh.js
+++ b/src/payments/p2wsh.js
@@ -6,6 +6,7 @@ const bscript = require('../script');
 const lazy = require('./lazy');
 const typef = require('typeforce');
 const OPS = bscript.OPS;
+const ecc = require('tiny-secp256k1');
 const bech32 = require('bech32');
 const EMPTY_BUFFER = Buffer.alloc(0);
 function stacksEqual(a, b) {
@@ -13,6 +14,14 @@ function stacksEqual(a, b) {
   return a.every((x, i) => {
     return x.equals(b[i]);
   });
+}
+function chunkHasUncompressedPubkey(chunk) {
+  if (Buffer.isBuffer(chunk) && chunk.length === 65) {
+    if (ecc.isPoint(chunk)) return true;
+    else return false;
+  } else {
+    return false;
+  }
 }
 // input: <>
 // witness: [redeemScriptSig ...] {redeemScript}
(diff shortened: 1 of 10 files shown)


ts_src/payments/p2wsh.ts
```diff
@@ -59,6 +69,9 @@ export function p2wsh(a: Payment, opts?: PaymentOpts): Payment {
   const _rchunks = lazy.value(() => {
     return bscript.decompile(a.redeem!.input!);
   }) as StackFunction;
+  const _rochunks = lazy.value(() => {
```
lukechilds, line 72: What does this `_rochunks` stand for? Read output chunks?

Maybe we could use a more descriptive name here?

lukechilds: Also why does it need to be lazily evaluated?

src/payments/p2wsh.js
```diff
@@ -14,6 +15,14 @@ function stacksEqual(a, b) {
     return x.equals(b[i]);
   });
 }
+function chunkHasUncompressedPubkey(chunk) {
+  if (Buffer.isBuffer(chunk) && chunk.length === 65) {
+    if (ecc.isPoint(chunk)) return true;
+    else return false;
+  } else {
+    return false;
+  }
```
d-yokoi, line 24: I think it could be even simpler.
```suggestion
  if (Buffer.isBuffer(chunk) && chunk.length === 65 && ecc.isPoint(chunk)) {
    return true;
  }
  return false;
```

ts_src/payments/p2wsh.ts
```diff
@@ -18,6 +19,15 @@ function stacksEqual(a: Buffer[], b: Buffer[]): boolean {
   });
 }
 
+function chunkHasUncompressedPubkey(chunk: StackElement): boolean {
+  if (Buffer.isBuffer(chunk) && chunk.length === 65) {
+    if (ecc.isPoint(chunk)) return true;
+    else return false;
+  } else {
+    return false;
+  }
```
d-yokoi, line 28: I think it could be even simpler.
```suggestion
  if (Buffer.isBuffer(chunk) && chunk.length === 65 && ecc.isPoint(chunk)) {
    return true;
  }
  return false;
```

junderw:
Resolved all comments.

Bigpappa221 (commented):
Good setup

PR #1579, opened 2020-05-30
Psbt: fix an issue in signInputAsync

pluswave:
psbt: fix an issue for error process in signInputAsync.
    
    If keyPair.sign returns a promise and async rejects,
    (i.e, user press reject on hardware wallet)
    catch and reject the wrapper promise now.

PR #1582, opened 2020-05-31
Fix signInputAsync when SignerAsync rejects

junderw:
Replaces #1579

### DIFF
--- a/ts_src/psbt.ts
+++ b/ts_src/psbt.ts
@@ -580,31 +580,28 @@ export class Psbt {
     keyPair: Signer | SignerAsync,
     sighashTypes: number[] = [Transaction.SIGHASH_ALL],
   ): Promise<void> {
-    return new Promise(
-      (resolve, reject): void => {
-        if (!keyPair || !keyPair.publicKey)
-          return reject(new Error('Need Signer to sign input'));
-        const { hash, sighashType } = getHashAndSighashType(
-          this.data.inputs,
-          inputIndex,
-          keyPair.publicKey,
-          this.__CACHE,
-          sighashTypes,
-        );
+    return Promise.resolve().then(() => {
+      if (!keyPair || !keyPair.publicKey)
+        throw new Error('Need Signer to sign input');
+      const { hash, sighashType } = getHashAndSighashType(
+        this.data.inputs,
+        inputIndex,
+        keyPair.publicKey,
+        this.__CACHE,
+        sighashTypes,
+      );
 
-        Promise.resolve(keyPair.sign(hash)).then(signature => {
-          const partialSig = [
-            {
-              pubkey: keyPair.publicKey,
-              signature: bscript.signature.encode(signature, sighashType),
-            },
-          ];
+      return Promise.resolve(keyPair.sign(hash)).then(signature => {
+        const partialSig = [
+          {
+            pubkey: keyPair.publicKey,
+            signature: bscript.signature.encode(signature, sighashType),
+          },
+        ];
 
-          this.data.updateInput(inputIndex, { partialSig });
-          resolve();
-        });
-      },
-    );
+        this.data.updateInput(inputIndex, { partialSig });
+      });
+    });
   }
 
   toBuffer(): Buffer {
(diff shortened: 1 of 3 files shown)


ts_src/psbt.ts
```diff
@@ -584,25 +584,45 @@ export class Psbt {
...
-              pubkey: keyPair.publicKey,
-              signature: bscript.signature.encode(signature, sighashType),
-            },
-          ];
+          result = keyPair.sign(hashAndSighash.hash);
 
-          this.data.updateInput(inputIndex, { partialSig });
-          resolve();
-        });
+          if (result && typeof (result as any).then === 'function') {
```
junderw, line 601: Note: Apparently the ECMAScript definition of a Promise is "check for then function in the prototype, if so, treat as Promise"... so this seems like the best way to check if it is a Promise.

junderw: `result instanceof Promise` also seems fine.

lukechilds:
A common pattern I use when returning a Promise and would recommend here is to wrap the function in:

```js
Promise.resolve().then(() => {
});
```

This ensures it's impossible to throw an error. Any errors thrown inside this block will just cause the parent Promise to reject.

Looks like your latest commit correctly handles all errors too but IMO this method is much more readable and less error prone.

lukechilds:
So this:

```js
signInputAsync(
   inputIndex,
   keyPair,
   sighashTypes = [transaction_1.Transaction.SIGHASH_ALL],
 ) {
   return new Promise((resolve, reject) => {
     if (!keyPair || !keyPair.publicKey)
       return reject(new Error('Need Signer to sign input'));
     let result;
     let hashAndSighash;
     try {
       hashAndSighash = getHashAndSighashType(
         this.data.inputs,
         inputIndex,
         keyPair.publicKey,
         this.__CACHE,
         sighashTypes,
       );
       result = keyPair.sign(hashAndSighash.hash);
     } catch (err) {
       return reject(err);
     }
     Promise.resolve(result)
       .then(signature => {
         const partialSig = [
           {
             pubkey: keyPair.publicKey,
             signature: bscript.signature.encode(
               signature,
               hashAndSighash.sighashType,
             ),
           },
         ];
         this.data.updateInput(inputIndex, { partialSig });
       })
       .then(resolve)
       .catch(reject);
   });
 }
```

Becomes:

```js
 signInputAsync(
    inputIndex,
    keyPair,
    sighashTypes = [transaction_1.Transaction.SIGHASH_ALL],
  ) {
    return Promise.resolve.then(() => {
      if (!keyPair || !keyPair.publicKey)
        throw new Error('Need Signer to sign input');
      const hashAndSighash = getHashAndSighashType(
        this.data.inputs,
        inputIndex,
        keyPair.publicKey,
        this.__CACHE,
        sighashTypes,
      );
      const signature = keyPair.sign(hashAndSighash.hash);
      const partialSig = [
        {
          pubkey: keyPair.publicKey,
          signature: bscript.signature.encode(
            signature,
            hashAndSighash.sighashType,
          ),
        },
      ];
      this.data.updateInput(inputIndex, { partialSig });
    });
  }
```

junderw:
@lukechilds Got it. Force pushed just now while you were commenting.

ts_src/psbt.ts
```diff
@@ -580,31 +580,28 @@ export class Psbt {
...
+      );
 
-        Promise.resolve(keyPair.sign(hash)).then(signature => {
-          const partialSig = [
-            {
-              pubkey: keyPair.publicKey,
-              signature: bscript.signature.encode(signature, sighashType),
-            },
-          ];
+      return Promise.resolve(keyPair.sign(hash)).then(signature => {
```
lukechilds, line 594: Isn't `keyPair.sign(hash)` sync?

So you can just

```js
const signature = keyPair.sign(hash);
```

Instead of wrapping with another Promise.
```

junderw:
@lukechilds Good idea. This makes the diff very clean and the primary changes will be just adding the tests thanks to @pluswave

ts_src/psbt.ts
junderw: It's either sync OR async

junderw: You can use a Signer with this if you want to free up the event loop.

PR #1618, opened 2020-09-10
P2PSH - output script name/type missing

motorina0:
### Curent State

- the unit tests fixures (p2sh.json) expect a name (for the payment) that indicates the type of the wrapped script:
```
      "expected": {
        "name": "p2sh-p2pk",
```

- `payment.utils.ts `currently does **not** check the name in the `equate()` method
    - `  if ('name' in b) t.strictEqual(a.name, b.name, 'Inequal *.name');` <-- **not present**

- **p2sh.ts** & **p2wsh** does not try to match the redeem scripts with an existing template (p2pk, p2pkh, etc)

### Expected State
 - **p2sh** & **p2wsh** should return the name of the wrapped script (when it is the case). Eg: `p2sh-p2pkh`

### Other Considerations
  - the `classify` functions can perform some heavy operations like `isCanonicalScriptSignature`, `isCanonicalPubKey` etc
      - this might be a problem, however the value for the `name` field is lazy computed
  - the names returned by the `clasify` functions differ from the ones used in the unit test fixures (eg: `pubkeyhash` vs`p2pkh`)
     - the unit test fixures have been changed

### DIFF
--- a/src/payments/p2sh.js
+++ b/src/payments/p2sh.js
@@ -1,5 +1,6 @@
 'use strict';
 Object.defineProperty(exports, '__esModule', { value: true });
+const classify = require('../classify');
 const bcrypto = require('../crypto');
 const networks_1 = require('../networks');
 const bscript = require('../script');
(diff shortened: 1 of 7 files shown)


junderw:
classify is an internal-only module and is going to be deprecated in v6 along with TransactionBuilder (since that is the only public facing module that depends on it).

Please don't use the naming conventions of classify.

Thanks.

motorina0:
OK, thank you!
Closing this PR.

Suggestion (I can make new PR):
   -  since it is never checked, remove the `name` from the `expected` values in `p2sh.json`
```
      "expected": {
        "name": "p2sh-p2pk",
```

junderw:
yes. or change them all to "p2sh" since that will help people reading the tests understand how the library works. Then add the name check.

The reason it doesn't get the `-xxx` added after is because we are creating objects from a JSON rather than using the payments API.

Perhaps I could also add a simple check using payments to find which script it is... we use something similar in psbt.ts... we could move it to payments.utils.ts and use it in a similar way you were trying to use classify.

PR #1657, opened 2020-11-27
Issue 1477

motorina0:
Remove Transaction Builder


ts_src/psbt.ts
```diff
@@ -117,8 +117,7 @@ export class Psbt {
       __NON_WITNESS_UTXO_BUF_CACHE: [],
       __TX_IN_CACHE: {},
       __TX: (this.data.globalMap.unsignedTx as PsbtTransaction).tx,
-      // Old TransactionBuilder behavior was to not confirm input values
-      // before signing. Even though we highly encourage people to get
+      // Even though we highly encourage people to get
```
junderw, line 120: I don't think these comments should be removed. Perhaps changed.

ie. TransactionBuilder -> "Psbt's previously removed predecessor" etc.

test/script.spec.ts
```diff
@@ -57,7 +56,7 @@ describe('script', () => {
   });
 
   describe('fromASM/toASM (templates)', () => {
-    fixtures2.valid.forEach(f => {
+    fixtures.valid2.forEach(f => {
```
junderw, line 59: :+1:

ts_src/psbt.ts
motorina0: - updated comments to include the reference to `TransactionBuilder`

PR #1676, opened 2021-02-15
Support BIP350 bech32m address serialization and future segwit versions

andrewtoth:
Adds support for BIP350. Relevant test vectors are added from [here](https://github.com/bitcoin/bips/blob/master/bip-0350.mediawiki#test-vectors-for-v0-v16-native-segregated-witness-addresses). Note that this includes implicit support for sending to future segwit versions beyond v1 BIP341, as instructed [here](https://github.com/bitcoin/bips/blob/master/bip-0350.mediawiki#test-vectors).

PR #1680, opened 2021-02-19
Backport #1676 to v4

andrewtoth:
Unfortunately, the merge commit of #1676 couldn't be applied cleanly. I've separated the same 3 commits out again for easier review.


andrewtoth:
Travis is failing because v4 doesn't build with node v14.

junderw:
I don't want to keep maintaining <=v4.

We're about to move to v6.

bech32m will be supported in v6.

forwards compatibility of witness versions, while specified in the BIP is not how people have implemented or understand bech32, so loosening the requirements could break assumptions by app devs.

andrewtoth:
@junderw I suppose that's fair. I am a bit disappointed since I maintain a project still on v4 and it would be nice to be able to send to taproot without having to upgrade. Is there any way to convince you to change your mind here? How about if it is released as a minor version instead of patch?

Also, I couldn't find any documentation regarding versioning and EOL timelines for different major versions. Is this documented anywhere?

Finally, your last comment left me wondering - will there at least be a v5 version with the capability to send to taproot and future segwit versions?

junderw:
> I maintain a project still on v4 and it would be nice to be able to send to taproot without having to upgrade.

v4 won't be changing unless there's a big vulnerability that needs fixing, so maintaining this patch won't be hard. Just point your dependency to your github branch. Maintaining this patch on your own won't be too hard since nothing will be changing constantly under your patch.

> Is there any way to convince you to change your mind here? How about if it is released as a minor version instead of patch?

This is not a minor change, sorry.

> Is this documented anywhere? (EOL)

Unfortunately, no. While some FOSS projects are nice enough to provide promises of support, most of those projects are funded by large companies. Currently, BitcoinJS is not officially funded by anyone and relies on volunteers including myself that may or may not be willing / able to merge in needed changes sometime down the road. So an EOL schedule (which is a promise to support a given version until a given date) is not possible since we don't even give any promise to support the latest version.

> will there at least be a v5 version with the capability to send to taproot and future segwit versions?

This is not a minor change, so no, unfortunately this will not be in v5.

andrewtoth:
Thanks for the response. When can we expect a V6 release then? Is there anything holding it back?

andrewtoth:
I think it's important to make the ability to send to taproot widely available as quickly and easily as is safe to do so. Many projects won't even consider using taproot unless there's already a good percentage of others that will be able to send to it. As we all know many projects move very slowly, that's why I think it's important to release the ability to send as soon as is safe.

PR #1716, opened 2021-08-17
fix(transaction): use writeInt32 to write version

OttoAllmendinger:
We are reading `version` as `int32` so we should write it as that as
well.

### DIFF
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -316,7 +316,7 @@ class Transaction {
     tbuffer = Buffer.allocUnsafe(156 + varSliceSize(prevOutScript));
     bufferWriter = new bufferutils_1.BufferWriter(tbuffer, 0);
     const input = this.ins[inIndex];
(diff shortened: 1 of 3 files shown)


OttoAllmendinger:
Interestingly `readInt32/writeInt32` is only used to write `version` and `hashType`. Arguably `uint32` makes more sense for both of them.

There is however tests that expect `version` to be a signed integer. This is probably just to be compatible with the C++ sematics, however I don't think even the core source code makes use of the negative `version` range.

> https://github.com/bitcoinjs/bitcoinjs-lib/blob/48bf08c0d3649fa36278637d4e16b8cf6231b767/test/block.spec.ts#L15

> https://github.com/bitcoinjs/bitcoinjs-lib/blob/9810049f4cec0421dbe66a8d1c761547ac4e1fba/test/transaction.spec.ts#L84

PR #1717, opened 2021-08-17
fix(transaction): use writeInt32 to write version in hashForWitnessV0

OttoAllmendinger:
We are reading `version` as `int32` so we should write it as that as
well.

### DIFF
--- a/test/bitcoin.core.spec.ts
+++ b/test/bitcoin.core.spec.ts
@@ -214,6 +214,16 @@ describe('Bitcoin-core', () => {
             (hash.reverse() as Buffer).toString('hex'),
             expectedHash,
           );
+
+          assert.doesNotThrow(() =>
+            transaction.hashForWitnessV0(
+              inIndex,
+              script,
+              0,
+              // convert to UInt32
+              hashType < 0 ? 0x100000000 + hashType : hashType,
+            ),
+          );
         },
       );
     });
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -316,7 +316,7 @@ class Transaction {
     tbuffer = Buffer.allocUnsafe(156 + varSliceSize(prevOutScript));
     bufferWriter = new bufferutils_1.BufferWriter(tbuffer, 0);
     const input = this.ins[inIndex];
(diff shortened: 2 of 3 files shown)


OttoAllmendinger:
Interestingly `readInt32/writeInt32` is only used to write `version` and `hashType`. Arguably `uint32` makes more sense for both of them.

There is however tests that expect `version` to be a signed integer. This is probably just to be compatible with the C++ sematics, however I don't think even the core source code makes use of the negative `version` range.

> https://github.com/bitcoinjs/bitcoinjs-lib/blob/48bf08c0d3649fa36278637d4e16b8cf6231b767/test/block.spec.ts#L15

> https://github.com/bitcoinjs/bitcoinjs-lib/blob/9810049f4cec0421dbe66a8d1c761547ac4e1fba/test/transaction.spec.ts#L84

test/bitcoin.core.spec.ts
```diff
@@ -214,6 +214,16 @@ describe('Bitcoin-core', () => {
...
             expectedHash,
           );
+
+          assert.doesNotThrow(() =>
+            transaction.hashForWitnessV0(
+              inIndex,
+              script,
+              0,
+              // convert to UInt32
+              hashType < 0 ? 0xffffffff + hashType : hashType,
```
junderw, line 224: `0x100000000 + -1 === 0xffffffff` since -1 should be all f in hex.
`0x100000000 + -2147483648 === 0x80000000` since -2147483648 should be only sign bit.

```suggestion
              hashType < 0 ? 0x100000000 + hashType : hashType,
```

OttoAllmendinger: ah yes of course

OttoAllmendinger: I would have preferred `>>> 0` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Unsigned_right_shift) but it doesn't play nice with `tslint` apparently

junderw (approved):
This is not urgent, since Bitcoin doesn't use a Transaction version > 31 bits anyways.

Issue #1522, opened 2019-12-09
TODO: Add Tapscript / Taproot feature


junderw:
This will be a place to brainstorm ideas...

Needs:

1. Tagged Hashes (maybe make a smaller separate package)
2. Taproot builder (can we wedge this into the existing Payments API??? it seems to act similar to p2sh and p2wsh, and the lazy-loading should help greatly with performance due to the crypto operations to calculate the tweaked pubkey)
3. Segwit v1 Address generation via the Payments(Taproot) API
4. Schnorr signing module. (Modify tiny-secp256k1... not sure what to do with native JS though.)

Questions:

1. When should this be implemented? I'm guessing we should wait for at least a testnet release of Taproot before spending time on it... But at the same time, it would help the development of taproot if we can look at taproot with fresh eyes / a new perspective and make suggestions based on problems while implementing.
2. This is going to be a pretty big change... so maybe it should be a major version bump and that way we can reconsider existing API structure when adding Taproot.

Thoughts and comments welcome.

junderw:
Working on it now.

Starting with tiny-secp256k1.

I am thinking of implementing TaggedHash and schnorr signing in tiny-secp256k1 itself since it is very bitcoin-specific. There is a PR currently up against secp256k1lib and I'll use that for the C++ addon for now.

WASM was considered to begin with for browsers but we'll put that off for now.

andrewtoth:
Is there any update on this? I would like to help out if possible.

junderw:
Hi @andrewtoth, thanks a lot.

There are quite a few large things I need to work on in order to get this out.

You helped a ton with # 3, thanks.

In order of difficulty, the next step would be to create a TaggedHash module. You can take care of that if you'd like. If you want you can just write a single js file and post a gist or create your own repo and I can take it into the BitcoinJS org. If you don't want to maintain it, it should be simple enough so we won't mind maintaining it.

After that... 2 is code-wise next difficult... but concept-wise it's one of the most difficult. I have been debating whether to just scrap the Payment API for a similar API that is more aligned with Output Descriptors. I envision an API where the methods would be able to calculate and output ranges of addresses, and there would be a `toOutputDescriptor()` method that would output the checksummed output descriptor.

As far as tiny-secp256k1 is concerned... I just need to sit down and do the work... I am thinking of borrowing code from node-secp256k1 so we can update to using N-API for native bindings.

[...]

PR #1726, opened 2021-09-22
feat: add initial p2tr output script support

sangaman:
This introduces support for creating p2tr output scripts and addresses from one or more pubkeys and/or one or more scripts according to BIPs 340 & 341.

Multiple pubkeys are aggregated according to the procedure defined by MuSig whereby all keys are concatenated and hashed, with the resulting hash multiplied against each pubkey to tweak it. These tweaked pub keys are added together to produce the internal pubkey. If only one pubkey is provided, then the internal pubkey is simply set to that. And if no pubkeys are provided, a key with no known secret key is used instead.

Any scripts provided are arranged into a Huffman tree according to their weighted probability and then hashed to form a merkle tree. The root hash of this tree is then added to the internal pubkey to tweak it. This tweaked key is finally used for creating the output script which is then encoded using Bech32m.

Test cases in this commit include examples from the Bitcoin Optech taproot workshop that derive a Bech32m address from scripts and an internal key.

Follow-ups to this commit should include expanding the coverage and number of test cases as well as possibly restructuring existing code or moving logic to other packages/modules. We should also consider support for taproot output descriptors, which are currently under review and developmnet in Bitcoin Core.

Ticket: BG-35571


ts_src/taproot.ts
```diff
@@ -0,0 +1,205 @@
...
+ */
+export function aggregateMuSigPubkeys(pubkeys: Buffer[]): Buffer {
+  // sort keys in ascending order
+  pubkeys.sort();
+
+  const trimmedPubkeys: Buffer[] = [];
+  pubkeys.forEach(pubkey => {
+    const trimmedPubkey = trimFirstByte(pubkey);
+    trimmedPubkeys.push(trimmedPubkey);
+  });
```
junderw, line 44: Shouldn't these be sorted after trimming?

junderw (commented):
A few comments.

ts_src/taproot.ts
junderw: Also, shouldn't there be a check (and throw an error) if two pubkeys are the same?

junderw: (for p2ms we allow multiple of the same pubkey since Bitcoin allows this... but for aggregating, I think two of the same pubkey is not mathematically possible? I could be wrong.)

ts_src/taproot.ts
```diff
@@ -0,0 +1,205 @@
...
+/**
+ * Trims the leading 02/03 byte from an ECDSA pub key to get a 32 byte schnorr
+ * pub key with x-only coordinates.
+ * @param pubkey A 33 byte pubkey representing an EC point
+ * @returns a 32 byte x-only coordinate
+ */
+export function trimFirstByte(pubkey: Buffer): Buffer {
+  assert(pubkey.length === 33);
+  return pubkey.slice(1, 33);
+}
```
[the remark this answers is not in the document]
sangaman: Yes I'm pretty sure this is the case. For taproot purposes, we assume the positive Y coordinate by default. This is covered in more detail in the "Implicit Y coordinates" section of BIP340. My thinking here is ideally we would work with 32 byte keys for the key addition/subtraction natively with the secp256k1 dependency, I think that's possible if it were updated to use the latest libsecp256k1 (would need to update `rust-secp256k1` first I believe). For now I'm just manually trimming the first byte where necessary, though.

ts_src/taproot.ts
sangaman: You're correct about sorting after trimming. I think a sanity check here to prevent duplicate pubkeys makes sense even if it's technically possible.

sangaman:
Thanks for the comments. I actually opened this PR by mistake - I was intending to open it against our BitGo fork (here https://github.com/BitGo/bitcoinjs-lib/pull/8) for internal review before trying to merge changes upstream once they're in a more complete state.

But I appreciate the input/feedback. I was meaning to bump the discussion in #1522 about the best path forward for developing and working on these changes.

I also notice now from rereading that issue that there's a tagged hash module. I could use that here. For now I'm going to tweak the approach in my branch to make it a bit cleaner/more reusable.

PR #1742, opened 2021-11-03
feat: add support for pay to taproot

motorina0:
## Summary
This is the first version for the `p2tr` payment. It is not complete, but it covers most of the functionality.
I'm opening this PR for others to test, review and provide feedback.
Main changes listed below:

### New payment type: `P2TR`
It follows the same approach as the other payments (`p2wsh`, `p2pkh`, ...)
- the `PaymentOpts` object takes an extra field: `eccLib?: TinySecp256k1Interface`
  - this field is required for `P2TR` (but not for the other payment types)
  - `eccLib` is required for the tweak operations. The `ECC` logic has been extracted out of this lib.
- examples of how to use `P2TR` can be found in the unit tests and integration tests (see tests section)

### `Payment` interface
The `Payment` interface has 3 new fields:
 - `internalPubkey` - the internal key as defined by [BIP341](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#constructing-and-spending-taproot-outputs). 
 - `scriptTree` - a Merkle tree whose leaves consist of a version number and a script (see [BIP341)](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#design)
 - `redeemVersion` - the version of the leaf being spent (if the script-path spend is used). Default value: `0xc0`

Some `Payment` fields have `taproot` specific meaning:
- `pubkey` - the tweaked `internalPubkey` as defined in [BIP341](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#design)
  - > is computed as P + hash(P||m)G for a public key P, and the root m of the Merkle tree
- `hash` - the hash of the root of the Merkle `scriptTree`. Empty if not `scriptTree` is used.
- `signature` - a Schnorr signature,  if key-path spent is used. Empty otherwise.
- `redeem.output` - the leaf script, for `script-path` spends

### Taproot Helpers
- taproot related logic has been extracted to `ts_src/payment/taprootutils.ts`
- in the future it might be moved to a `bip341` module. See https://github.com/bitcoinjs/bitcoinjs-lib/issues/1780

### Address
- add stricter rules for Taproot (data size 32 and public key valid x-only coordinate)
- `from|toOutputScript`  takes in an optional `eccLib`. If missing, then it skips the `p2tr` matching falls down to the final Error

### PSBT
- it can now add Taproot inputs and outputs
- for the `key-path` spend the signer has to be tweaked (it does **not** perform the tweak internally)
- for the `script-path` spend a custom finalizer (`FinalScriptsFunc`) is **required** when a Taproot input is finalized. See `test/psbt.utils.ts` for such an example
- `PsbtOptsOptional` has a new field `eccLib?: TinySecp256k1Interface;`. `eccLib` is required if the PSBT instance uses taproot ins/outs
- the `Signer` interface has a new method: `signSchnorr?(hash: Buffer): Buffer;`. This method is required if a Taproot UTXO is being spent.

### Tests
 - unit test data
   - `p2tr.json`: data generated using the bitcoin test framework [test/functional/wallet_taproot.py](https://github.com/bitcoin/bitcoin/blob/master/test/functional/wallet_taproot.py)
   - `BIP341 test cases`: [Test vectors for scriptPubKey construction](https://github.com/sipa/bips/blob/202110_bip341_vectors/bip-0341/wallet-test-vectors.mediawiki) recommended [here](https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2021-November/019587.html) by Pieter Wuille
 - integration tests (`test/integration/taproot.spec.ts`)
 - **script-path** spend on `testnet` ([code sample and transaction links](https://gist.github.com/motorina0/e6bb1b32b3d5d2fc80e1bd63d5efea4f))

### Further Comments
Note for reviewers: all `.js` files are generated, only the `.ts` files require review


motorina0:
Note: `lint` expected to fail, see **Summary**

junderw:
Thanks for the PR!

Just a quick note before I review:

This effort is duplicated with BitGo's fork, which is adding native JS schnorr and using TransactionBuilder.

https://github.com/BitGo/bitcoinjs-lib/blob/master/ts_src/payments/p2tr.ts

I will most likely pick and choose parts from both efforts and add everyone as a co-author.

motorina0:
Thanks @junderw! That is fine with me.
Do you have a timeline for the official release that will include the `p2tr` functionality?

junderw:
timeline? Not really.

Right now:

1. Moving ecpair and bip32 libraries to a modular method of injecting the needed ecc interfaces. (following tiny-secp256k1, so we will ensure that tiny-secp256k1 implements the needed interface)
2. Adding segwit v1 sigHash to Transaction.
3. Adding p2tr Payment.
4. Adding support for Psbt.

4 will be more difficult since right now the extensions needed for taproot have not been decided yet, so it will need to be made as an experimental API (that isn't importable/exportable, similar to how TransactionBuilder used to hold extra data that can't be represented in the transaction serialization) added onto Psbt, which we will mark as experimental, and make it clear that the API will break on a patch/minor version change without warning.

I am thinking of just releasing the new v6 once we're done with 3, then we can just have a few examples of creating a Transaction manually using just the Transaction class and its methods...

Then once the taproot PSBT extensions are in stone, we will need to add support to bip174 for v2 PSBTs (I think taproot will require v2 support), then implement the taproot extensions, then implement into Psbt here, then maybe release v6.1.0.

I hope to get done with up to number 3 by beginning of next week ish.

junderw:
Also, I think p2tr should also take and injected interface for the tweak operations. (which tiny-secp256k1 will implement)

I want to keep ECC stuff out of bitcoinjs-lib moving forward.

Basically, people will just add tiny-secp256k1 and pass it into ecpair, bip32, and p2ts etc. and it should just work.

Then anyone who implements a native JS version of tiny-secp256k1 with schnorr moving forward (ie. BitGo) they can inject it.

junderw:
1-3 won't take too long, I just need to find time.

If you are looking for something to help on, a PR adding v2 support for bip174 would help implement 4 faster once taproot extensions are decided.

junderw:
v6 pushed.

motorina0:
rebased

junderw:
I was hoping to keep p2tr a non-breaking change.

So rather than use a Factory to get all payments, I was thinking either:

1. p2tr is a factory
2. p2tr input arguments include an optional eccLib (which is the tiny-secp256k1 minimal interface needed) which will only be validated when the inputs require it... ie. if only address is input, we can only output the contents of the address, so we don't need eccLib... but if the input has a pubkey then we need eccLib to do the tweaking, so if we have pubkey without eccLib the validation logic will throw an Error.

I think 2 is better, since it allows people who don't need to perform calcs with ecc to not need an ecc library.

motorina0:
> p2tr input arguments include an optional eccLib
 - [x] done
> which is the `tiny-secp256k1` minimal interface needed
 - [x] done. Only `isXOnlyPoint()` and `xOnlyPointAddTweak()` are exposed/required
> which will only be validated when the inputs require it...
 - [x] done. Cases when the `ecc` lib is required:
    - check if  `internalPubkey` and `pubkey` pass `isXOnlyPoint()`
    - `internalPubkey` and `tweak` are provided -> compute `pubkey` (taproot output key), using `xOnlyPointAddTweak()`
       - also validates that the tweaked `internalPubKey` matches the `pubkey` (taproot output key)

ts_src/payments/p2tr.ts
```diff
@@ -9,275 +9,285 @@ import {
...
-        if (_address().version !== TAPROOT_VERSION)
-          throw new TypeError('Invalid address version');
-        if (_address().data.length !== 32)
-          throw new TypeError('Invalid address data');
-        pubkey = _address().data;
-      }
+export function p2tr(
+  a: Payment,
+  opts?: PaymentOpts,
+  eccLib?: TinySecp256k1Interface,
```
junderw, line 24: I was thinking it should be a member of the first argument `a`

motorina0: - `a` and the output `o` are related data structures (`o` is `a` after the computations and validations have been applied)
  - if `eccLib` is not expected to be present on `o` then it shouldn't be on `a` either
- the `PaymentOpts` looks be a better place for `eccLib` (`eccLib` is actually an option for the payment)

junderw: good point. PaymentOpts sounds better

motorina0: > good point. PaymentOpts sounds better
Done!

motorina0:
I'm trying to figure out how this PR can be improved in order to get it merged sooner rather than later.
This is what I could think off so far:
1. more code comments (explain code in relation to `BIP341`) 
1. more unit tests (what scenarios?)
1. documentation on how to use `p2tr`

Any other `ideas`/`suggestions`/`requests`/`complaints` ?

ts_src/payments/index.ts
```diff
@@ -17,28 +19,36 @@ export interface Payment {
...
+  scriptsTree?: any; // todo: solve
+  scriptLeaf?: TaprootLeaf;
   witness?: Buffer[];
 }
 
-export type PaymentCreator = (a: Payment, opts?: PaymentOpts) => Payment;
+export type PaymentCreator = (
+  a: Payment,
+  opts?: PaymentOpts,
+  eccLib?: TinySecp256k1Interface,
```
junderw, line 36: not needed anymore

ts_src/payments/testecc.ts
```diff
@@ -0,0 +1,176 @@
...
+      assert(Buffer.from(r!.xOnlyPubkey).equals(h(t.result)));
+    }
+  });
+}
+
+function assert(bool: boolean): void {
+  if (!bool) throw new Error('ecc library invalid');
+}
+
+const tweakAddVectors = [
```
junderw, line 55: only include the 1st 2nd and 5th vector. Remember these will be included in bundles people make, so the bare minimum to make sure it handles an edge case, and 1-2 main cases should be enough.

junderw:
I wonder if we should offer a tool for sorting leafs in order of weight (likelihood of use)
It should not be required to sort in any sort of order, but we should maybe offer some sort of helper function to help out.

I don't think it should be a show stopper, but it would be nice to have.

motorina0:
> ... but we should maybe offer some sort of helper function to help out.

It would definitely be nice to have this. I have added this task for it: https://github.com/bitcoinjs/bitcoinjs-lib/issues/1766
It might start some conversations about how the Inputs/Outputs should look like, so it is better to keep it separate from this PR.

ts_src/payments/testecc.ts
motorina0: Done!

ts_src/payments/index.ts
motorina0: Done!

motorina0:
Me again...

I have made the following changes to the `Psbt` class. It allows to spend from taproot UTXOs and to send to taproot addresses.
A `testnet` transaction created with this library:
https://mempool.space/testnet/tx/735321667cf43ded23dc935f355c4b1b1e77d9a47d11c77d384af5094e1ad171


The main changes are listed below. They can be rolled back and included in a separate PR if needed. They are useful for testing the `p2tr` functionality:
 - `signInput()` creates and serializes the Schnorr signature for the taproot inputs (key spend)
 - `validateSignaturesOfInput()` validates taproot inputs for `Key Spend signatures`
 - only `SIGHASH_DEFAULT` supported at the moment
 - add `tweakSigner()` as static method
 - add unit tests
 - the `Signer` interface has an optional `privateKey` field (used for tweaking)
 - the `signInputHD()` and `signInputHDAsync()` logic not implemented yet (waiting for feedback)
 - direct dependency to `tiny-secp256k1` introduced :(
    - it is awkward to pass an `ecc` lib from outside _<- must be revisited_
    - added  'tiny-secp256k1' to `package.json` deps _<- must be revisited_

I know that `"4. Adding support for Psbt."` requirements are more broad, this is just a start...

ts_src/psbt.ts
```diff
@@ -810,6 +782,40 @@ export class Psbt {
...
+ * Helper method for converting a normal Signer into a Taproot Signer.
+ * Note that this helper method requires the Private Key of the Signer to be present.
+ * Steps:
+ *  - if the Y coordinate of the Signer Public Key is odd then negate the Private Key
+ *  - tweak the private key with the provided hash (should be empty for key-path spending)
+ * @param signer - a taproot signer object, the Private Key must be present
+ * @param opts - tweak options
+ * @returns a Signer having the Private and Public keys tweaked
+ */
+export function tweakSigner(signer: Signer, opts: TaprootSignerOpts): Signer {
```
junderw, line 795: This helper function is out of place. It also is returning ECPair which we removed from the library in v6.

It would be much easier to create a new interface called TweakedSigner which has a method called `signTweaked(hash: Buffer, tweak: Buffer, extraEntropy?: Buffer): Buffer` etc.

Then we can implement that for ECPair and bip32 separately.

junderw: Although, another way to deal with it is to just tell the Signer it is to sign itself tweaked... and just automatically do that with signSchnorr when the internal flag is set... (it knows the tweak since the "recommended" tweak is of its own xonlypubkey)

I don't know which way is more useable... but this helper function is not the right path imo.

motorina0:
>  ... but this helper function is not the right path imo
 - I agree... 

> it knows the tweak since the "recommended" tweak is of its own xonlypubkey
 - this would mean adding a dependency in `ecpair` to `'create-hash'` (in order to compute the tweak)
 - it would also not work when the tweak of the script tree is used

One option is to change the `signSchnorr()` of `ecpair`:
**From**:
 - `signSchnorr(hash: Buffer): Buffer`
 
**To**:
 - `signSchnorr(hash: Buffer,  tweakHash?: Buffer): Buffer`

This would mean:
- the `tweakHash` is computed outside the Signer, no tweaking if falsy
- only `signSchnorr()` can sign tweaked (so far)
- `signSchnorr()` should also take `extraEntropy` optional param (missing now in `ecpair`)

Implementation quick view:
```
    signSchnorr(hash: Buffer, tweakHash?: Buffer, e?: Buffer): Buffer {
      if (!this.privateKey) throw new Error('Missing private key');
      if (!ecc.signSchnorr)
        throw new Error('signSchnorr not supported by ecc library');
      if (!tweakHash)
        return Buffer.from(ecc.signSchnorr(hash, this.privateKey, e));

      const privateKey =
        this.publicKey[0] === 2
          ? this.privateKey
          : ecc.privateNegate(this.privateKey!);

      const tweakedPrivateKey = ecc.privateAdd(privateKey, tweakHash);
      if (!tweakedPrivateKey) {
        throw new Error('Invalid tweaked private key!');
      }
      return Buffer.from(ecc.signSchnorr(hash, tweakedPrivateKey, e));
    }
```

motorina0:
> One option is to change the signSchnorr() of ecpair...

I've open a quick PR with these change https://github.com/bitcoinjs/ecpair/pull/6

motorina0:
> This helper function is out of place. It also is returning ECPair which we removed from the library in v6.
What was I thinking? I have removed these changes.

motorina0:
@junderw what can I do to get this PR merged?
Can we create a list of tasks to implement/test/document?
I'm willing to put in the effort, but I feel a bit lost...

junderw:
- Needs integration tests (so we can verify that we can actually make PSBTs and sign them and they are valid to Bitcoin Core)
- ECPair decisions might still affect the Signer interface changes...
- bip174 library should add support for the input keys that store taproot info. (I don't think we need to support PSBTv2 in order to support those new keys.)
- Psbt should also support taproot info. At least holding it. And the finalizing function should be able to access that data (so people could write their own custom finalizer for their script path spends.

Is about what I'm thinking right now.

Some of this could be split off into separate PRs tho.

1. Def want integration tests before merge
2. I would like to nail down the Signer changes before any published release. We could merge to master tho.

junderw:
I think writing integration tests using Psbt will help uncover any shortcomings.

tbh, I think when you first write an integration, just do it normally like you would with any app you made. (Don't try to over optimize it with the createPayment abstraction I have there currently etc.)

That way you get some experience actually using the p2tr payment and Psbt with taproot, and maybe you'll find some things about the API that are unintuitive etc.

junderw:
from|toOutputScript should take in an optional eccLib, and if there is no eccLib, skip the p2tr matching and let the logic fall down to the final Error (each function ends with a throw Error) or maybe figure out a good way to throw an error when we only know that they could be trying to en|decode p2tr...

motorina0:
`from|toOutputScript`:
 - changes here (if OK, will merge into this branch): 
     - https://github.com/bitcoincoretech/bitcoinjs-lib/pull/2
 - original issue here: 
     - https://github.com/bitcoinjs/bitcoinjs-lib/issues/1750

junderw:
Oh yeah that's good. It's probably better to merge into this PR.

brandonblack:
With the goal of eventually converging to a single implementation, I have updated the BitGo version of Taproot support to work with v6 here: https://github.com/BitGo/bitcoinjs-lib/tree/6_sync . Not sure if anything we have there is currently of use, but wanted to mention it.

Main differences:
* This PR includes PSBT updates
* BitGo code includes MuSig2 aggregate key creation (but not signing)

tiero:
> * taproot related logic has been extracted to `ts_src/payment/taprootutils.ts`. There might be a better place for it.

What about introducing a `bip341.ts` like has been done for `bip66`?

(we are trying to merge v6 into LiquidJS, a bitcoinJS fork for Liquid Network, and we found the need to have some utils as well, you can look our work over [here](https://github.com/vulpemventures/liquidjs-lib/pull/39/commits/1dee4f315cc2aa9b89e317a5045426ab9362b3b7?file-filters%5B%5D=.ts&show-viewed-files=true#diff-43b5140098b71f4ba8d237d8f366a111598c25a291ffdab4d4408721f46b8350), would love to be compatible)

motorina0:
@brandonblack 
 - thank you! There is certainly some useful functionality on that branch.
 - I will try to see if there are some low hanging fruits.
 - please not the this branch is still WIP, changes to the public APIs are still happening (mainly `p2tr`)

motorina0:
@tiero that is a great idea. I have created this issue in order to keep the discussion separate (this tread is quite large already)
https://github.com/bitcoinjs/bitcoinjs-lib/issues/1780

motorina0:
@junderw 
- integration tests have been added (in `taproot.spec.ts`) for:
   - [x] 2 x taproot key-path spend
     - p2tr with just and internal pub key
     - p2tr with a scriptTree (unused)
   - [x] 3 x taproot script-path spend
     - OP_CHECKSIG
     - OP_CHECKSEQUENCEVERIFY
     - OP_CHECKSIGADD (3-of-3)
       -  2-of-3 should be possible with some changes to the taproot finalizer
- more unit tests for `psbt` and `p2tr`
- various refactorings

Note: currently investigating the failed integration test (regression)

motorina0:
**Updates**:
 - update the PR **description** and hide old comments (make it easier for reviewers to understand the changes)
 - **script-path** spend example with this lib on `testnet` ([code sample and transaction links](https://gist.github.com/motorina0/e6bb1b32b3d5d2fc80e1bd63d5efea4f))
 - open [small bug bounty](https://twitter.com/motorina0/status/1504067431622488067?s=20&t=Q9--GB-m9EI547EWojnXkA) (retweets help)

ts_src/payments/testecc.ts
```diff
@@ -0,0 +1,74 @@
+import { TinySecp256k1Interface } from '../types';
```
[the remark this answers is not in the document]
motorina0: - calling `testEcc(...)` is actually done on purpose. The reason is to do a basic check on the `ecc` library.
- if a bad/missing `ecc` lib is used, then the validation an tweaking will be affected.
- maybe `verifyeEcc()` would have been a better name.

ts_src/payments/taprootutils.ts
```diff
@@ -0,0 +1,144 @@
...
+
+/**
+ * Build the hash tree from the scripts binary tree.
+ * The binary tree can be balanced or not.
+ * @param scriptTree - is a list representing a binary tree where an element can be:
+ *  - a taproot leaf [(output, version)], or
+ *  - a pair of two taproot leafs [(output, version), (output, version)], or
+ *  - one taproot leaf and a list of elements
+ */
+export function toHashTree(scriptTree: TaprootLeaf[]): HashTree {
```
louisinger, line 48: it seems that `scriptTree` type should be `Array<[TaprootLeaf, TaprootLeaf] | TaprootLeaf>` ?

ts_src/payments/testecc.ts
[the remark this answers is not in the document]
louisinger: @tiero I think it checks the implementation of the injected ecc lib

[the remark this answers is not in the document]
tiero: Oh I see the need now (never been used before)

I personally prefer `verifyeEcc` indeed, is more explicit and immediate

ts_src/payments/taprootutils.ts
motorina0: Solved, thanks!

ts_src/payments/testecc.ts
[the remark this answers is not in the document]
motorina0: Done. Renamed `testecc` to `verifyecc`. Not to be confused with unit tests and integration tests.

brandonblack:
I reviewed this other than psbt.ts.

It was looking difficult to make PR comments for my feedback, so I opened a PR against this branch with my suggestions: https://github.com/bitcoincoretech/bitcoinjs-lib/pull/4

I'll continue with reviewing PSBT on Monday.

brandonblack:
From an API perspective, I notice that neither script-only nor key-only `p2tr`s are automatically constructed in the BIP-recommended way (ie. a user could specify a provably non-script hash, or an internal key with no known secret key, but the library doesn't help). The BitGo version generates values for internal key and hash if no script or no key are specified (still not 100% in the recommended way, but closer).

I was inclined to accept this behavior as "too much magic", but [this comment](https://github.com/bitcoincoretech/bitcoinjs-lib/pull/4#issuecomment-1074521771) made me wonder if it's appropriate to protect users more, in the BIP-recommended way.

[3rd and 4th bullets](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#constructing-and-spending-taproot-outputs) for reference.

ts_src/payments/p2tr.ts
```diff
@@ -0,0 +1,349 @@
...
+        if (!getEccLib().isXOnlyPoint(internalPubkey))
+          throw new TypeError('Invalid internalPubkey for p2tr witness');
+
+        const leafVersion = controlBlock[0] & TAPLEAF_VERSION_MASK;
+        const script = witness[witness.length - 2];
+
+        const leafHash = tapleafHash({ output: script, version: leafVersion });
+        const hash = rootHashFromPath(controlBlock, leafHash);
+
+        const outputKey = tweakKey(internalPubkey, hash);
```
brandonblack, line 302: Two comments on tweaking:
1) The tweaked key should be cached (probably just cached as `o.publicKey` and `o.parity`
2) Relatedly, the user should be able to specify `a.parity` if they are performing the tweak externally.

src/payments/p2tr.js
```diff
@@ -0,0 +1,308 @@
...
+  });
+  lazy.prop(o, 'internalPubkey', () => {
+    if (a.internalPubkey) return a.internalPubkey;
+    const witness = _witness();
+    if (witness && witness.length > 1)
+      return witness[witness.length - 1].slice(1, 33);
+  });
+  lazy.prop(o, 'signature', () => {
+    if (a.signature) return a.signature;
+    if (!a.witness || a.witness.length !== 1) return;
```
brandonblack, line 142: I believe technically a single sig witness is allowed to have an annex, so this should use _witness().

ts_src/payments/p2tr.ts
```diff
@@ -0,0 +1,349 @@
...
+  });
+  lazy.prop(o, 'internalPubkey', () => {
+    if (a.internalPubkey) return a.internalPubkey;
+    const witness = _witness();
+    if (witness && witness.length > 1)
+      return witness[witness.length - 1].slice(1, 33);
+  });
+  lazy.prop(o, 'signature', () => {
+    if (a.signature) return a.signature;
+    if (!a.witness || a.witness.length !== 1) return;
```
brandonblack, line 152: I believe technically a single sig witness is allowed to have an annex, so this should use _witness().

ts_src/payments/p2tr.ts
```diff
@@ -0,0 +1,349 @@
...
+  typef(
+    {
+      address: typef.maybe(typef.String),
+      input: typef.maybe(typef.BufferN(0)),
+      network: typef.maybe(typef.Object),
+      output: typef.maybe(typef.BufferN(34)),
+      internalPubkey: typef.maybe(typef.BufferN(32)),
+      hash: typef.maybe(typef.BufferN(32)), // merkle root hash, the tweak
+      pubkey: typef.maybe(typef.BufferN(32)), // tweaked with `hash` from `internalPubkey`
+      signature: typef.maybe(typef.BufferN(64)),
```
brandonblack, line 43: Can be 64 or 65 bytes.

brandonblack:
I'm working on a couple-few changes to this PR:
* Taptree should be `Payment | [Taptree, Taptree]` so that future payment types that would work as Taproot `redeem` can be put straight in. This will mean that certain operations will have to check `leaf.output` and fail if absent, but I think it's still the right call.
* Adding `isCanonicalSchnorrSignature` based on the BitGo Taproot to address some level of correct validation of `signature` on both p2tr and future Taproot `redeem`s.
* Adding a `p2tr_ns` from the BitGo Taproot, because its existence is helping to drive these API changes, and I think having such a Payment as both an example and a constraint on the API is going to be useful.
* Still trying to fully digest the PSBT code and see if it might make sense to expose `controlBlock` and/or `annex` from `p2tr`.

ts_src/psbt.ts
```diff
@@ -642,14 +665,31 @@ export class Psbt {
...
-      {
+    const scriptType = this.getInputType(inputIndex);
+
+    if (isTaprootSpend(scriptType)) {
+      if (!keyPair.signSchnorr) {
+        throw new Error(
+          `Need Schnorr Signer to sign taproot input #${inputIndex}.`,
+        );
+      }
+      const partialSig = this.data.inputs[inputIndex].partialSig || [];
```
brandonblack, line 676: `PartialSig` 0x02 can't be reused for Taproot -- we need to implement `TapScriptSig` 0x14 :(

`PartialSig` is specifically an XY pubkey and ECDSA signature.

This problem shows up when trying to improve the `test/psbt.utils.ts` `buildTapscriptFinalizer` to properly sequence signatures to match the script, and also would come up if a PSBT happened to have signatures corresponding to potentially multiple different spend paths.

junderw: bip174 library is pretty modular so adding TapScriptSig should be trivial.

brandonblack: Cool. I'm working on adding all of the BIP371 extensions (they're part of BIP174 since January, so this seems appropriate).

ts_src/payments/p2tr.ts
motorina0: It looks like [option 3 was chosen](https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki#design), has something changed?


> Encoding R and public key point P There exist several possibilities for encoding elliptic curve points:
> 
>     1. Encoding the full X and Y coordinates of P and R...
>     2. Encoding the full X coordinate and one bit of the Y...
>     3. Encoding only the X coordinate, resulting in 32-byte public keys and 64-byte signatures.
> 
> Using the first option would be slightly more efficient for verification (around 10%), but we prioritize compactness, and therefore choose option 3. 

Also in the [bitcoin core secp256k1](https://github.com/bitcoin/bitcoin/blob/master/src/secp256k1/include/secp256k1_schnorrsig.h#L110):

```c
/** Create a Schnorr signature.
* ...
*   Out:   sig64: pointer to a 64-byte array to store the serialized signature.
SECP256K1_API int secp256k1_schnorrsig_sign(
```

ts_src/payments/p2tr.ts
motorina0: done!

ts_src/payments/p2tr.ts
brandonblack: It can include an optional `SIGHASH_TYPE`. A 64-byte signature encodes `SIGHASH_DEFAULT` => `0x00`. A 65-byte signature encodes an explicit `SIGHASH_TYPE`.

motorina0: this should be included as a precondition for the `isCanonicalSchnorrSignature` check

ts_src/payments/p2tr.ts
motorina0: > 2. Relatedly, the user should be able to specify a.parity if they are performing the tweak externally.

I'm trying to think of a scenario where the `a.publickey` and `a.parity` are provided (but `a.internalPublickey` not).
Will the `parity` be used in any computation of `o`?

brandonblack: > I'm trying to think of a scenario where the a.publickey and a.parity are provided (but a.internalPublickey not).
If I perform the tweaking operation outside of `p2tr` eg. for MuSig. I will still provide `internalPubkey` but I don't want `p2tr` to tweak too. I guess it doesn't hurt fundamentally though.
> Will the parity be used in any computation of o?
To produce the correct control block.

motorina0: > ...but I don't want p2tr to tweak too. I guess it doesn't hurt fundamentally though

- this looks more like a performance optimization, we should not change the public API for it

> 1. The tweaked key should be cached...

- agree, I will look into it!

> To produce the correct control block.

- yes, of course. But `internalPublickey` is still required for the control block.

brandonblack: > this looks more like a performance optimization, we should not change the public API for it

I'm partially convinced by this reasoning. Will see what shakes out as I integrate with this, but it might be OK as is.

junderw:
https://github.com/jonasnick/bips/blob/musig2/bip-musig2.mediawiki

This is interesting.

brandonblack:
Ah, nice! I've been helping review that doc in the secp256k1-zkp repo. I have an implementation in JS that works with both noble and tiny secp here: https://github.com/brandonblack/musig-js

junderw:
> I'm kinda hoping to convince you to adopt the project into the bitcoinjs umbrella

lol, let's get generic taproot out first before we start thinking about that.

tiero:
Is there any actionable move we can take to bring this forward? Testing, ACKing etc..?

Also what about sort of "PR Review Club" call sometime in the next weeks?

junderw:
I published bip174 v2.1.0 which will be used here.

brandonblack:
> Is there any actionable move we can take to bring this forward? Testing, ACKing etc..?

Right now, there's a known issue with this PR, which is that it puts Taproot values into non-Taproot compatible fields of PSBTs. I added (and @junderw released) support for the Taproot fields in bip174, and I've started work on making use of them, but unfortunately doing it with the right fields is significantly more complex than the changes to psbt.ts in this PR. If anyone else wants to take that work on instead of me, I won't be offended in the slightest :)

junderw:
TBH, since PSBT itself has separate areas for taproot data, we should just separate the callstacks for taproot and non-taproot.

ie. If I call sign, it won't even accept or check taproot, and explicitly reject it as a possibility. But if I call signTaproot (or whatever we call it) it will not sign if the meaningful script is non-taproot.

This is quite a large amount of work to refactor.

brandonblack:
> TBH, since PSBT itself has separate areas for taproot data, we should just separate the callstacks for taproot and non-taproot.

Interesting, so `signAll*` would take 1 or 2 signers (2 if needing to sign a mixed TX) and the `signAll*` would inspect call the right `sign` depending on the input type?

junderw:
... *sigh* to be honest just having only signInput for taproot is enough for now... lol

`signInputTaproot` etc.

motorina0:
I have updated this branch to use the `BIP 371` fields added by @brandonblack [here](https://github.com/bitcoinjs/bip174/pull/29/files).
- the taproot logic is separate from the non-taproot logic (make it easier to review)
  - `if` (**isTaprrot**) **doTaprootStuff()** `else` **doOldStuff()**
- some taproot methods (eg: `signTaprootInput`) are also made public in order to offer taproot specific logic (sign only a particular leaf)
- in order to keep the `psbt.ts` module readable:
   - taproot specific utils have been extracted to `psbt/bip371.ts`
   - some non-taproot specific utils have been extracted to `psbt/psbtutils.ts`
- in a next step `psbt.ts` module should be moved to `psbt/index.ts` (kept it like this for now in order to easier view the diffs)
- this changes cover all taproot input fields except `tapBip32Derivation`
- this changes do NOT cover the taproot output specific logic (only the one contained in the `bip174` lib)
- full commit history can be found on [this branch](https://github.com/bitcoincoretech/bitcoinjs-lib/commits/p2tr-v1-psbt-v1)

Looking forward for reviews & feedback!

test/integration/taproot.spec.ts
```diff
@@ -393,20 +416,25 @@ describe('bitcoinjs-lib (transaction with taproot)', () => {
...
     psbt.signInput(0, leafKeys[2]);
+    psbt.signInput(0, leafKeys[0]);
 
-    const tapscriptFinalizer = buildTapscriptFinalizer(
-      internalKey.publicKey,
-      scriptTree,
-      regtest,
-    );
-    psbt.finalizeInput(0, tapscriptFinalizer);
+    psbt.finalizeInput(0);
```
tiero, line 437: what do you think to add a test case where we need to add extra witness data on the stack to spend the input?

ie. a very dumb `OP_ADD 3 OP_EQUAL`
My understanding is that we need to pass custom finalizer function?

junderw: Might be good to create a simple template, like 3-4 branches with OP_ADD {N} OP_EQUAL where N=3..6 or something.

Then have a few copy-pasted tests that try spending each of the paths.

Maybe the templates in csv and cltv integration tests can be of help? (maybe not)

motorina0: By popular demand, a taproot finalizer example has bee added. (see `...- custom finalizer ` test)
It build a tree with 8 leaves and then tries to spend them.

However, there is an issue with it and I could not figure out how to fix it.
When broadcasting the tx I get the error: `non-mandatory-script-verify-flag (Operation not valid with the current stack size)`
The final witness script looks correctly constructed:
image

Script: `OP_ADD OP_2 OP_EQUAL`
Witness: `OP_1 OP_1`




Any ideas what I'm doing wrong?

brandonblack: Can you paste the whole txhex that fails to broadcast in a copy-pasteable way?

motorina0: Sure, this is in [this link](https://bitcoincore.tech/apps/bitcoinjs-ui/index.html?tx_hex=02000000000101046c90993098a060bf9e1872067bcd0f12340cc902dda5b1095db1af8561ea910000000000ffffffff01904106000000000022512005a793464e0c3bd5f299a3712779f3ee2291d2d4246f830eb742343fc841a166030251510393528781c1bf8b8edaafa09256a30c1990bc822e506e79f74f01f1026c7f5041d26b13f5b1bbf3102f5a885ca9dc13fd5b6f54a40c409081b9702bae4ba4419fbcfe00a7b156491e8e3fd1f0051468b49dad9f9ccefd9e41b88b82c4aac491f967857f116999401a2485c859a8e2ab5f01570896db5efc9a7a98033be66f5216f4d96fba8000000000) url. You can copy the hex from the URL or use `Export`.

The partial/failed integration test is also committed on this branch.

brandonblack: Ah, I see the issue -- you have the 2 OP_1s in a single witness element instead of 2 separate elements.

junderw: I verified my solution works. I fixed the test and some test lints.

junderw (dismissed):
Great job. Looks good to me.

I would like one more review from @brandonblack if possible before merging.

brandonblack:
I have time scheduled to review today or tomorrow!

src/types.d.ts
```diff
@@ -10,6 +10,29 @@ export declare function Signer(obj: any): boolean;
 export declare function Satoshi(value: number): boolean;
 export declare const ECPoint: any;
 export declare const Network: any;
+export interface XOnlyPointAddTweakResult {
+    parity: 1 | 0;
+    xOnlyPubkey: Uint8Array;
```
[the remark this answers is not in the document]
motorina0: Yes, `XOnlyPointAddTweakResult` is used in:

```ts
export interface TinySecp256k1Interface {
  xOnlyPointAddTweak(
    p: Uint8Array,
    tweak: Uint8Array,
  ): XOnlyPointAddTweakResult | null;
```

`TinySecp256k1Interface` is a subset of the exposed interface of `tiny-secp256k1`, it is declared [here](https://github.com/bitcoinjs/tiny-secp256k1/blob/master/src_ts/index.ts#L259)

src/payments/taprootutils.d.ts
```diff
@@ -0,0 +1,42 @@
+/// <reference types="node" />
```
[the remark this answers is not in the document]
motorina0: Fine with me. If nobody else objects, I will rename it.

brandonblack:
Amazing work here @motorina0 !!

I've got a few nits I'm working through, and adding some additional PSBT test cases.

Thoughts on [this comment](https://github.com/bitcoinjs/bitcoinjs-lib/pull/1742#issuecomment-1074722881)?

junderw:
> Amazing work here @motorina0 !!
> 
> I've got a few nits I'm working through, and adding some additional PSBT test cases.
> 
> Thoughts on [this comment](https://github.com/bitcoinjs/bitcoinjs-lib/pull/1742#issuecomment-1074722881)?

This is the "use hash of key spend key as the script spend hash" thing, right?

brandonblack:
Yep

brandonblack:
There's a bug relating to this actually. I see that @motorina0 did start tweaking the key with a hash of itself by default, but that's not correctly handled in PSBT, because if that's done, the hash of the key should be specified in the `tapMerkleRoot`. I fixed the bug in order to get my new test vector working, but then it broke on other vectors. I'll see if I can get a PR up for this today.

I'm also working on adding HD signing support with `tapBip32Derivation`, which helps inform some of this. Looks like we'll need to change the `*Signer*` interfaces to allow for tweaked signing, because we should be matching the internal key to the signer's public key and then doing tweaked signing. (Here's Core's interface for schnorr signing: `bool CKey::SignSchnorr(const uint256& hash, Span<unsigned char> sig, const uint256* merkle_root, const uint256& aux) const`)

tiero:
> because now it's up to the signer to know how to apply a tweak, so we're not doing any elliptic curve stuff in PSBT

I am 100% in favor in this: in other languages/libraries usually the PSBT packages only accept a signature to be attached by the updated/signer role (ie. btcutil/btcd golang). I find it totally fine to expect either an instance of an abstract interface  or simply just expecting the partialSig/tapKeySig/tapScriptSig as input should be enough

junderw:
iirc there were discussions about making an interface... I think we decided against it for some reason.

I forget the reasoning... maybe @motorina0 remembers.

Maybe it was a problem with the implementation and not the idea...

I'm open to any ideas that make things easier for consumers of the library to not shoot themselves in the foot.

junderw:
On the other hand, this PR is getting large, and if it's logically consistent and stable, we could just merge it, but hold off on a release until we get a few more QoL features in.

I would like to make this a minor release though, so no breaking changes.

motorina0:
@brandonblack regarding: [this comment](https://github.com/bitcoinjs/bitcoinjs-lib/pull/1742#issuecomment-1074722881)
- [bullet point 3 (BIP341 spec)](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#constructing-and-spending-taproot-outputs):  if no no internal key is provided then `extract`/`aggregate`/`derive from constant`
      - if someone decides to implement this logic then I suggest adding it to `bip341.ts` (former `taprootutils.ts`) and expose it in the lib API
      - the users of the lib can then compute this internal key if they choose to and pass it to `p2tr`
      - this requirement is not a blocker for the current PR
      
- [bullet point 4](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#constructing-and-spending-taproot-outputs): 
   - in `p2tr`: the internal key is tweaked with a hash of itself if no `scriptTree` or `hash` is provided
   - in `psbt`: 
      - `input` - indeed, it is a good idea to add a validation that `tapInternalKey` + `tapMerkleRoot` => `witnessUtxo.script`
         - it is not mandatory for `tapMerkleRoot` to be the valid root of a `tapTree`. In the case of  [Silent Payments](https://gist.github.com/RubenSomsen/c43b79517e7cb701ebf77eec6dbb46b8) this tweak value is in fact a shared secret
      - `output` - anything to change here?

motorina0:
> Maybe it was a problem with the implementation and not the idea...

This is the `ecpair` [PR with the discussion](https://github.com/bitcoinjs/ecpair/pull/6). TL;DR
The open question is how to pass the `hash` for the `tweak`. There are several options:
 1. add an optional parameter to `signSchnorr` (the above open PR does this)
    - `signSchnorr(hash: Buffer, extraEntropy?: Buffer, tweak?: Buffer)` 
 1. add a new optional method on `ECPairInterface`
    - `tweak?(t: Buffer): ECPairInterface;`
    - `signSchnorr()` was intentionally added to `ECPairInterface` instead of `Signer` ([see thread](https://github.com/bitcoinjs/ecpair/pull/4#issuecomment-994135318)). Maybe `[sign|verify]Schnorr()` can be moved to `Signer` interface now (@junderw ?).
 1. ~~add a new `tweak?: Buffer` field to `ECPairOptions` (tweak at init phase)~~

All 3 options work, but have their tradeoffs:
- **Option 1**:
  - minimal interface change
  - Q: how can check inside `psbt` if the provided `Signer` supports tweaking or not (passing the `tweak` param is not a noop)
- **Option 2**:
  - a new method is added to the interface
  - a new instance of `Signer` is produced on `tweak()`
- **Option 3**:
  - excluded, requires knowledge of the `ECPair` API


**Notes**: 
 - only the `key-path` spend requires tweaking,  for `script-path` the plain signer is used.
 - for Options 1&2 the user of the lib can pass either a tweaked or a non-tweaked `Signer` and the `psbt` logic can detect if it needs tweaking (provided the either `tapInternalKey` or `witnessUtxo.script` are present)
    - easier for the consumer of the library
 - `ecpair` and `bip32` should follow the same aproach for tweaking

src/payments/taprootutils.d.ts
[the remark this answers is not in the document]
motorina0: I will rename this after @brandonblack submits his changes (to avoid conflicts)

ts_src/payments/index.ts
```diff
@@ -17,11 +19,14 @@ export interface Payment {
...
   input?: Buffer;
   signatures?: Buffer[];
+  internalPubkey?: Buffer;
   pubkey?: Buffer;
   signature?: Buffer;
   address?: string;
   hash?: Buffer;
   redeem?: Payment;
+  redeemVersion?: number;
+  scriptTree?: Taptree;
```
motorina0, line 29: Should `scriptTree` be renamed to `tapTree`  (for `psbt` consistency sake)? Please vote:
 - 👍 (yes)
 - 👎 (no)
- 😕  (doesn't matter)

tiero: 😕 (doesn't matter)

brandonblack:
> we use eccLib also for xOnlyPointAddTweak() and isXOnlyPoint()

Right, and I think that requiring either `eccLib` or `isXOnlyPointFn` to be passed into `payments.p2tr` is fine. `initEccLib` was added to avoid threading an `eccLib` through `psbt.ts`, which would not be required now. The only place in `psbt.ts` that calls `payments.p2tr` is in finalizing keypath spend, which doesn't require `isXOnlyPoint`.

Seems like the right move is to add a function like `make bip341Signer` on `bip341.ts` which takes an eccLib, and a `Bip32` or `ECPair` and returns a signer that accepts an optional tweak for use with PSBT (probably 2 functions, one for async and one for sync).

motorina0:
> The only place in psbt.ts that calls payments.p2tr is in finalizing keypath spend

I'm not sure how `xOnlyPointAddTweak()` can be removed without removing several integrity checks.
This is roughly the dependency tree I could build:

- `xOnlyPointAddTweak()`
  - `tweakKey()`
    - `bip371.tweakInternalPubKey()`
      - `psbt.getTaprootHashesForSig()`
      - to be added on check for `witnessUtxo.script.equals(OP_1 + tweakKey(tapInternalKey))`
    - `p2tr.o.pubkey` <- provided the internalKey, compute the output key
      - used in `bip371` for different checks 
    
Removing `initEccLib()/getEccLib()` at a later stage (if possible) would not be a breaking change as it relaxes the conditions.

src/payments/taprootutils.d.ts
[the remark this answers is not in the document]
motorina0: - [x] rename `taprootutils` to `bip341` (done)

brandonblack:
> * xOnlyPointAddTweak()
>     * tweakKey()
>         * bip371.tweakInternalPubKey()
>             * psbt.getTaprootHashesForSig()

That's the path that I think should be removed, because it's only needed if we allow passing a pre-tweaked signer. If we restrict the user to passing a signer that can be tweaked (option 1 or 2 from the ecpair issue) then there is no need to use `getEccLib` in PSBT.

motorina0:
> That's the path that I think should be removed

- we can explore that after the merge in `master` (and before publish maybe)
- this PR already requires a lot of effort to effort to keep up with all the changes

brandonblack:
@motorina0 I think I'm missing something. I don't see why we would need to remove those if we stop using `eccLib` in `psbt.ts`. Can you link me to the point in code that I'm missing?

I like the idea of potentially merging for better collaboration. I think the only thing I'd want to see before that is do no tweaking at all inside `psbt.ts`, so that a plain `internalKey` === `output.script.slice(2)` case works. In short, let's start with doing less if we're going to merge something that we know needs more work.

brandonblack:
Gotcha, thank you. PSBT should support people who do $Q = P$ and those who do $Q = P + hash_{TapTweak}(P){\cdot}G$ - the current way of doing `checkTaprootOutputFields` can only really support one or the other. That is somewhat neither here nor there, but I think it might be a smell indicating that correctness of the `script` provided relative to the other fields is not something we should be trying to verify at that point. We can actually tell that the designer of BIP371 didn't intend this validation because they didn't include a `PSBT_OUT_TAP_MERKLE_ROOT`.

brandonblack:
> that would have been redundant info since it can be computed from PSBT_OUT_TAP_TREE

No, it wouldn't, because it's needed to differentiate between $Q = P$ and $Q = P + hash_{TapTweak}(P){\cdot}G$.

> not sure I follow. Do you mean for tapInternalKey, or Signer, or...?

I mean in `checkTaprootOutpuFields` for a key-only taproot output, that function can only do one of the two ways of validating that the output is correct relative to the internal key.

motorina0:
How about we do all the tweak checks in `extractTransaction()`

- change from
```ts
extractTransaction(disableFeeCheck?: boolean): Transaction
```

- to
```ts
interface ExtractTransactionOpts {
  disableFeeCheck?: boolean;
  disableEccCheck: boolean = false;
  eccLib?: TinySecp256k1Interface
}

extractTransaction(opts?: ExtractTransactionOpts): Transaction
```

- if the user wants to `disableEccCheck` it must do this explicilty
- if eccCheck is enabled and `eccLib` is provided, then do as much validation as possible on tweaks and xOnly points

brandonblack:
> if eccCheck is enabled and eccLib is provided, then do as much validation as possible on tweaks and xOnly points

Sounds good. If `PSBT_OUT_TAP_INTERNAL_KEY` is provided and `PSBT_OUT_TAP_TREE` is not, then we can check `script.slice(2).equals(tapInternalKey) || script.slice(2).equals(selfTweak(tapInternalKey))`.

motorina0:
image

motorina0:
@junderw can we get [your 2 sats](https://legend.lnbits.com/lnurlp/1761) on [this comment](https://github.com/bitcoinjs/bitcoinjs-lib/pull/1742#issuecomment-1144585153)

motorina0:
Opened PR on `ecpair` for [add tweak() method](https://github.com/bitcoinjs/ecpair/pull/8)
Please leave your comments. If all good it will be added to `bip32` so `signHD()` can be implemented here.

junderw:
> @junderw can we get [your 2 sats](https://legend.lnbits.com/lnurlp/1761) on [this comment](https://github.com/bitcoinjs/bitcoinjs-lib/pull/1742#issuecomment-1144585153)

Breaking change.

Maybe offer a separate extractTransactionV2 or something.

junderw:
Also, feel free to ping me on Matrix or Telegram or whatever.

You can find me in the `#bitcoinjs-dev:matrix.org` room on Matrix.

junderw:
My Github notifications are a mess, and it keeps bugging out and marking things as read even when they're not read...

Mix that with being busy a lot, and my response times are poor.

I tend to find that the notification bugs tend to happen with longer running issues/PRs with lots of comments, so I tend to be quicker with simple issues that new people send.

mahnunchik:
Any news?

junderw:
> Any news?

Reviews are welcome. Right now I think the status is:

This is a big change, and every time it seems like it's about to be merged an issue of discussion pops up, then everyone gets busy with other things.

Having more people that review the code and give comments to move things along would help.

Thanks.

DmytroShalaiev:
when review?

junderw:
Let's merge this for now.

1. Some open comments from @antonilol, fix any you feel need fixing. @motorina0 
2. Fix merge conflicts

I'll merge it to master for now........ then I'll publish a version for `v.6.1.0.rc.0` and if no bugs come in after a couple weeks I'll push v6.1.0

junderw:
After fixing for merge, maybe run `npm version preminor --preid=rc` to add the version commit for `v6.1.0.rc.0` so we can do it in one go.

ts_src/psbt/bip371.ts
```diff
@@ -0,0 +1,458 @@
...
+  newOutputData: PsbtOutput,
+): void {
+  if (!newOutputData.tapTree && !newOutputData.tapInternalKey) return;
+
+  const tapInternalKey =
+    newOutputData.tapInternalKey || outputData.tapInternalKey;
+  const tapTree = newOutputData.tapTree || outputData.tapTree;
+
+  if (tapInternalKey) {
+    const { script: scriptPubkey } = outputData as any;
```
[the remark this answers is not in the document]
motorina0: This requires changes to `bip174`, can be done in a separate smaller PR.

ts_src/payments/bip341.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  return {
+    parity: res.parity,
+    x: NBuffer.from(res.xOnlyPubkey),
+  };
+}
+
+function tapBranchHash(a: Buffer, b: Buffer): Buffer {
+  return bcrypto.taggedHash('TapBranch', NBuffer.concat([a, b]));
+}
```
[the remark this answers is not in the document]
motorina0: - both approaches work, however in at least one context the sorting is required also before `tapBranchHash` is called. This would result in sorting twice.
- it can be handled in a different smaller PR

motorina0:
@antonilol: thank you for your inputs. I have made the changes you suggested (except 2 of them, see comments)
@junderw:
 - [x] rebase master and solve conflicts
 - [x] bump version

junderw (approved):

ts_src/types.ts
```diff
@@ -65,6 +67,48 @@ export const Network = typeforce.compile({
...
+
+/**
+ * Binary tree repsenting script path spends for a Taproot input.
+ * Each node is either a single Tapleaf, or a pair of Tapleaf | Taptree.
+ * The tree has no balancing requirements.
+ */
+export type Taptree = [Taptree | Tapleaf, Taptree | Tapleaf] | Tapleaf;
+
+export function isTaptree(scriptTree: any): scriptTree is Taptree {
+  if (!Array(scriptTree)) return isTapleaf(scriptTree);
```
OttoAllmendinger, line 97: was this supposed to be `Array.isArray()`? This is calling the constructor and always returns a truthy value

OttoAllmendinger: nevermind - we have a `const Array = ` further down

PR #1786, opened 2022-03-23
Do real secp256k1 point->curve checking

brandonblack:
* This is a breaking change, as it requires the JS environment to have BigInt (all supported releases that I can find do).
* This check may prevent loss of funds by eliminating a category of unspendable addresses from being created.
* Performance is almost as fast as tiny-secp256k1 39-42us vs 33-35us.


brandonblack:
A note on performance: `tiny-secp256k1` builds a full point to check the curve equation, which requires performing a modular square root. This check avoids performing the square root by using the Jacobi Symbol to check that it can be done, which is sufficient to say that the Y coordinate exists, and therefore the X coordinate is on the curve.

brandonblack:
Also of note: I confirmed the accuracy of this function by comparing it to `tiny-secp256k1.isXOnlyPoint` over > 1 million random 32-byte arrays.

junderw:
1. Needs careful consideration.
2. Needs tests.

junderw:
3. If we're going to use bigint, a notable addition that should also be considered is the satoshi unit. This would also be a breaking change, but allow for forks of bitcoin that have extremely high amounts (ie. doge) to be handled with just a network object like litecoin.

brandonblack:
1. Agreed, figured opening a proper top-level PR was the place for that consideration, once I found I was able to bring the performance in line with what is proposed for point checking in the Taproot branch.
2. Done -- LMK if there are other particular cases you'd like to see covered.
3. You know that we at BitGo wouldn't complain about making Doge support easier :-D

junderw:
Also, this exposes the methods from the public API via crypto. (we export crypto directly as is)

I don't want to export these. (which is why types is not exported)

brandonblack:
Would you like them in a new file, in types, or another file?

junderw:
In types is fine.

motorina0:
**NACK** on:
 > This is a breaking change, as it requires the JS environment to have BigInt
- this change will make `bitcoinjs-lib` unusable as-is in some environments (I'm thinking of React Native on mobile)
   - disclaimer: I work at `Exodus`, this would be indeed a breaking change :(

motorina0:
> This check may prevent loss of funds by eliminating a category of unspendable addresses from being created.
  - I totally agree with this stricter check, the question is only if this logic should be part of `bitcoinjs-lib`
  - the hole point of having `TinySecp256k1Interface` was to have the `ECC` logic (imported lib) decoupled from this lib and to allow the users to provide their own implementation

**My Suggestion**:
 - add `isPoint()` to `TinySecp256k1Interface`
    - note: the current `types.isPoint()` is used in all payments (`p2pk`, `p2pkh`, `p2wpkh` ...)
 - once we have [initEccLib](https://github.com/bitcoincoretech/bitcoinjs-lib/pull/4#issuecomment-1074887907) in place:
   - the `payments` will use the `isPoint()` of the `eccLib` (if provided)
   - or will default to the existing (less strict) `isPoint()` otherwise
 - this way the user of the lib can enforce stricter checks and there is no backwards compability issue

brandonblack:
> this change will make bitcoinjs-lib unusable as-is in some environments (I'm thinking of React Native on mobile) 

Interesting, what version of which JS engine is that?

junderw:
> > this change will make bitcoinjs-lib unusable as-is in some environments (I'm thinking of React Native on mobile)
> 
> Interesting, what version of which JS engine is that?

React: aka FacebookJS, aka ES5-or-bust

brandonblack:
And now that I've spelunked into the odd world of mobile device support cycles, I see that a version of iOS that didn't have BigInt in its JS engine was released as recently as September, 2021, and so using BigInt in bitcoinjs-lib should probably wait until that's aged out a bit more. :cry:

junderw:
Not sure if I want to ask for a rebase (we merged taproot yay!!!) or just close this...

If we're gonna jump to BigInt requirements, might as well compile everything to WASM from Rust and rewrite the entire library in Rust! (can you tell I've become a Rust evangelist since the beginning of this year??? lol)

Roscoe61nine (approved):

paulmillr:
React Native supported bigints since august. It doesn't support bigint literals 42n, so this PR is ok.

@brandonblack is jacobi symbol faster than sqrt bit-by-bit unroll?

junderw:
I noticed RN support for BigInt recently when I was checking for support when merging noble into bip32.

It's been almost half a year, but notably BlueWallet (one of our biggest/most popular dependents) is on an old version of RN that doesn't support it.

Once BlueWallet supports BigInt, I'd support merging this, in addition to switching the concept of "amount" (currently type `number`) to use `bigint` type. (Would have to change some of the runtime type checks to check for `0 <= x < 2**63`.)

re: @Overtorment mentioned needing to bump to RN70 in October.

https://github.com/BlueWallet/BlueWallet/pull/5073#issuecomment-1264759949

Overtorment:
Can confirm, BlueWallet is working on bumping RN to 70+. Not easy as we have some native dependencies lagging behind/unmaintained, so we are looking into workarounds 

cc @marcosrdz

Leonardbrett:
> React Native supported bigints since august. It doesn't support bigint literals 42n, so this PR is ok.
> 
> @brandonblack is jacobi symbol faster than sqrt bit-by-bit unroll?

This woman offers fast and legit investment. Connect with Charlotte on 🅦🅗🅐🅣🅢🅐🅐🅟+𝟭𝟰𝟳𝟱𝟮𝟰𝟮𝟲𝟯𝟯𝟴

Leonardbrett:
My first investment with Charlotte give me profits of over $7,000 and ever since then she has never failed to deliver and I can even say she's the must sincere broker I have ever known'' communicate with her to be a successful investor. On WhatsApp:+𝟭𝟰𝟳𝟱𝟮𝟰𝟮𝟲𝟯𝟯𝟴

Overtorment:
btw BW migrated to RN70+ and theoretically has BigInt now

jasonandjay:
Maintainer triage: this PR is pending re-validation against current `master` / v7 direction.

Please:
1) rebase (or merge) latest `master`
2) ensure full CI passes (`format`, `lint`, `lint-tests`, `gitdiff`, `unit`, `integration`, `coverage`, `build-doc`)
3) confirm this PR still matches current roadmap/design assumptions

If there is no update in ~14 days, we may close this PR for backlog hygiene. You can always reopen or submit a refreshed PR later.

PR #1815, opened 2022-06-13
Update setup-node action to use matrix and lts notation

junderw:


motorina0:
Just curious, why do `lint` and `lint-test` run against multiple `node` versions, but `format` does not?

junderw:
format is more of a "nice have" and the worst case of a mismatch is ugly code, whereas a lint mismatch could result in bad (incorrect) code.

I highly doubt there will ever be a mismatch, so it really doesn't matter to me, I just picked lint = yes, format = no.

The biggest gripe I have is the repitition of the matrix for each job... I wish there was a way to separate jobs into matrix based jobs and non-matrix jobs, and only write the matrix once.

junderw:
1. Reordered the jobs to have all matrix jobs at the top with comments separating them (so easier to update matrix later)
2. I removed the matrix from lint jobs

PR #1847, opened 2022-11-25
docs: add navigator line for each example method

Hyunhum:
Just added navigator line for dev to find out example methods easier


junderw:
Thanks for the PR.

We used to do it this way, but it became too much of a hassle to maintain when adding tests etc.

Closing as "known issue" and "do not fix"

junderw:
I would be interested in a way to automate the updates for these lines though... if that exists.

If there was some semi-automatic way (running an npm run script etc.) that could update this I would consider it.

PR #1859, opened 2022-12-29
chore: initialise TAGGED_HASH_PREFIXES on-demand

headfire94:
performance optimisation: loop through TAGS and calculating hash prefixes on library require takes some time. instead calculate it on-demand when `taggedHash` is called.


junderw:
If we will do this I think maybe instead we can just put the hashes in as "magic numbers" and add a test that will make sure these constants equal the correct hash.

package.json
```diff
@@ -39,7 +40,8 @@
     "prettier": "prettier \"ts_src/**/*.ts\" \"test/**/*.ts\" --ignore-path ./.prettierignore",
     "prettierjs": "prettier \"src/**/*.js\" --ignore-path ./.prettierignore",
     "test": "npm run build && npm run format:ci && npm run lint && npm run nobuild:coverage",
-    "unit": "npm run build && npm run nobuild:unit"
+    "unit": "npm run build && npm run nobuild:unit",
+    "generate:prefixes": "node scripts/generate-tagged-hash-prefixes.js && yarn prettierjs && yarn lint:fix"
```
junderw, line 44: (fyi, we don't have yarn as a dev dependency)

I was thinking more along the lines of just replacing the existing code area with an object definition with `as const` or something.

Then adding a test to the crypto test file that will iterate through all the keys, calculate the hash (in the test) and assert it's equal.

Also, if we're going to do magic numbers anyways, using numbers instead of hex is fine. (`Buffer.from([1, 2, 3, 4])` etc.)

So basically, just taking the output of this script (maybe without using hex) and just copy pasting over the existing code as a const, then adding a test.

Thanks.

headfire94: I was thinking that someone may want to add another tag in future and decided to provide as script to simplify the process. But we can go we hardcoded variables

headfire94:
I'm not sure why CI failed

motorina0:
> I'm not sure why CI failed

- `gitdiff` usually fails because not all transpiled files have been committed (`*.js`, `*.d.ts`)

junderw (approved):

PR #1874, opened 2023-02-15
Add example using BIP86 vector to verify the sending to and from a BIP86 generated taproot address

### LINKED ISSUE #1873: How to create a Taproot address from a mnemonic seed and then an output script example?
Can we please get an example of how to go from creating a seed/mnemonic to a taproot address and then a p2tr with that address?

It would be a huge help to see it go from end to end and then finally spend that output.

Thanks in advance

junderw:
Closes #1873

This example will help people figure out how to use BIP86 derivation to use taproot.


BasedOrg:
Wonderful, thank you!

PR #1891, opened 2023-02-27
fix: android `TypedArray.of requires its this argument to subclass a TypedArray constructor`

fboucquez:
Hi team, thank you for providing and maintaining this library.

Physical android devices don't like the `Buffer.of(num)`. We are noticing `TypedArray.of requires its this argument to subclass a TypedArray constructor` signing taproot utxos. 

This could be an env problem poly filling `Buffer.of`, just that this pretty safe `Buffer.alloc` change saves this. In case other also have this problem


junderw:
Let's use Buffer.from([0x00])

fboucquez:
> Let's use Buffer.from([0x00])

Done

PR #1932, opened 2023-06-03
Fix: Find hashes WITHOUT leafHash instead

### LINKED ISSUE #1931: PSBT signature validation issue
Hello,

I am trying to test the validity of a signature on a PSBT.  In my test case the psbt has a **bad** signature but validateSignaturesOfAllInputs is returning true.  Am I doing something wrong?

My test code is as follows:
``` 
  protected schnorrValidator = (
      pubkey: Buffer,
      msghash: Buffer,
      signature: Buffer,
  ): boolean => {
    return ecc.verifySchnorr(msghash, pubkey, signature);
  }

...

    test() {
        const psbtBase64 = "cHNidP8BAFICAAAAAe1h73A6zedruNERV6JU7Ty1IlYZh2KO1cBklZqCMEy8AAAAAAD/////ARAnAAAAAAAAFgAUS0GlfqWSeEWIpwPwrvRIjBbJQroAAAAAAAEA/TgBAQAAAAABAnGJ6st1FIvYLEVbJMQaZ3HSOJnkw5C+ViCuJYiFEYosAAAAAAD9////xuZd0xArNSaBuElLX3nzjwtZW95O7L/wbz94v+v0vuYAAAAAAP3///8CECcAAAAAAAAiUSAVbMSHgwYVdyBgfNy0syr6TMaFOGhFjXJYuQcRLlpDS8hgBwAAAAAAIlEgthWGz3o2R7WpgjIK52ODoEaA/0HcImSUjVk6agZgghwBQIP9WWErMfeBBYyuHuSZS7MdXVICtlFgNveDrvuXeQGSZl1gGG6/r3Aw7h9TifGtoA+7JwYBjLMcEG6hbeyQGXIBQNSqKH1p/NFzO9bxe9vpvBZQIaX5Qa9SY2NfNCgSRNabmX5EiaihWcLC+ALgchm7DUfYrAmi1r4uSI/YaQ1lq8gAAAAAAQErECcAAAAAAAAiUSAVbMSHgwYVdyBgfNy0syr6TMaFOGhFjXJYuQcRLlpDSwEDBIMAAAABCEMBQZUpv6e1Hwfpi/PpglkkK/Rx40vZIIHwtJ7dXWFZ5TcZUEelCnfKOAWZ4xWjauYM2y+JcgFcVsuPzPuiM+z5AH+DARNBlSm/p7UfB+mL8+mCWSQr9HHjS9kggfC0nt1dYVnlNxlQR6UKd8o4BZnjFaNq5gzbL4lyAVxWy4/M+6Iz7PkAf4MBFyC6ZCT2zZVrEbkw/T1fyS8eLKQaP2MH6rzdlMauGvQzLQAA";

        const psbt = Psbt.fromBase64(psbtBase64);
        let valid: boolean;

        try {
            valid = psbt.validateSignaturesOfAllInputs(this.schnorrValidator)
        } catch (e) {
            valid = false;
        }
        console.log(`All signatures are valid: ${valid}`)
    }
```
and the result is
```
All signatures are valid: true
```

When I attempt to view this psbt in Sparrow wallet it correctly indicates that the signature is invalid.
sparrow

What am I missing?

junderw:
Fixes #1931 

This function should be switched to default to false and return early with true as well..

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 6.1.2
+__fixed__
+- validateSignaturesOfInput for taproot inputs returned true for invalid signatures in specific cases. (#1932)
+
 # 6.1.1
 __added__
 - add example using BIP86 vector to verify the sending to and from a BIP86 generated taproot address
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "bitcoinjs-lib",
-  "version": "6.1.1",
+  "version": "6.1.2",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bitcoinjs-lib",
-  "version": "6.1.1",
+  "version": "6.1.2",
   "description": "Client-side Bitcoin JavaScript library",
   "main": "./src/index.js",
   "types": "./src/index.d.ts",
--- a/src/psbt.js
+++ b/src/psbt.js
@@ -468,14 +468,16 @@ class Psbt {
           this.__CACHE,
         );
     if (!allHashses.length) throw new Error('No signatures for this pubkey');
-    const tapKeyHash = allHashses.find(h => !!h.leafHash);
+    const tapKeyHash = allHashses.find(h => !h.leafHash);
+    let validationResultCount = 0;
     if (tapKeySig && tapKeyHash) {
       const isValidTapkeySig = validator(
         tapKeyHash.pubkey,
         tapKeyHash.hash,
-        tapKeySig,
+        trimTaprootSig(tapKeySig),
       );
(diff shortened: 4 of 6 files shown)


junderw:
We should add a test case to fail on the PSBT from the other issue.

junderw:
Also, the PSBT supposedly signed using SINGLE | ANYONECANPAY... but those sighash types are not accounted for when validating.

The PSBT is kind of weird, the finalScriptWitness has the signature encoded as 0x01 + sighash type + 64 byte sig

But the keySig part only has the 64 byte signature.

The sighash type field was in the input for the PSBT though... we should read that when validating the signatures (since it changes the hash preimage)

motorina0 (approved):

junderw:
I only pushed the version bump since the review.

junderw:
Published as v6.1.2

PR #1944, opened 2023-06-20
docs: add reference for mobile integration

ChrisCho-H:
I think many moblie dev have troubles to integrate this lib, so I think it would be nice if there's a simple guidance to refer for it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,9 @@ Which you can then import as an ESM module:
 <script type="module">import "/scripts/bitcoinjs-lib.js"</script>
 ````
 
+### Mobile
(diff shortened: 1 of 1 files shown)


junderw:
This will imply some sort of support or help from the library maintainers.

If you feel that the search function is inadequate, perhaps you can make a new issue with a name and body designed to have good SEO for issues search for various terms, then close the issue just to have it as a guide for people who search. (If I saw a closed issue called "Guide to Mobile integration" I would probably assume there's some good solution in there since it's closed)

Thanks for the PR and for responding to the other issue, though.

PR #1972, opened 2023-09-01
Added support to handle custom async finalizers

summraznboi:
Currently you can only use synchronous custom finalizers when finalizing an input - it would be greatly useful to also support async finalizers as well. This PR is an attempt to implement that as simply as possible. Any feedback is greatly welcome :)


summraznboi:
Included in this PR are some quality of life updates to Typescript types, preferring to optional keys for finalScriptSig/finalScriptWitness so that users don't have to explicitly specify undefined for unused key.

junderw:
I'd be interested to hear a concrete use case.

Signing is obvious, with HSMs and separate signing devices requiring user input.

Finalizing seems like it lacks the need for async IO with some outside service / device...

summraznboi:
Actually, I agree, will close out now.

PR #1979, opened 2023-09-18
Upgrade bip174 to 2.1.1.

arik-so:
This fixes adding unknownKeyVals to outputs, which otherwise are being added to inputs.


junderw (approved):

junderw:
Published as v6.1.5

PR #1986, opened 2023-09-29
✨feature: add doc

jasonandjay:
Add bitcoinjs-lib documentation

- [x] Automatically read ts file and generate
- [x] Automatically triggered by github actions (github pages need to be enabled)

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,8 @@
     "prettier": "prettier \"ts_src/**/*.ts\" \"test/**/*.ts\" --ignore-path ./.prettierignore",
     "prettierjs": "prettier \"src/**/*.js\" --ignore-path ./.prettierignore",
     "test": "npm run build && npm run format:ci && npm run lint && npm run nobuild:coverage",
-    "unit": "npm run build && npm run nobuild:unit"
+    "unit": "npm run build && npm run nobuild:unit",
+    "doc": "npx typedoc"
   },
   "repository": {
     "type": "git",
@@ -53,6 +54,7 @@
     "bech32": "^2.0.0",
     "bip174": "^2.1.1",
     "bs58check": "^3.0.1",
+    "typedoc": "^0.25.1",
     "typeforce": "^1.11.3",
     "varuint-bitcoin": "^1.1.2"
   },
--- a/package.json
+++ b/package.json
@@ -54,7 +54,6 @@
     "bech32": "^2.0.0",
     "bip174": "^2.1.1",
     "bs58check": "^3.0.1",
-    "typedoc-plugin-bitcoinjs-runcase": "^1.0.1",
     "typeforce": "^1.11.3",
     "varuint-bitcoin": "^1.1.2"
   },
@@ -90,6 +89,7 @@
     "tiny-secp256k1": "^2.2.0",
     "ts-node": "^8.3.0",
     "typedoc": "^0.25.1",
+    "typedoc-plugin-bitcoinjs-runcase": "^1.0.1",
     "typescript": "^4.4.4"
   },
   "license": "MIT"
--- a/package.json
+++ b/package.json
@@ -89,7 +89,6 @@
     "tiny-secp256k1": "^2.2.0",
     "ts-node": "^8.3.0",
     "typedoc": "^0.25.1",
-    "typedoc-plugin-bitcoinjs-runcase": "^1.0.1",
     "typescript": "^4.4.4"
   },
   "license": "MIT"
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
     "bech32": "^2.0.0",
     "bip174": "^2.1.1",
     "bs58check": "^3.0.1",
+    "typedoc-plugin-bitcoinjs-runcase": "^1.0.1",
     "typeforce": "^1.11.3",
     "varuint-bitcoin": "^1.1.2"
   },
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "coverage-report": "npm run build && npm run nobuild:coverage-report",
     "coverage-html": "npm run build && npm run nobuild:coverage-html",
     "coverage": "npm run build && npm run nobuild:coverage",
+    "doc": "typedoc",
     "format": "npm run prettier -- --write",
     "formatjs": "npm run prettierjs -- --write",
     "format:ci": "npm run prettier -- --check && npm run prettierjs -- --check",
@@ -39,8 +40,7 @@
     "prettier": "prettier \"ts_src/**/*.ts\" \"test/**/*.ts\" --ignore-path ./.prettierignore",
     "prettierjs": "prettier \"src/**/*.js\" --ignore-path ./.prettierignore",
     "test": "npm run build && npm run format:ci && npm run lint && npm run nobuild:coverage",
-    "unit": "npm run build && npm run nobuild:unit",
-    "doc": "npx typedoc"
+    "unit": "npm run build && npm run nobuild:unit"
   },
   "repository": {
     "type": "git",
@@ -54,7 +54,6 @@
     "bech32": "^2.0.0",
     "bip174": "^2.1.1",
     "bs58check": "^3.0.1",
-    "typedoc": "^0.25.1",
     "typeforce": "^1.11.3",
     "varuint-bitcoin": "^1.1.2"
   },
@@ -89,6 +88,7 @@
     "rimraf": "^2.6.3",
     "tiny-secp256k1": "^2.2.0",
     "ts-node": "^8.3.0",
+    "typedoc": "^0.25.1",
     "typescript": "^4.4.4"
   },
   "license": "MIT"
--- a/.github/workflows/main_ci.yml
+++ b/.github/workflows/main_ci.yml
@@ -8,6 +8,9 @@ on:
 
 permissions:
   contents: read
+  pages: write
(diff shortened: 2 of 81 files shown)


package.json
```diff
@@ -53,6 +54,7 @@
     "bech32": "^2.0.0",
     "bip174": "^2.1.1",
     "bs58check": "^3.0.1",
+    "typedoc": "^0.25.1",
```
junderw, line 57: Should be dev dependency

package.json
```diff
@@ -39,7 +39,8 @@
     "prettier": "prettier \"ts_src/**/*.ts\" \"test/**/*.ts\" --ignore-path ./.prettierignore",
     "prettierjs": "prettier \"src/**/*.js\" --ignore-path ./.prettierignore",
     "test": "npm run build && npm run format:ci && npm run lint && npm run nobuild:coverage",
-    "unit": "npm run build && npm run nobuild:unit"
+    "unit": "npm run build && npm run nobuild:unit",
+    "doc": "npx typedoc"
```
junderw, line 43: Don't need npx, and script key (doc) should be alphabetical order

junderw:
TODO:

- Figure out why deploy-doc is broken.
- Add more comments to various items in the library so that the docs are more fleshed out.

jasonandjay:
> TODO:
> 
> * Figure out why deploy-doc is broken.
> * Add more comments to various items in the library so that the docs are more fleshed out.

I guess it may caused by get pages token failed.
You can enable github pages first and select source from github actions, then you can re-run failed job.

junderw:
Still not working.

This is the yaml file it recommended me when I tried to configure a static HTML setup through Github Actions.

Perhaps something is missing from your diff? Could you look into it for me?

```yaml
# Simple workflow for deploying static content to GitHub Pages
name: Deploy static content to Pages

on:
  # Runs on pushes targeting the default branch
  push:
    branches: ["master"]

  # Allows you to run this workflow manually from the Actions tab
  workflow_dispatch:

# Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
permissions:
  contents: read
  pages: write
  id-token: write

# Allow only one concurrent deployment, skipping runs queued between the run in-progress and latest queued.
# However, do NOT cancel in-progress runs as we want to allow these production deployments to complete.
concurrency:
  group: "pages"
  cancel-in-progress: false

jobs:
  # Single deploy job since we're just deploying
  deploy:
    environment:
      name: github-pages
      url: ${{ steps.deployment.outputs.page_url }}
    runs-on: ubuntu-latest
    steps:
      - name: Checkout
        uses: actions/checkout@v3
      - name: Setup Pages
        uses: actions/configure-pages@v3
      - name: Upload artifact
        uses: actions/upload-pages-artifact@v2
        with:
          # Upload entire repository
          path: '.'
      - name: Deploy to GitHub Pages
        id: deployment
        uses: actions/deploy-pages@v2
```

junderw:
Looks like `actions/configure-pages` is missing?

jasonandjay:
> Looks like `actions/configure-pages` is missing?

logger report [Branch "doc" is not allowed to deploy to github-pages due to environment protection rules.]
To allow the "doc" branch to deploy to GitHub Pages, you'll need to adjust the branch protection rules. Follow these steps:

Go to your GitHub Repository:
Open the repository in your browser.

Click on "Settings":
It's usually located near the right end of the menu bar, next to "Actions" and "Insights".

Navigate to "Branches":
In the left sidebar, click on the "Branches" tab.

Select "Add rule":
You should see an option to add a new branch protection rule. Click on it.

Enter "doc" as the branch name:
In the "Branch name pattern" field, type "doc" to specify the branch you want to adjust the rules for.

Configure Protection Rules:
Here, you'll have several options. You'll want to pay attention to the following:

Require pull request reviews before merging: If you have this enabled, make sure it allows merging without approval. You might need to disable it temporarily for the "doc" branch.

Require status checks to pass before merging: Ensure that the necessary checks (e.g., GitHub Pages deployment) are not required for the "doc" branch.

Include administrators: Make sure this option is checked if you want administrators to be exempt from these rules.

Save Changes:
After making the necessary adjustments, click on the "Save changes" button.

junderw:
Ok, well, I can verify the results of the doc command locally. Perhaps we should make the doc jobs only run on master.

Having it run on pull requests will just pollute the CI with failed tests.

jasonandjay:
Deploy doc task has updated, it will only take effect on master branch.

junderw:
I wonder about versioning.

If it would be possible to have it recognize whether it's being deployed on master or a tag and change the URL path?

Then maybe people could navigate to older versions by clicking a link to the docs generated for a specific tag.

I would hope that the default page is the last tagged version or something.

If that's not possible, I guess it's fine.

I'd also like to flesh out some of the documentation in this PR so that the first release of the docs isn't so bare bones.

If you can help by documenting any of the smaller stuff I'd appreciate it. The more the better.

jasonandjay:
I think add version control is possible and necessary but with low priority.
And  flesh out documentation is high priority.
I want to add descriptions and example to class、function and namespace. 
example can be copied from tests.

such as address namespace, do you think any other content or content style need to be added?
demo

jasonandjay:
I think add version control is possible and necessary but with low priority.
And flesh out documentation is how priority.
I want to add description and examples to function、class、interface and namespace.
examples can be copied from tests, so users can easy to use and it`s better for us to maintain.

such as fromBase58Check, do you think any other content or content style do we need to add?
demo

jasonandjay:
For properties, we can add description like this, do you think its ok?
proper

junderw:
> I want to add description and examples to function、class、interface and namespace. examples can be copied from tests, so users can easy to use and it`s better for us to maintain.
> 
> such as fromBase58Check, do you think any other content or content style do we need to add?

This looks fine. I am a little worried about tests and their copy-pasted doc-equivalents going stale... but I think as long as I remember to mention fixing the doc strings in review it should be fine.

> I think add version control is possible and necessary but with low priority.

Yeah, this is a "nice to have"

You are awesome, thanks for your help!

package.json
```diff
@@ -54,6 +54,7 @@
     "bech32": "^2.0.0",
     "bip174": "^2.1.1",
     "bs58check": "^3.0.1",
+    "typedoc-plugin-bitcoinjs-runcase": "^1.0.1",
```
junderw, line 57: Shouldn't this be a devDependency?

jasonandjay: Yes, it should me. I remember this **but maybe something is wrong!!**

I want to tell you about another thing. 
**I added an executable environment to the document to help us more flexibly display related cases and provide user testing**. 
The background is as follows

1. We have a lot of test cases. Moving too many cases will affect the organizational structure of our code, and the workload will be huge.
2. For related cases, single testing can be combined with the context to understand more, more comprehensively and more efficiently.

To achieve this goal, I did the following things, which also cost me a lot of time:

1. Build bitcoinjs-lib browser version
2. Develop typedoc-plugin-bitcoinjs-runcase plug-in to support sneaking into the executable environment
3. Modify relevant comment tags

You can view the relevant effects here: https://jasonandjay.github.io/bitcoinjs-lib/functions/address.fromBase58Check.html




I think this is a great feature, but I need your help evaluating whether it's suitable

jasonandjay: That way we can tell the user to test here and find more cases in the single test

Then we only need to **give a correct usage method in example** (I customized a @case tag), so that the construction efficiency of our document will be much higher

jasonandjay: @junderw please take a look

pajasevi: @jasonandjay  `typedoc-plugin-bitcoinjs-runcase` is a production dependency and has `react` and `react-dom` as dependencies. That doesn't belong in bitcoinjs at all.

jasonandjay: @pajasevi yes, i will change it later, i want to know is this the right direction to add an executable environment to the document to simplify the case in the single test? so as to proceed with the next step.

junderw: This will take a while to review.

junderw: Perhaps this should be in a follow up pr

jasonandjay: Review typedoc plugin or comments in our repository?

junderw: Adding `typedoc-plugin-bitcoinjs-runcase` should be in a separate PR is what I'm saying.

jasonandjay:
Got it
typedoc-plugin-bitcoinjs-runcase is a devDependencies, so it will not intrude into our program and help provide a run code environment when building documents.

It will now be more efficient for us to improve the documentation

Or, let’s add the key comments in the document first, and then add the run case. What do you think?

Muniru0 (approved):
Amazing work.

Muniru0 (approved):

jasonandjay:
I did two things
1. ⏪ rollback: separate `typedoc-plugin-bitcoinjs-runcase` to another PR
2. Add key comments to the code to make the document look more complete.

junderw:
Thanks. That lowers the scope of the review significantly. I'll take another look later today.

jasonandjay:
Is there anything that needs to be optimized? Maybe we can push document construction to next step.

junderw (approved):
LGTM. Great job. Github is being weird. Crossing finger that this works lol.

junderw:
Deployed successfully!

https://bitcoinjs.github.io/bitcoinjs-lib/

Maybe add this to the README?

PR #1991, opened 2023-10-12
remove redundant code

jasonandjay:
1. extract stacksEqual
2. decode P2TR address with fromBech32
3. convergence hashType judgment to isDefinedHashType

I think its necessary to remove some redundant code while adding document comment to avoid repeat comment


junderw (approved):
LGTM

Issue #1966, opened 2023-08-06
Better types for opcodes


vostrnad:
ops.ts exports the list of opcodes as:
```ts
const OPS: { [key: string]: number }
```
That's not a very safe type, as it permits access with any string key. It also doesn't work with code completion.

Can this type be made more specific? It would technically be a breaking change for anyone using this in an unsafe manner, but I believe it's worth it. I'm happy to work on this.

junderw:
OPS is pretty straightforward, (just as `as const` at the end) but REVERSE_OPS might be tricky to get to only respond to specific number keys...

Admittedly, when I migrated this library to TypeScript I was a TS beginner, and only feel like I now have a semi-intermediate grasp on it...

PRs are welcome, but I think this won't land until the next major version whenever that is.

ChrisCho-H:
How about using enum as bitcoin core does?
https://github.com/bitcoin/bitcoin/blob/master/src/script/script.h#L72

ChrisCho-H:
Made PR for this https://github.com/bitcoinjs/bitcoinjs-lib/pull/2028.
Any review or commit would be grateful!

PR #2028, opened 2024-01-05
feature: opcode enum type implemented in a compatible way with current version

ChrisCho-H:
- https://github.com/bitcoinjs/bitcoinjs-lib/issues/1966 mentions that current opcode is type-unsafe. Type safe version is considered as tricky and breaking change as it might function in a bit different way from it does so far(in any versions) and have reverse opcode, too. 
- However, as TS numeric Enum compiles with reverse mapping(https://www.typescriptlang.org/docs/handbook/enums.html#reverse-mappings), we can just change OPS to enum, and export OPS and OPS as REVERSE_OPS for reverse opcode.
- IMO It could be good change as it provides type-safe ops and reverse_ops, without changing the way to use this module(no breaking change)
- Also, I deleted OP_PUBKEY and OP_PUBKEYHASH, which are not use in bitcoin anymore


ts_src/script.ts
```diff
@@ -173,7 +173,8 @@ export function fromASM(asm: string): Buffer {
   return compile(
     asm.split(' ').map(chunkStr => {
       // opcode?
-      if (OPS[chunkStr] !== undefined) return OPS[chunkStr];
+      if (OPS[chunkStr as keyof typeof OPS] !== undefined)
```
junderw, line 176: I think this kind of proves the "compatible way" part not being 100% accurate.

If we look at the JS, then yeah. But TypeScript people will have breakage.

junderw: This is a great PR though.

I'm thinking of how to push this through.

junderw: Not sure how many people are writing script parsers from scratch just pulling in the OPS though...

The use case of "get opcode from arbitrary string" is probably very niche. But still it's a possible breakage.

ts_src/ops.ts
```diff
@@ -1,143 +1,135 @@
...
-  OP_NOP6: 181,
-  OP_NOP7: 182,
-  OP_NOP8: 183,
-  OP_NOP9: 184,
-  OP_NOP10: 185,
-
-  OP_CHECKSIGADD: 186,
-
-  OP_PUBKEYHASH: 253,
-  OP_PUBKEY: 254,
```
junderw, line 133: These two are missing. This is not compatible.

junderw (approved):
I'll approve and set it for the next major version milestone due to breakages.

jasonandjay:
Maintainer triage: this PR is pending re-validation against current `master` / v7 direction.

Please:
1) rebase (or merge) latest `master`
2) ensure full CI passes (`format`, `lint`, `lint-tests`, `gitdiff`, `unit`, `integration`, `coverage`, `build-doc`)
3) confirm this PR still matches current roadmap/design assumptions

If there is no update in ~14 days, we may close this PR for backlog hygiene. You can always reopen or submit a refreshed PR later.

PR #2068, opened 2024-04-04
Only warn once for forward segwit warnings

### LINKED ISSUE #1800: Reduce warnings about future segwit version when parsing output scripts
https://github.com/bitcoinjs/bitcoinjs-lib/blob/93af5afe67dbc491e36bdfc8d48a00179093f7d1/ts_src/address.ts#L52

I'm writing an app that continuously parses lots of transactions (for analysis purposes) This warning is printed every time such output-script is parsed, and it really spams my logs.

Please make it possible to avoid this warning, maybe via an environment variable.
Another reasonable option is to print the warning only after the first time such an output-script is found, and not in following occurrences.

Thanks!

junderw:
Fixes #1800 

This will lower the noise especially for long running parsing processes.


junderw:
This is a simple fix.

Issue #2030, opened 2024-01-06
Switch from typeforce to something else


junderw:
General musings:

- Typia - https://typia.io/docs/
  - I really like the API and the ease of use. The speed is also very good. However, it's relatively young and is only being maintained by one person, so longevity might be a concern.
- zod - https://zod.dev/
  - This seems like the one that everyone likes. It seems a bit complex, and it's slow. It is also essentially maintained by one person, but they have almost 100 Github sponsors, which at least shows the willingness for people to pay to support the project.

I am also open to other suggestions.

Replacing typeforce in the bitcoinjs ecosystem will be breaking changes most likely, since a lot of these require changes to the typescript types.

pajasevi:
I did some research and from what I observed, there isn't a widely-used validation library that has more than one maintainer. Some people might suggest AJV but that doesn't really fit the use-case of bitcoinjs as it's JSON schema based. Other libraries that have good performance either aren't widely used, rely on JSON schema or have complicated API.

Regarding the two libraries already mentioned: Zod is certainly well established in the space and is widely used but the speed might be a concern. Typia on the other hand is super fast due being strictly Ahead Of Time compiled and looks healthy right now in terms of development. Also the API is very intuitive.

One thing to look at is also library dependencies: Typia has 4 dependencies while Zod has zero. 

So while I think that Typia might be a better fit due to API, performance and ease-of-use, I would be cautious.

motorina0:
What are the current/future limitations of `typeforce`?
Why is the change needed?

junderw:
@pajasevi mentioned a few things in the PR that mentions this issue...

Perhaps @pajasevi could go into further details.

pajasevi:
@motorina0 I see several issues with typeforce.
- It is unmantained - last change 2 years ago
- no ES module support - minor issue in this case
- very bad extendability - current set of type checks is very limited and the API is very clunky when you want to add a sophisticated extension, like I tried with Uint8Array of length 32
- no Typescript support - personally my biggest issue with the library is that it doesn't have any type definitions for its API (not aven a third-party @types package) and results of type assertions therefore aren't typed at all

This could be easily solved even by having a standard set of type guard functions that could be reused in the bitcoinjs ecosystem. Using a different third-party library that supports all the features might be easier of course.

samholmes:
May I suggest [Cleaners](https://cleaners.js.org)?

pajasevi:
[...]

PR #2072, opened 2024-04-08
Remove methods that neither used nor export

jasonandjay:
1.  For Signer & Network, we have defined and exported it from psbt module.
2.  For UInt31, BIP32Path and ECPoint, we have neither used nor exported them. The check of BIP32Path should be related to BIP32, and ECPoint should be related to ECPair. This project is not suitable for checking.

So I think it's time to remove this methods.

In addition, I suggest that we stop using typeforce to verify input parameters and use ts types instead for better maintenance.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ test/*.js
 test/integration/*.js
 !test/ts-node-register.js
 docs
+plugin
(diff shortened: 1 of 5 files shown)


junderw:
> use ts types instead

TS types don't check at runtime. Some people still use plain old JavaScript, and some people also use `@ts-ignore` etc. against all best practices because "it just works".

If anything, there are probably some functions that SHOULD have runtime type checks that don't, and we should add more.

That said, I have #2030 which I think is necessary.

junderw (approved):

PR #2090, opened 2024-04-29
Allow PSBT to use `bip174`'s getTransaction method

mikedotexe:
Small change that allows this repository's `Psbt` class to access a method present in the `PsbtBase`, which comes from the `bip174` package. Currently, this method is not accessible:

https://github.com/bitcoinjs/bip174/blob/25fd6842b76984b169d8f0be2adcb3db322fe98c/ts_src/lib/psbt.ts#L188-L190

---

## Context

I am working with a multi-party computation system where I'd like to build a PSBT with intermediate steps. Among those steps is asking the MPC to sign each input UTXO.

In order to get the sighash for each input, it seems the `Transaction` object is needed, since it contains methods like `hashForSignature`.

My original approach was to build a PSBT incrementally, but I could not determine how to extract the Transaction object from the PSBT. My attempts seemed to always necessitate "finalizing inputs" which I cannot do until I've gathered the signatures from the MPCs for each input.

If I'm tracking this problem correctly, I believe I'd currently have to create independent Transaction objects in this library for each input UTXO, get the signatures, and then be able to create the PSBT, extract the finalized transaction, and broadcast. With the small change in this PR, the PSBT can exist from the start.

### DIFF
--- a/src/psbt.d.ts
+++ b/src/psbt.d.ts
@@ -119,6 +119,7 @@ export declare class Psbt {
     updateGlobal(updateData: PsbtGlobalUpdate): this;
     updateInput(inputIndex: number, updateData: PsbtInputUpdate): this;
     updateOutput(outputIndex: number, updateData: PsbtOutputUpdate): this;
+    getTransaction(): Buffer;
     addUnknownKeyValToGlobal(keyVal: KeyValue): this;
     addUnknownKeyValToInput(inputIndex: number, keyVal: KeyValue): this;
     addUnknownKeyValToOutput(outputIndex: number, keyVal: KeyValue): this;
--- a/src/psbt.js
+++ b/src/psbt.js
@@ -901,6 +901,9 @@ class Psbt {
     this.data.updateOutput(outputIndex, updateData);
     return this;
   }
+  getTransaction() {
+    return this.data.getTransaction();
+  }
   addUnknownKeyValToGlobal(keyVal) {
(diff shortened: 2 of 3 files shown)


junderw:
This is not the answer to your problem.

The whole point of bitcoinjs-lib Psbt is to abstract away sighashes and ECDSA operations.

Adding a method that which exposes the internals goes against what it is for.

There are two methods you can use:

1. Create the sighashes yourself. This is actually easier than you'd think, but it requires looking at the related BIPs. You can also take a peek at the corresponding hashForSignature methods on the Transaction class to help you.
2. Utilize the async signer interface. The only time the hash for signing is exposed during Psbt use is inside the sign() method of the signer interface you pass into the signing methods on Psbt. You could create a dummy signer that will grab the hash from the sign() method, and throw (you could even throw the hash, JS lets you do that) or something to get the hash out of the Psbt. Or if the MPC parties are always-on devices, you could have the sign method wait async-ly for each response to calculate the final signature. (This would be less feasible if the MPC parties require human intervention and could take minutes/hours)

If you want to create a dummy signer, the pubkey must return the final pubkey (we use the pubkey of the signer interface to search for inputs to sign by searching for the pubkey).

If you have any further questions, you can ask me here and I'd be glad to help.

If you can narrow down scope: ie. "We will only sign Taproot inputs" etc. I can give you more precise advice, but as it stands Bitcoin has a lot of signing hash algorithms.

Issue #2096, opened 2024-05-09
toASM throws TypeError: Cannot read properties of null (reading 'map')


O2rki:
Happens when `decompile` returns a null object, then `toASM` tries to `map`. 
https://github.com/bitcoinjs/bitcoinjs-lib/blob/6396fb488df79dd7cb27eb1b5f9fb56318055ff5/src/script.js#L158-L162

Do I have to check myself (in my code) if `decompile` returns a null before calling `toASM`?.

jasonandjay:
We can optimize decomplie return value to avoid type error.

But you should check your logic why get incorrect chunks to exec toASM?

O2rki:
I was debugging the lib. Incorrect chunks may get there for whatever reason.

jasonandjay:
OK, we will fix it.

O2rki:
Thanks @jasonandjay

PR #2097, opened 2024-05-10
 perfs: optimize script.decomplie return type

jasonandjay:
Optimize script.decomplie return type to Stack 
1. avoid type error. for example: null.map https://github.com/bitcoinjs/bitcoinjs-lib/issues/2096
2. avoid unnecessary type assertion
3. simplify the judgment of returned results

### DIFF
--- a/src/script.js
+++ b/src/script.js
@@ -159,7 +159,7 @@ function toASM(chunks) {
     chunks = decompile(chunks);
   }
   if (!chunks) {
-    throw new Error('convert invalid chunks to ASM');
+    throw new Error('Could not convert invalid chunks to ASM');
   }
   return chunks
     .map(chunk => {
--- a/src/script.js
+++ b/src/script.js
@@ -158,6 +158,9 @@ function toASM(chunks) {
   if (chunksIsBuffer(chunks)) {
     chunks = decompile(chunks);
   }
+  if (!chunks) {
+    throw new Error('convert invalid chunks to ASM');
+  }
   return chunks
     .map(chunk => {
       // data?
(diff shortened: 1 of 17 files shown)


junderw:
There is a semantic difference between `[]` and `null`

`null` indicates "the decompilation failed / there is no valid decompilation that exists for this Buffer" without need for throwing an exception.

`[]` means "we decompiled it and it was an empty script."

By collapsing this, `[]` can now mean two things.

This is a breaking change.

I agree that the bug in the other issue should be fixed, but this is not the way we should fix it.

Thanks for the PR though.

jasonandjay:
> There is a semantic difference between `[]` and `null`
> 
> `null` indicates "the decompilation failed / there is no valid decompilation that exists for this Buffer" without need for throwing an exception.
> 
> `[]` means "we decompiled it and it was an empty script."
> 
> By collapsing this, `[]` can now mean two things.
> 
> This is a breaking change.
> 
> I agree that the bug in the other issue should be fixed, but this is not the way we should fix it.
> 
> Thanks for the PR though.

I agree

How about throw error replace return null ?

junderw:
I'd rather not have to thing about the knock on effects.

A simple bug fix is easier to approve.

jasonandjay:
@junderw how about this?
just throw an error from toASM when decompile failed
Minimal impact on existing logic

src/script.js
```diff
@@ -158,6 +158,9 @@ function toASM(chunks) {
   if (chunksIsBuffer(chunks)) {
     chunks = decompile(chunks);
   }
+  if (!chunks) {
+    throw new Error('convert invalid chunks to ASM');
```
junderw, line 162: This should be `'Could not convert invalid chunks to ASM'` instead.

jasonandjay:
Please help reRun actions
it`s failed for netowrk issue, reRun failed works will repair

Issue #2106, opened 2024-06-12
Can't use tiny-secp256k1 (required by bitcoinjs-lib for ECC) in browser: fs_1.readFileSync is not a function


SuperHenkie:
I'm having trouble getting BitcoinJS to work in a browser. Very unfamiliar with node/npm so maybe I'm missing something, here are my exact steps to reproduce the issue. In a new directory I do this:

`npm init -y`
`npm install bitcoinjs-lib tiny-secp256k1 browserify`

I create a file `main.js` containing:

`const bitcoin = require('bitcoinjs-lib');`
`const ecc = require('tiny-secp256k1');`
`window.bitcoin = bitcoin;`

Then I do `browserify main.js -o bundle.js` to create a browserified package.

I create a file `test.html` containing:

`<!DOCTYPE html><html><head><meta charset="utf-8">`
`<script src="bundle.js"></script>`
`</head><body></body></html>`

When I open this test.html in the browser (latest Firefox or Chrome or Brave) the console shows an error:

> Uncaught TypeError: (0 , fs_1.readFileSync) is not a function

Note that this error does NOT occur when I remove or comment the `const ecc = ...` line in main.js. However I need tiny-secp256k1 for functionality like ECPair etc.

Am I doing something wrong or is BitcoinJS not supposed to be used in the browser with tiny-secp256k1 features?

VictorZhang2014:
The first option, you need to use `bundle.js` in an environment, like `http://localhost:3000` or `https://google.com/web3`, but do not use the file like `index.html` opened in the browser directly, it won't work in this way.

The second option, find the error position like`(0 , fs_1.readFileSync)` and annotate the code, and replace it by loading the wasm file through `fetch`.

SuperHenkie:
Thanks, how do you mean I can't open `test.html` in the browser directly? Note that I'm running this in an offline environment, just on my local computer (that doesn't run a webserver). It always worked like this with previous BitcoinJS versions (v5 and earlier) but I think since v6 there have been some changes resulting in this browser incompatibility. 

Note that it works fine if I comment the `const ecc` line, it's just this particular `tiny-secp256k1` that seems to cause an issue when using in a browserified package (this was not the case with previous versions).

[...]

PR #2109, opened 2024-06-13
Update readme 

jasonandjay:
**In case of https://github.com/bitcoinjs/bitcoinjs-lib/issues/2106**

**1. Documented Error:**
Added details about the Uncaught TypeError: (0 , fs_1.readFileSync) is not a function error with tiny-secp256k1 due to WASM issues in browsers.

**2. ECC Library Alternatives:**
Introduced @bitcoinjs-lib/tiny-secp256k1-asmjs and @bitcoinerlab/secp256k1 as alternatives for improved browser support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,14 +103,15 @@ Which you can then import as an ESM module:
 ````
 
 **Using Taproot:**  
-When utilizing Taproot features with bitcoinjs-lib, you may need to include an additional ECC (Elliptic Curve Cryptography) library. The commonly used tiny-secp256k1 library, however, might lead to compatibility issues due to its reliance on WASM (WebAssembly).  
+When utilizing Taproot features with bitcoinjs-lib, you may need to include an additional ECC (Elliptic Curve Cryptography) library. The commonly used `tiny-secp256k1` library, however, might lead to compatibility issues due to its reliance on WASM (WebAssembly).  
 
 If you encounter the following error:
 
 ```
 Uncaught TypeError: (0 , fs_1.readFileSync) is not a function
 ```
-This indicates that tiny-secp256k1's WASM implementation is not fully compatible with your browser environment.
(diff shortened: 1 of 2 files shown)


jasonandjay:
unexpected test failed of audit, please have a look @junderw

junderw:
Add ` -x 1097496` to the audit script to ignore. This vulnerability will at worst crash when running the tests or eslint.

junderw:
1. Install a clean node_modules from `npm ci` (from package-lock)
2. Run `npm ls braces`
3. It shows which dependencies use it.
4. Upgrade all those dependencies to latest.
5. Refresh node_modules by deleting and re-runing npm ci
6. Check again

Since it did not fix the audit, that means there is no fix out yet and/or the chain of dependencies hasn't upgraded yet.

We can't do anything, but check the content of the vulnerability and see if we can ignore it.

I think we can ignore it.

junderw:
I think instead of modifying the audit script in package.json we should just modify the invocation in CI.

If some other user wants to run audit, we shouldn't mess with it.

... maybe `-l high` should also be moved from package.json to main_ci?

README.md
```diff
@@ -102,6 +102,22 @@ Which you can then import as an ESM module:
 <script type="module">import "/scripts/bitcoinjs-lib.js"</script>
 ````
 
+**Using Taproot:**  
+When utilizing Taproot features with bitcoinjs-lib, you may need to include an additional ECC (Elliptic Curve Cryptography) library. The commonly used tiny-secp256k1 library, however, might lead to compatibility issues due to its reliance on WASM (WebAssembly).
```
junderw, line 106: Backticks around tiny-secp256k1 so it's monospaced and stands out.

jasonandjay:
Which part needs to be optimized?

junderw (approved):

PR #2110, opened 2024-06-14
fix: added error with clear message when signing a tx without outputs

gomezdn:
There's this line of code ```this.outs.map(output => 8 + varSliceSize(output.script)).reduce((a, b) => a + b)``` that, when the unsigned transaction has no outputs set, ```this.outs``` is an empty array and the only error it throws says "Reduce of empty array with no initial value", which makes one dig into the bitcoinjs code and understand what it was trying to do. Now the error message helps to understand what is happening.

### DIFF
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -318,6 +318,8 @@ class Transaction {
       hashSequences = bcrypto.sha256(bufferWriter.end());
     }
     if (!(isNone || isSingle)) {
+      if (!this.outs.length)
+        throw new Error('Add outputs to the transaction before signing.');
       const txOutsSize = this.outs
         .map(output => 8 + varSliceSize(output.script))
         .reduce((a, b) => a + b);
--- a/ts_src/transaction.ts
+++ b/ts_src/transaction.ts
@@ -414,6 +414,8 @@ export class Transaction {
     }
 
     if (!(isNone || isSingle)) {
+      if (!this.outs.length)
+        throw new Error('Add outputs to the transaction before signing.');
       const txOutsSize = this.outs
         .map(output => 8 + varSliceSize(output.script))
         .reduce((a, b) => a + b);


junderw:
Perhaps adding a default value of 0 is a better solution.

IOVgomezdn:
> Perhaps adding a default value of 0 is a better solution.

Yes that was my first solution, but then I thought, why one would sign a transaction with no outputs? It's probably a mistake, and an error would prevent the user to broadcast it.

gomezdn:
also, if somebody can give me a hint about the ci test that failed it would be appreciated

junderw:
> also, if somebody can give me a hint about the ci test that failed it would be appreciated

See #2109

gomezdn:
> > also, if somebody can give me a hint about the ci test that failed it would be appreciated
> 
> See #2109

so? what do you suggest me to do? Also what do you think about the fix? adding a default value is better or the error?

junderw:
> so?

This is rude. Since you seem to be based in a country where English is not your native language, I take no offence from it. But for future reference, "so?" as a response to anything is pretty rude.

> what do you suggest me to do?

1. Read the PR I linked.
2. Notice the discussion about the same issue (audit CI fails)
3. Notice the commit that was added by the PR author after the discussion (https://github.com/bitcoinjs/bitcoinjs-lib/pull/2109/commits/1a868e8ce4e2302a465b53bf47cc68bae55b113a)

Or just wait until the other PR is merged and then rebase on master. Either one is fine.

> Also what do you think about the fix? adding a default value is better or the error?

Either one is fine. I just wanted to clarify the reasoning behind why you chose one solution over the other.

junderw:
I merged https://github.com/bitcoinjs/bitcoinjs-lib/pull/2113 to master. Please rebase and the audit will pass.

junderw (approved):

gomezdn:
> > so?
> 
> This is rude. Since you seem to be based in a country where English is not your native language, I take no offence from it. But for future reference, "so?" as a response to anything is pretty rude.
> 
> > what do you suggest me to do?
> 
> 1. Read the PR I linked.
> 2. Notice the discussion about the same issue (audit CI fails)
> 3. Notice the commit that was added by the PR author after the discussion ([1a868e8](https://github.com/bitcoinjs/bitcoinjs-lib/commit/1a868e8ce4e2302a465b53bf47cc68bae55b113a))
> 
> Or just wait until the other PR is merged and then rebase on master. Either one is fine.
> 
> > Also what do you think about the fix? adding a default value is better or the error?
> 
> Either one is fine. I just wanted to clarify the reasoning behind why you chose one solution over the other.

so sorry! But no means I wanted to be rude. Thanks for clarifying it. Yes I'm not an English native speaker, will have it in mind from now on.

I asked because I didn't want to interfere beyond the purpose of my fix. Thanks!

Issue #2132, opened 2024-07-19
signAllInputsHD not working for taproot key


cpuchainorg:
The function doesn't work with tapBip32Derivation params

https://github.com/cpuchain/btc-signer/blob/main/src/wallet.ts#L754C18-L754C33

Would like to know if it could be solved without iterating for the inputs

junderw:
No it is currently not supported.

junderw:
@jasonandjay want to give a crack at it?

jasonandjay:
> @jasonandjay want to give a crack at it?

of course

jasonandjay:
Maintainer triage: we are using this issue as the main thread for taproot HD signing support. Related report #2000 is being closed as duplicate to keep scope unified.

paulee:
@jasonandjay @junderw Any chance to review my PR on this issue?

PR #2137, opened 2024-07-24
feat: support taproot for signAllInputsHD

jasonandjay:
- issue discussion: https://github.com/bitcoinjs/bitcoinjs-lib/issues/2132
- feat
   - affected Signature Methods 
      - signAllInputsHD
      - signAllInputsAsyncHD
      - signInputHD
      - signInputAsyncHD
   - remove conflict default sighashTypes of signInputHD
      - Transaction.SIGHASH_DEFAULT for _signTaprootInput
      - Transaction.SIGHASH_ALL for _signInput
      - so we should remove the outermost default signType
   - add integration test for HDWallet with tapBip32Derivation
      - we will need more integration tests later.

### DIFF
--- a/ts_src/psbt.ts
+++ b/ts_src/psbt.ts
@@ -19,7 +19,11 @@ import { fromOutputScript, toOutputScript } from './address';
 import { cloneBuffer, reverseBuffer } from './bufferutils';
 import { bitcoin as btcNetwork, Network } from './networks';
 import * as payments from './payments';
-import { tapleafHash } from './payments/bip341';
+import {
+  calculateScriptTreeMerkleRoot,
+  tapleafHash,
+  tapTweakHash,
+} from './payments/bip341';
 import * as bscript from './script';
 import { Output, Transaction } from './transaction';
 import {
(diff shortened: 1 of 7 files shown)


jasonandjay:
When we use key-path to unlock UTXO, we don't need leafHashes, so I submitted another MR to solve this problem https://github.com/bitcoinjs/bip174/pull/38

ts_src/psbt.ts
```diff
@@ -1189,6 +1190,14 @@ export interface HDSigner extends HDSignerBase {
...
    */
   sign(hash: Buffer): Buffer;
+  /**
+   * Adjusts a keypair for Taproot payments by applying a tweak to derive the internal key.
+   *
+   * In Taproot, a keypair may need to be tweaked to produce an internal key that conforms to the Taproot script.
+   * This tweak process involves modifying the original keypair based on a specific tweak value to ensure compatibility
+   * with the Taproot address format and functionality.
+   */
+  tweak(t: Buffer): Signer;
```
junderw, line 1200: This is a breaking change. We now require this of all users of signHD.

jasonandjay:
How about code review progress?

Salarja (approved):

cpuchainorg:
@junderw Could you check this PR?

landabaso:
I have an observation regarding the asymmetry in `signInput` when using `ecpair`.  

I'm not saying this is incorrect, but this PR allows `psbt.signInputHD(index, node)`, which detects if the input is a Taproot input and internally tweaks the signer (`getTweakSignersFromHD`).  

On the other hand, in `psbt.signInput(index, ecpair)`, the user must pre-tweak the signer manually.  

As a result, it's not possible to use `signAllInputs` with a common `ecpair` for both Taproot and non-Taproot inputs, whereas `signAllInputsHD` allows using a common node for both.  

Again, I'm not saying this is necessarily wrong. Perhaps this is the only viable approach, but I wanted to highlight this to see if there’s anything further we can do, or at least reflect this behavior more clearly in the comments at the top of `psbt.ts`.

bittoby:
@jasonandjay Are you still working on this? May I complete this feature if it is needed? I can do it perfectly. thanks

jasonandjay:
> @jasonandjay你还在努力吗？如果需要，我可以完成这个功能吗？我能完美地做到。谢谢

PR Welcome

PR #2160, opened 2024-09-04
Converge payment related methods

SpiderMan670:
1. converge payment related methods
2. update the reference of payment method judgment


jasonandjay (approved):

junderw:
Thanks for the pull request, but this is not the direction we want to take.

PR #2161, opened 2024-09-04
Add CJS and ESM support

Nesopie:
- Add support for CJS and ESM
- Moved from typeforce (does not have esm support) to Valibot 
- Replaced Buffer usage to Uint8Array
- There are 44 less test cases now as all logic with regards to writeUint64 and related functions has been moved to `uint8array-tools`. 
- All amount / value fields now use bigints


junderw (approved):
Legend!

Let's publish rc1 versions of all the major dependencies, then fix them in package.json here, then push rc1 version for bitcoinjs-lib.

landabaso:
Great work on this HUGE refactor, @Nesopie! 👏

I noticed two related changes:

* [`9c3e8cc`](https://github.com/bitcoinjs/bitcoinjs-lib/commit/9c3e8cc3635dc6f5d8b5c632eb2d250b112c8fb7#diff-e5b89f81c57c3fa73e9159ff39f27ce1be14650a1f464dbec1e76bf520eca75c%C3%A7) – drops the named exports **`ripemd160`**, **`sha1`**, and **`sha256`**.
* [`847090c`](https://github.com/bitcoinjs/bitcoinjs-lib/commit/847090cbb2e42206481f48f8c29d4813e6a029e8#diff-62a0b072e334225376480973111fbfe81389fcc2a66d1386e8d2c722c0189430L8) – removes the unit tests that covered those helpers.

Because the PR is so large, I'm not sure whether that removal was **intentional** or an unintended side-effect.

A few of my own projects (and several others I found doing a quick internet search) use `crypto.sha256`, for example. Restoring the helpers would simplify the upgrade path to v7.

*If the removal was accidental,* I'm happy to open a follow-up PR that re-exports wrappers around Noble's hashes and brings the tests back.
*If it was deliberate,* it would be helpful to note it in the CHANGELOG.

Thanks again for all the work that went into this release!

junderw:
Still haven't released 7.0.0 yet, so we can definitely get that in before 7.0.0 release if you want.

landabaso:
> Still haven't released 7.0.0 yet, so we can definitely get that in before 7.0.0 release if you want.

Great. I'll take it from here.

BTW, `npm install bitcoinjs-lib` now pulls in [6.1.7](https://www.npmjs.com/package/bitcoinjs-lib?activeTab=versions).
`7.0.0-rc0` [used to be the default](https://github.com/bitcoinjs/bitcoinjs-lib/issues/2176), but I think that changed once the v6 patch from [#2184](https://github.com/bitcoinjs/bitcoinjs-lib/pull/2184) was later published to npm.

Once the helpers and tests are restored, if we still want the RC to be the `latest`, we can push a fresh RC and retag **latest** back to v7.

PR #2170, opened 2024-09-30
Add stripWitness to Transaction Issue 2169

JacksonDMiller:
This PR is intended to solve [issue 2169](https://github.com/bitcoinjs/bitcoinjs-lib/issues/2169)

- Adds a method to the transaction object to remove witness data. 
  - This is needed to construct a coinbase transaction with properly formatted inputs.
- Adds a test to cover the new method.

      
Credit to @benjamin-wilson for pointing it out in [issue 1951](https://github.com/bitcoinjs/bitcoinjs-lib/issues/1951.)

Issue #2169, opened 2024-09-25
Add stripWitnesses to Transaction


junderw:
Replaces #1951

Iterate through ins and replace witness with EMPTY_WITNESS.

@JacksonDMiller wanna take a crack at it?

junderw:
I made a `v6` branch for backporting.

Please make the PR towards v6 for now and I'll release it as a minor version bump.

If you want to make a second PR against master as well that would be appreciated.

JacksonDMiller:
@junderw Yeah, I would. I'll start reading up and try to get something in soon.

JacksonDMiller:
@junderw no rush at all, but I made a [pull request](https://github.com/bitcoinjs/bitcoinjs-lib/pull/2170) whenever you have time.

junderw:
I'll leave this issue open until we backport to v6.

(Pull the v6 branch, do essentially the same commit on top, then open a PR against the v6 branch instead of master)

JacksonDMiller:
> I'll leave this issue open until we backport to v6.
> 
> (Pull the v6 branch, do essentially the same commit on top, then open a PR against the v6 branch instead of master)

Oops, thanks for reminding me. Some tests are failing on V6 at the moment, so it was easier for me to work against Master while I'm still trying to learn the codebase. I meant to swing back to v6, but it slipped my mind. 

Here is the [PR](https://github.com/bitcoinjs/bitcoinjs-lib/pull/2171), but I have one question.

JacksonDMiller:
@junderw are we good to close this one out now?

PR #2171, opened 2024-10-01
Update v6 branch: Add stripWitness to Transaction Issue 2169 #2170

JacksonDMiller:
This PR aligns the v6 branch with the recent [changes](https://github.com/bitcoinjs/bitcoinjs-lib/pull/2170) to master. 

This PR is intended to solve [issue 2169](https://github.com/bitcoinjs/bitcoinjs-lib/issues/2169)

- Adds a method to the transaction object to remove witness data. 
  - This is needed to construct a coinbase transaction with properly formatted inputs.
- Adds a test to cover the new method.

### DIFF
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -1,536 +1,482 @@
-'use strict';
-Object.defineProperty(exports, '__esModule', { value: true });
+"use strict";
+Object.defineProperty(exports, "__esModule", { value: true });
 exports.Transaction = void 0;
-const bufferutils_1 = require('./bufferutils');
-const bcrypto = require('./crypto');
-const bscript = require('./script');
-const script_1 = require('./script');
-const types = require('./types');
+const bufferutils_1 = require("./bufferutils");
+const bcrypto = require("./crypto");
+const bscript = require("./script");
+const script_1 = require("./script");
+const types = require("./types");
 const { typeforce } = types;
 function varSliceSize(someScript) {
-  const length = someScript.length;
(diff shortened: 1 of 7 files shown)


src/transaction.js
```diff
@@ -1,536 +1,482 @@
...
-        types.UInt32,
-      ),
-      arguments,
-    );
-    if (
-      values.length !== this.ins.length ||
-      prevOutScripts.length !== this.ins.length
-    ) {
-      throw new Error('Must supply prevout script and value for all inputs');
+    stripWitnesses() {
```
JacksonDMiller, line 130: @junderw I wasn't sure what to do with these changes. Should I just not commit the build changes?

junderw: It might be a version diff.

Wipe node_modules and install deps using `npm ci` and it should be correct. (That's what CI does)

I think master has a newer version of the formatter and maybe your local node_modules still had that newer version (whereas CI will pull the older version).

junderw: Then just run `npm run format && npm run build` and it should give you only the diff in your js similar to the ts, and a new function for the d.ts types file.

junderw:
Also, remove the audit CI job and the better-npm-audit dev dependency. (We got rid of it)

JacksonDMiller:
> Also, remove the audit CI job and the better-npm-audit dev dependency. (We got rid of it)

Done. I'd be happy to make a separate PR for this if you want, but it seems pretty safe.

junderw (approved):

PR #2184, opened 2024-12-02
feat: initEccLib skip verification (v6)

fboucquez:
verifyEcc could be expensive, taking up to 500 ms to finish on some devices. This PR allows the client code to avoid running the verification on boot with a flag. Devs can run the verification on unit tests to verify if the provided ECC is correct. 

This is for V6, we are currently running this version

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 6.1.7
+__added__
+- skip ecc library verification via DANGER_DO_NOT_VERIFY_ECCLIB flag
+
 # 6.1.6
 __fixed__
 - Fix sighash treatment when signing taproot script sign scripts using Psbt (#2104)
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "bitcoinjs-lib",
-  "version": "6.1.6",
+  "version": "6.1.7",
   "lockfileVersion": 2,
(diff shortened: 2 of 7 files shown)


junderw:
I would like to see it be a bit more self-documenting.

Someone could add a single line "true" diff and most people won't blink an eye.

Instead of a boolean, maybe the type should be `skipVerification?: { DANGER_DO_NOT_VERIFY_ECCLIB: boolean; }`

fboucquez:
Hi @junderw , what do you think about this one?

fboucquez:
Thanks @junderw , what do you think now?

samholmes:
How about `opt?: { DANGER_DO_NOT_VERIFY_ECCLIB: boolean; }` (naming it `opt` for options and that pattern is flexible for these sort of function flag anti-patterns.

fboucquez:
> How about `opt?: { DANGER_DO_NOT_VERIFY_ECCLIB: boolean; }` (naming it `opt` for options and that pattern is flexible for these sort of function flag anti-patterns.

Updated

kewde (approved):
utACK

junderw:
Bump the patch version (both package and package-lock) and add something to CHANGELOG.

junderw:
Published.

fboucquez:
Nice! thanks @junderw. I'll update v7

PR #2189, opened 2024-12-14
Optimize OPS type from Record to Enum

decentraldev1:
- Optimize OPS type from Record to Enum
- Standardize the definition and use of keys in OPS
- Dealing with previous issues and PRs
PR：https://github.com/bitcoinjs/bitcoinjs-lib/pull/2028/files
Issue: https://github.com/bitcoinjs/bitcoinjs-lib/issues/1966


decentraldev1:
Going a step further, maybe we can remove REVERSE_OPS directly.

Which a bit redundant

ts_src/script.ts
```diff
@@ -200,14 +200,19 @@ export function toASM(chunks: Uint8Array | Array<number | Uint8Array>): string {
...
 
+  // Compile the ASM string into a Uint8Array
   return compile(
-    asm.split(' ').map(chunkStr => {
-      // opcode?
-      if (OPS[chunkStr] !== undefined) return OPS[chunkStr];
-      v.parse(types.HexSchema, chunkStr);
+    asm.split(' ').map((chunk: string): number | Uint8Array => {
+      // Check if the chunk is an opcode
+      if (chunk in OPS) {
```
junderw, line 207: This is subtly different.

"81 82 83 84 85" should output 5 PUSHDATAs but because of the embedded REVERSE_OPS it will come out as the same as "OP_1 OP_2 OP_3 OP_4 OP_5"

junderw (changes requested):
Please make fix for the logic change. Also add a test to show that hex pushdatas that look like numbers in REVERSE_OPS are properly done as PUSHDATA.

junderw:
Thanks for the fix and tests.

Could you also add a test for fromASM similar to how this edge case test for toASM works?:

https://github.com/bitcoinjs/bitcoinjs-lib/blob/e0c3ede2bc24eef83f700982cd432173f87875d4/test/script.spec.ts#L44-L57

See how this test is testing an edge case for toASM (making sure empty array is OP_FALSE)

Can you add an edge case test for fromASM testing that `'82 83 84'` becomes `[1, 130, 1, 131, 1, 132]`

decentraldev1:
> Thanks for the fix and tests.
> 
> Could you also add a test for fromASM similar to how this edge case test for toASM works?:
> 
> https://github.com/bitcoinjs/bitcoinjs-lib/blob/e0c3ede2bc24eef83f700982cd432173f87875d4/test/script.spec.ts#L44-L57
> 
> See how this test is testing an edge case for toASM (making sure empty array is OP_FALSE)
> 
> Can you add an edge case test for fromASM testing that `'82 83 84'` becomes `[1, 130, 1, 131, 1, 132]`

of course

test/script.spec.ts
```diff
@@ -41,6 +41,29 @@ describe('script', () => {
...
+    const OPS = bscript.OPS;
+    it('decodes OP_FALSE as empty buffer', () => {
+      const string = 'OP_RETURN OP_FALSE';
+      assert.deepStrictEqual(
+        bscript.fromASM(string),
+        Uint8Array.from([OPS.OP_RETURN, OPS.OP_FALSE]),
+      );
+    });
+
+    it("decodes a series of numbers from '82 to 96' correctly", () => {
```
junderw, line 54: 82 to 90 (130 + 14 = 144 = 128 + 16 = 10010000 = 0x90)

Also maybe "hex numbers" or "numbers in hexidecimal"

Thanks for these tests. I think this is the last fix.

junderw (approved):

Issue #1862, opened 2023-01-10
Some functions cannot be used directly because they are not exported.


h76oeI6pMxU9g4p8aCpc6Q:
For example:
```ts
import { toXOnly, tapTreeToList, tapTreeFromList } from '../../src/psbt/bip371';
import { witnessStackToScriptWitness } from '../../src/psbt/psbtutils';
```
Users should not required to manually access the "node_modules" directory to get these functions after they are alreadly installed bitcoinjs-lib from npm.

bitcoinjs-lib should exporting it explicitly.

I hope I can import these functions directly on the next npm version of bitcoinjs-lib.

junderw:
Users should not import functions that are not exposed publicly.

That is correct.

This issue is making an implicit assumption that these functions MUST be exported, but they aren't.

---

I think what you meant to say was "I think these functions are useful and would like to see them exported publicly in future versions. These are some of my use cases for them."

junderw:
Keep in mind, helper functions that we use internally are sometimes explicitly not exported because we don't want to commit to their interface.

ie. If you import them like you do in your example, your app might break during a patch or minor upgrade of bitcoinjs-lib, since they are not a part of the public API, I could flip the order of the arguments if I wanted to, and I wouldn't need to bump the major version of this library.

Bzvwf16WSDKoSZd8:
The example code is copy from the integration test, not written by me.
https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/test/integration/taproot.spec.ts
```ts
import { toXOnly, tapTreeToList, tapTreeFromList } from '../../src/psbt/bip371';
import { witnessStackToScriptWitness } from '../../src/psbt/psbtutils';
```
These function are neccesseary when doing some functions which bitcoinjs-lib aim to providing to users., e.g. custom script with custom finalizer (both p2sh or p2wsh), and some taproot (p2tr) operations.

I found that already had a opened issues on Nov 2, 2020 with the similar problem, but still not have any response.
https://github.com/bitcoinjs/bitcoinjs-lib/issues/1648

I think the best solution is making those function completely internally. That mean toXOnly, tapTreeToList, tapTreeFromList, witnessStackToScriptWitness are done internally. No longer need users to importing to when writing a custom script with custom finalizer, and doing some taproot operations.

Bzvwf16WSDKoSZd8:
[...]

PR #2191, opened 2024-12-14
export toXOnly

Heracles4179:
handle issue: https://github.com/bitcoinjs/bitcoinjs-lib/issues/1862

Although this function is small, it is often used when building Taproot type transactions, and exporting it is also a good choice


junderw:
Please add tests when changing the public API surface.

Thanks

junderw (approved):

PR #2202, opened 2025-01-24
Fix transaction version field not properly serialized

adambor:
The transaction's version is serialized as an int, instead of uint, disallowing serializing transactions with versions >0x80000000 and allowing serializing transactions with negative version field.

### DIFF
--- a/src/cjs/transaction.cjs
+++ b/src/cjs/transaction.cjs
@@ -99,7 +99,7 @@ class Transaction {
   static fromBuffer(buffer, _NO_STRICT) {
     const bufferReader = new bufferutils_js_1.BufferReader(buffer);
     const tx = new Transaction();
-    tx.version = bufferReader.readInt32();
+    tx.version = bufferReader.readUInt32();
     const marker = bufferReader.readUInt8();
     const flag = bufferReader.readUInt8();
     let hasWitnesses = false;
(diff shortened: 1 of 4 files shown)


Armss9936 (approved):

andrewtoth (approved):
ACK

Since https://github.com/bitcoin/bitcoin/pull/29325 it is stored in `bitcoind` as a `uint32_t`, and now in Bitcoin Core v28 the JSON-RPC response returns it as an unsigned 32 bit integers. This makes transactions `c659729a7fea5071361c2c1a68551ca2bf77679b27086cc415adeeb03852e369` and `35e79ee733fad376e76d16d1f10088273c2f4c2eaba1374a837378a88e530005` fail to parse.

andrewtoth (changes requested):
Actually, (this line)[https://github.com/bitcoinjs/bitcoinjs-lib/pull/2202/files#diff-9bb3e9634fa4b7c5d0d21e142dd235c18ea52709bb8888fd09adb06f4006115aR597] needs to be updated to use `writeUInt32` as well.

Armss9936 (approved):

jasonandjay (approved):

junderw:
Fixed in #2219

PR #2269, opened 2025-10-08
chore(payment): Implement silent payments

jvgelder:
Draft for now looking for a bit of feedback on how we can expose the functionality to the developer

Implements [bip352](https://github.com/bitcoin/bips/blob/master/bip-0352.mediawiki)
Note it tests against vectors taken from: https://github.com/bitcoin/bips/blob/master/bip-0352/send_and_receive_test_vectors.json

TODO:
- [ ] Rework to proper interface to make partial tx creation and external signing possible (bip375 PSBT with abstract signing interface?)
- [ ] Create a example scanner service which contains the proper interfaces but implementation can be transitioned to secp256k1 implementation
- [ ] Decide whether to support https://github.com/bitcoin/bips/pull/2026

Do note that when this https://github.com/bitcoin-core/secp256k1/pull/1765 gets merged and we make it available in
https://github.com/rust-bitcoin/rust-secp256k1/pull/876 and later in https://github.com/bitcoinjs/tiny-secp256k1. Part of code created in this pr will become redundant but hopefully we have proper interfaces in place to make the transition smoothly.

# Resources

## Implementations

- Bluewallet TS implementation: https://github.com/BlueWallet/SilentPayments/blob/master/src/index.ts
- Blind bit with PSBT:  https://github.com/setavenger/blindbit-wallet-cli/blob/dc555fb7ce2d6cb79f38a8444bfd518e91e35c70/pkg/wallet/send.go#L316
- Sparrow: https://github.com/sparrowwallet/drongo/tree/master/src/main/java/com/sparrowwallet/drongo/silentpayments
- https://github.com/cygnet3/rust-silentpayments
- Bip375 example implementation : https://github.com/macgyver13/bips/tree/bip375-reference-testvectors/bip-0375

## Regarding encoding formats for various combinations of keys:

- https://github.com/bitcoin/bips/pull/2026
- https://delvingbitcoin.org/t/bip352-private-key-formats/2080
- https://btctranscripts.com/bitcoin-core-dev-tech/2024-04/silent-payment-descriptors

## PSBT support discussion:

https://delvingbitcoin.org/t/bip352-psbt-support/877

## External tweak data

- https://block-dn.org/


Armss9936 (approved):

junderw:
I've been in discussions on the best way to handle this, and I think it would be a better idea to pass in a cached key.

Instead of calculating the sum of the inputs in real time every time, The payment should only take a special key ie. `inputAggregate` or something with an easier name. Then separate from the Payments API we would offer a helper function that can calculate the aggregate key from given inputs array.

This is primarily because our friends over at BlueWallet see a future where apps like electrs or mempool etc will pre-calculate the aggregate keys and index them for us....

But... yeah... ballooning Payments API into something that encompasses a whole transaction seems like a bit much. At that point we might as well turn it into a method on the Psbt.

junderw:
@Overtorment if you'd like to chime in, feel free.

jvgelder:
> I've been in discussions on the best way to handle this, and I think it would be a better idea to pass in a cached key.
> 
> Instead of calculating the sum of the inputs in real time every time, The payment should only take a special key ie. `inputAggregate` or something with an easier name. Then separate from the Payments API we would offer a helper function that can calculate the aggregate key from given inputs array.
> 
> This is primarily because our friends over at BlueWallet see a future where apps like electrs or mempool etc will pre-calculate the aggregate keys and index them for us....
> 
> But... yeah... ballooning Payments API into something that encompasses a whole transaction seems like a bit much. At that point we might as well turn it into a method on the Psbt.

ye maybe use the following logic:

- If shared secret `S?: Uint8Array` and `inputTweak?: Uint8Array` are set we go straight to `deriveOutputs`
- else if `inputs?: Input[]` and `privKeys?: Array<{ priv: Uint8Array; isXOnly: boolean }>` we do calculate the tweak and shared secret 
- else we throw an error.

1. Is there a use case to allow the user to provide A_sum instead of the private keys?
1. Besides returning outputs we may also want to return a transaction that can be signed or is this not the place to do that?
1. Similar to the point above do we want to return signed outputs?

Overtorment:
ill take a look.

btw we have a TS implementation of SP here: https://github.com/BlueWallet/SilentPayments/

jvgelder:
> > I've been in discussions on the best way to handle this, and I think it would be a better idea to pass in a cached key.
> > Instead of calculating the sum of the inputs in real time every time, The payment should only take a special key ie. `inputAggregate` or something with an easier name. Then separate from the Payments API we would offer a helper function that can calculate the aggregate key from given inputs array.
> > This is primarily because our friends over at BlueWallet see a future where apps like electrs or mempool etc will pre-calculate the aggregate keys and index them for us....
> > But... yeah... ballooning Payments API into something that encompasses a whole transaction seems like a bit much. At that point we might as well turn it into a method on the Psbt.
> 
> ye maybe use the following logic:
> 
>     * If shared secret `S?: Uint8Array` and `inputTweak?: Uint8Array` are set we go straight to `deriveOutputs`
> 
>     * else if `inputs?: Input[]` and `privKeys?: Array<{ priv: Uint8Array; isXOnly: boolean }>` we do calculate the tweak and shared secret
> 
>     * else we throw an error.
> 
> 
>     1. Is there a use case to allow the user to provide A_sum instead of the private keys?
> 
>     2. Besides returning outputs we may also want to return a transaction that can be signed or is this not the place to do that?
> 
>     3. Similar to the point above do we want to return signed outputs?

Allow various cases for deriving of outputs : https://github.com/bitcoinjs/bitcoinjs-lib/blob/4f52dde9292fbbeb1a5a29e0da4fccf3b3448a13/ts_src/payments/p2sp.ts#L89

ts_src/payments/p2sp.ts
```diff
@@ -0,0 +1,665 @@
...
+    if (a.address) return a.address;
+    const version = a.version !== undefined ? a.version : 0;
+    return encodeSilentPaymentAddress(
+      o.spendPubkey!,
+      o.scanPubkey!,
+      version,
+      network,
+    );
+  });
+  lazy.prop(o, 'outputs', () => {
```
jvgelder: This is something I am unsure about.

1. Should outputs be the ones create to send?
2. Where and how do we store resulting UTXO when we scanned?

jvgelder:
> ill take a look.
> 
> btw we have a TS implementation of SP here: https://github.com/BlueWallet/SilentPayments/

Ye I mentioned on the top :) (thought I referred to to something inside the core bluewallet app). 
Reason for this PR is that I think most of the code belongs in bitcoinjslib (see the use of tagged hash for example).
If you think you have more elegant parts feel free to point out / overwrite some of my code here ( I am not that familiar with typescript)

junderw:
> ye maybe use the following logic:
> 
> * If shared secret `S?: Uint8Array` and `inputTweak?: Uint8Array` are set we go straight to `deriveOutputs`
> * else if `inputs?: Input[]` and `privKeys?: Array<{ priv: Uint8Array; isXOnly: boolean }>` we do calculate the tweak and shared secret
> * else we throw an error.
> 
> 1. Is there a use case to allow the user to provide A_sum instead of the private keys?
> 2. Besides returning outputs we may also want to return a transaction that can be signed or is this not the place to do that?
> 3. Similar to the point above do we want to return signed outputs?

I would like to repeat what I said before:

> ballooning Payments API into something that encompasses a whole transaction seems like a bit much. At that point we might as well turn it into a method on the Psbt.

It seems like your reply is ignoring this and talking about placing all of that inside the p2sp Payment API still... am I misunderstanding your response? Could you please clarify?

jvgelder:
> > ye maybe use the following logic:
> > 
> > * If shared secret `S?: Uint8Array` and `inputTweak?: Uint8Array` are set we go straight to `deriveOutputs`
> > * else if `inputs?: Input[]` and `privKeys?: Array<{ priv: Uint8Array; isXOnly: boolean }>` we do calculate the tweak and shared secret
> > * else we throw an error.
> > 
> > 
> > 1. Is there a use case to allow the user to provide A_sum instead of the private keys?
> > 2. Besides returning outputs we may also want to return a transaction that can be signed or is this not the place to do that?
> > 3. Similar to the point above do we want to return signed outputs?
> 
> I would like to repeat what I said before:
> 
> > ballooning Payments API into something that encompasses a whole transaction seems like a bit much. At that point we might as well turn it into a method on the Psbt.
> 
> It seems like your reply is ignoring this and talking about placing all of that inside the p2sp Payment API still... am I misunderstanding your response? Could you please clarify?

- scratch point 2 and 3 indeed.

For the inputs/outputs i was thinking of a bare minimal set of data, the parsing of the transactions into receiving inputs would be done elsewhere. I think the lazy output property kinda reflects this now right? or do you still think even that is too much here?

Overtorment:
i feel a bit out of element here, every time i get distracted from SP i completely forget how they work and getting familiar again means re-reading all the specs again, and it takes time.

are there any specific questions for me? if yes, dumb it down for me.
otherwise, the standalone lib https://github.com/BlueWallet/SilentPayments/ does the job well, and embeds nicely into whatever js library for bitcoin you are going to use (not necessarily bitcoinjs-lib, more modern js apps choose to use scure-btc-signer instead )

jvgelder:
> i feel a bit out of element here, every time i get distracted from SP i completely forget how they work and getting familiar again means re-reading all the specs again, and it takes time.
> 
> are there any specific questions for me? if yes, dumb it down for me. otherwise, the standalone lib https://github.com/BlueWallet/SilentPayments/ does the job well, and embeds nicely into whatever js library for bitcoin you are going to use (not necessarily bitcoinjs-lib, more modern js apps choose to use scure-btc-signer instead )

no worries thanks for making me aware of scure-btc-signer also mentioned the blue wallet repo there as there was an issue open for silent payments too.

ts_src/payments/p2sp.ts
```diff
@@ -0,0 +1,665 @@
...
+        const S = calculateSharedSecret(
+          inputHashTweak,
+          value.B_spend_pub,
+          aSum,
+        );
+        deriveOutput(S, value.B_spend_pub, index);
+      });
+    }
+    // If we have all the inputs, aSum and only the spend keys for the recipients we need to calculate the input hash and secret
+    else if (
```
jvgelder: this case drags in inputs (hashes) and we may not want to have this here

ts_src/payments/p2sp.ts
```diff
@@ -0,0 +1,665 @@
...
+        const S = calculateSharedSecret(
+          inputHashTweak,
+          value.B_spend_pub,
+          a.aSum,
+        );
+        deriveOutput(S, value.B_spend_pub, index);
+      });
+    }
+    // If we have all the inputs, privKeys and only the spend keys for the recipients we need to calculate the Sum, the input hash and secret
+    else if (
```
jvgelder: same here for inputs

ts_src/payments/p2sp.ts
```diff
@@ -0,0 +1,611 @@
...
+  for (let i = 0; i < 32; i++) if (a[i] !== 0) return false;
+  return true;
+}
+
+/**
+ * Smallest outpoint = lexicographic min of (txidLE || voutLE)
+ * @param inputs an array of inputs you want the first lexicographically sorted result
+ * @returns the first output after sorting lexicographically
+ */
+export const findSmallestOutpoint = (inputs: Array<Input>) =>
```
jvgelder: Maybe make this part of Transaction ? or create a SilentPaymentTransaction that extends transaction ?

ts_src/payments/p2sp.ts
```diff
@@ -0,0 +1,611 @@
...
+/**
+ * Scans a transaction's inputs and outputs to find any silent payments for the receiver.
+ * @param receiverScanPrivkey - b_scan
+ * @param receiverSpendPrivkey - b_spend
+ * @param inputHashTweak
+ * @param summedSenderPubkey - A_sum
+ * @param outputsToCheck - array of hex xOnly encoded outputs to check
+ * @param labelNonces
+ */
+export function scanForSilentPayments(
```
jvgelder: May want to move this to a SilentPaymentTransaction class which allows you to process an entire transaction to see if there are any outputs for you

ts_src/payments/p2sp.ts
```diff
@@ -0,0 +1,611 @@
...
+    .sort((a, b) => tools.compare(a, b))[0];
+
+/**
+ * Serialize output with number little endian
+ * (used to sort outputs)
+ * @param txidHexBE - big endian encoded tx
+ * @param vout - output index
+ * @returns the serialized little endian encoded output
+ */
+export const serOutpointLE = (txidHexBE: Uint8Array, vout: number) => {
```
jvgelder: Should be moved with `findSmallestOutpoint`

ts_src/SilentPaymentTransaction.ts
```diff
@@ -0,0 +1,374 @@
...
+    vout: number;
+    value: number;
+  };
+}
+
+/**
+ * Extends the bitcoinjs-lib Transaction class to provide methods
+ * for creating and scanning for BIP-352 Silent Payments.
+ */
+export class SilentPaymentTransaction extends bitcoin.Transaction {
```
jvgelder: Hopefully a class like this makes more sense @junderw 
I do think some functions should change to make it easier and more clear for the developer

ts_src/SilentPaymentTransaction.ts
```diff
@@ -0,0 +1,374 @@
...
+  /**
+   * Scans a finalized transaction to find any outputs belonging to the receiver.
+   * @param tx The finalized bitcoinjs-lib Transaction object.
+   * @param prevouts An array of Previous Outputs being spent by the transaction, used to get prevout scripts.
+   * @param receiverKeys The receiver's scan and spend private keys.
+   * @param labelNonces An optional array of label nonces to scan for.
+   * @returns An array of found silent payment outputs with spending information.
+   */
+  static scan(
+    tx: Transaction,
```
jvgelder: Instead of passing the TX we can just use the current transaction

Armss6556 (approved):

junderw:
This PR is getting to be way out of scope for this library.

Wipe everything and start from the beginning.

What is your goal? Don't be vague. "To support BIP352" is vague. Describing a list of all the things you did before planning anything is also vague... it doesn't tell me your intentions.

1. What are your intentions?
2. What are the requirements?
3. What are the specific changes you need to fill each of the requirements?

Let's start there.

junderw:
Stop coding. Discussions only for now.

jvgelder:
1. I try to get Silent payments in this lib so that its easy for anyone already depending on this library to support spending and receiving from and to silent payment addresses
2. We need:
- to be able to generate a silent payment address to receive funds on
- to be able to spend to a silent payment address
- To be able to find payments to our silent payment addresses
3. 
- Silent payment address decoding / encoding
- Silent output derivation
- Scanning transactions to find payments that belong to one of our silent payment addresses
- Optionally we can have an external provider make us aware of UTXOs which belong to us

Note that the silent payment address looks one way when you share it and different on chain as the addresses are derived from the silent payment address.

junderw:
1. Silent payment address decoding / encoding
    1. Currently the address module does output > address, address > output conversions. This will be difficult to include here since the output > address doesn't have a direct 1:1 mapping. address > output could be possible if we create a separate function for SP only, and pass in the information necessary. However, the dependency graph of necessary information to create the output is complicated and lends itself well to the Payments API's lazy getters. So if you have t_k and B_spend, you don't need to pass in an array of A_n pubkeys or privkeys at all.
2. Silent output derivation
    1. This can be done with Payments API via lazy getters, but the problem is the dependency graph has two "paths" the sender (private a_n public B_spend) and the receiver (Public A_n private b_spend) which we should DEFINITELY support both paths, while validating for nonsensical paths.
3. Scanning transactions + external provider (3 & 4)
    1. I think this is where we start to get out of scope.

---

I think that perhaps instead of using the existing Payments API, we should create a new SilentPayment class that doesn't extend Payment, it just exists on its own.

It contains either a pub or priv of both spend and scan keys. It should have a static constructor method fromStealthAddress(s: string): SilentPayment that creates one with both public keys.

Starting from this baseline, we could try to make this class interact with Transaction and Psbt.

I don't think any code in bitcoinjs-lib should try to connect to RPC or some external service. It should be a simple data in data out API. ie. pass in a Transaction and Uint8Array[] (where each scriptPubkey Uint8Array corresponds to the same index input of the Transaction) Or maybe as simple as inputs, outputs (the outputs spent by the inputs)

This might even be better suited for its own library that depends on bitcoinjs-lib... you could do that and just upstream the minimal changes you need from us to better support your library.

I'm still open to including it in bitcoinjs-lib, but one thing I'm a bit unsure about is the Psbt support.

For sending TO a stealth address, there should really be stealth payment info embedded in the Psbt so that we can update the output addresses if someone adds an input to a partially signed Psbt.

I believe the BIP352 authors are currently discussing how to deal with Psbt...

However, we DID release P2TR support without Psbt support initially (and it was very hacky and no one liked it) so I'm open to all ideas.

Please discuss a concrete plan for an API (ie. like a d.ts style pseudo-type-declaration just to give a rough idea) before you start coding though so we don't waste your time.

If you decide to go the separate library route, feel free to mention me on any issues you have over there and I'll be happy to join the discussion.

jvgelder:
I think its good if the base is this in library as I assume it will get more eyeballs and checks from more people.

I tried to come with a very basic layout but its tougher than I thought, I dont want to expose the developer to too much of the internals. Should be as simple as I want to send 0.2 to silent payment address sp213123...


# initial spending and receiving layout

```typescript
class SilentPayment{
   
      /**
      * Encode a Silent payment address for others to spend to
      * Encodes spend and scan public keys into a Bech32m Silent Payment address.
      * @param {Uint8Array} [B_spend] - spend public key.
      * @param {Uint8Array} [B_scan] - public scan key.
      * @param {number} [version] - Optional version number of the silent payment scheme.
      * @param network - testing, regtest or prod
      * @returns bech32m encoded string
      */
      export function encodeSilentPaymentAddress(
          B_spend: Uint8Array,
          B_scan: Uint8Array,
          version = 0,
          network: Network = BITCOIN_NETWORK,
      ): string 

      // TODO what kind of output do we want here ?
      fuction createOutput(inputs : Input[],  recipients :  {silentAddr: string, value : number}[]) :  ?
      
     /**
     * Convenience function to create outputs from a transaction
     * Note: Silent payment depend on inputs, so if any input changes the output also changes
     * This function removes all other outputs already set on the transaction to prevent outputs created on different input 
     * set.
     * @param recipients - list of silent payment recipients
     * @param tx - Partially build transactions already containing the required inputs
     * @returns a transaction with 
     */
     export function createSilentPaymentTransaction(recipients :  {silentAddr: string, value : number}[], tx: Transaction) : Transaction
     
     /**
     Storage interface for silent address and generated in/output addresses
     Used to display to the user which on chain addresses belong to which silent payment addresses
     */
     interface SilentPaymentOutputGroup{
        silentPaymentAddress : string
       sent : {txid : string, vout: number, pub_key : string, destSilentPaymentAddress : string}[],
       received : {txid : string, vout: number, pub_key : string, originSilentPaymentAddress?: string}[]
     }
}
```


# Scanning

I think we may want to move scanning to a different PR and first focus on spending and sending.
However i did create a few options for scanning results sorted by trust level as it may influence how we structure code.
Hopefully external service are created on top of the low trust model so that privacy is preserved
For doing business with entities that require travel rule it might be convenient to share the shared_secret S
This gives on chain privacy and gives regulatory compliance

```typescript
class SilentPayment{


      /**
      verify if given outputs really belong to us
      @param potentialOutputs - List of outputs potentially to be for us
      @param S - shared secret
      @param B_spend - public spend key
      @param b_scan - private scan key
      @returns a list of SilentPaymentOutputs that can be used to construct an actual p2tr output or Pbst
      */
      verifyScanResult(potentialOutputs : Output[], S: Uint8Array, B_spend : Uint8Array,  b_scan : Uint8Array) : SilentPaymentOutput []

      /**
      scan if the given outputs belong to us
      @param  tk - tweaked hash
      @param potentialOutputs - List of outputs potentially to be for us
      @param B_spend - public spend key ( to check if 
      @param b_scan - private scan key
      @returns a list of SilentPaymentOutputs that can be used to construct an actual bip341 output
      */
      scan(tk: Uint8Array, potentialOutputs : Output[], B_spend : Uint8Array,  b_scan : Uint8Array) : SilentPaymentOutput []

      /**
      check if a given transaction contains UTXO for our key public key B_scan
      @param  decodedTx - decoded transaction
      @param B_spend - public spend key
      @param b_scan - private scan key
      */
      scan(decodedTx: Transaction, B_spend : Uint8Array,  b_scan : Uint8Array) : SilentPaymentOutput []
      /**
      check if a given transaction contains UTXO for our key public key B_scan
      @param  rawTx - raw transaction bytes
      @param B_spend - public spend key
      @param b_scan - private scan key
      @returns a list of SilentPaymentOutputs that can be used to construct an actual p2tr output or Pbst
      */
      scan( rawTx: Uint8Array,  B_spend : Uint8Array, b_scan : Uint8Array) : SilentPaymentOutput []
      
     
}
```

junderw:
> createOutput

1. We would need the private keys to every eligible Input's pubkey.

> createSilentPaymentTransaction

1. Same as createOutput, we need some way to represent the fact that we have all the private keys for all eligible inputs' pubkeys.

> SilentPaymentOutputGroup

I didn't understand the use case for this, nor how a hypothetical function that output this interface would work.

> scan

Without the previous outputs (pointed at by the input) it's impossible to know the public key we need for creating the aggregate pubkey.

---

I am thinking more along the lines of:

1. SilentPayment contains 2 keys as attributes on the class. These can both be either private or public.
2. If both are public. It has methods to output a stealth address. It has a method to generate an output script based on passing in the aggregate private key and lowest lexical outpoint as an argument.
3. If scan is private, enable methods for scanning a given transaction + previous outputs array. Maybe add intermediate methods that can handle intermediate states (ie just passing in the shared secret, or just passing in the aggregate pubkey and lowest serialized outpoint etc.
4. If spend is private, add a method to output a labelled stealth address (since we have the spend private we can change the Bm pubkey) also add a method to output the private key for a given output we own (similar to the scan function without spend private, but in the spend private case the private key for the output should be given so that we can spend the transaction like normal.

I think this would be a good minimal interface to start with.

jvgelder:
> > createOutput
> 
>     1. We would need the private keys to every eligible Input's pubkey.
Ye and i think we want to support mixing different type of inputs? What is the best object type to use here?
> 
> > createSilentPaymentTransaction
> 
>     1. Same as createOutput, we need some way to represent the fact that we have all the private keys for all eligible inputs' pubkeys.
Ye was hoping we could do this on sign but we depend on `a_sum` for creating the shared secret `S` :/
> 
> 
> > SilentPaymentOutputGroup
> 
> I didn't understand the use case for this, nor how a hypothetical function that output this interface would work.

I think we should provide a structure to group silent addresses and their onchain outputs, so that the developer can develop UX where the user is aware of what he sent to and is not surprised  with a different onchain address.
> 
> > scan
> 
> Without the previous outputs (pointed at by the input) it's impossible to know the public key we need for creating the aggregate pubkey.
> 
> I am thinking more along the lines of:
> 
>     1. SilentPayment contains 2 keys as attributes on the class. These can both be either private or public.
Something along the lines of `scan : Uint8Array` and `spend: Uint8Array` and then check the prefix? or should we make it a bit more fool proof and use different structures ?
>     2. If both are public. It has methods to output a stealth address. It has a method to generate an output script based on passing in the aggregate private key and lowest lexical outpoint as an argument.

If either one is a private key I guess we can just get the pubKey that belongs to that private key and still output the address?
```typescript
  lazy.prop(o, 'silentPaymentAddress', () => {
    if (!a.scan || !a.spend) return;
    const scanPubKey = if( isPub(a.scan)) a.scan else getPub(a.scan)
    const spendPubKey = if( isPub(a.spend)) a.scan else getPub(a.spend)
    return encodeSilentPaymentAddress(scanPubKey, spendPubKey)
  });
```

Regarding creating the output something like:
```typescript
  lazy.prop(o, 'createOutputScript', () => {
    if (!a.B_spend || a.k === undefined) return null;
    
    if ( a.aSum && a.outPointL  && a.B_scan ) {
       return createOutputScript(a.aSum, a.outPointL, a.B_spend, a.B_scan, a.k, a.labelTweak)
    } else if(a.S ){
       return createOutputScript(a.S, a.B_spend, a.k, a.labelTweak)
    } else return null
  });

  function createOutputScript(aSum : Uint8array, outPointL : Uint8Array, B_spend : Uint8Array,  B_scan : Uint8Array, k: number, labelTweak?: Uint8Array = null)
  function createOutputScript(S : Uint8array,  B_spend : Uint8Array, k: number, labelTweak?: Uint8Array = null)
```

> 
>     3. If scan is private, enable methods for scanning a given transaction + previous outputs array. Maybe add intermediate methods that can handle intermediate states (ie just passing in the shared secret, or just passing in the aggregate pubkey and lowest serialized outpoint etc.

```typescript
 //Do we want something like this ?
  lazy.prop(o, 'scan', () => {
    if (!a.scan || !isPriv(a.scan) || !a.prevOuts || !a.prevOuts.length ) return;
    return scan(a.scan, a.prevOuts, a.transaction)
  });

 function scan(prevOuts : Output[], tx : Transaction ){
   if( !scan || !isPriv(scan)) return null
   return scan(scan, prevOuts : Output[], tx : Transaction)
 }
 ....
 // We can make similair calls for other functions mentioned here
 function scan(prevOuts : Output[], tx : Transaction)
 function scan(outputsToCheck : Output[], inputHash: Uint8Array, B_spend : Uint8Array,  b_scan : Uint8Array)
 function scan(outputsToCheck : Output[], S: Uint8Array, B_spend : Uint8Array)    
```

What should the output of `scan` be?

>     4. If spend is private, add a method to output a labelled stealth address (since we have the spend private we can change the Bm pubkey) also add a method to output the private key for a given output we own (similar to the scan function without spend private, but in the spend private case the private key for the output should be given so that we can spend the transaction like normal.

```typescript
   lazy.prop(o, 'labledStealthAddress', () => {
    if (!a.scan || !ecc.isPrivate(a.scan) || a.m === undefined || !a.spend) return null;
    const spendPubkey = if( isPriv(a.spend)) getPub(a.spend) else spend
    return labledStealthAddress(a.scan, spendPubKey, a.m)
  });
```

TODO priv key for given output (will get back to that later ;) )

> 
> I think this would be a good minimal interface to start with.

jvgelder:
Not sure what we want to do with outputs and whether to offer `labledStealthAddress` as function or attribute and what object type do we want to use for both the output of scan and input of prevout.

Something along the lines of:

```typescript
export interface UTXO {
  txid: Uint8Array; vout: number; value?: bigint;
  xOnlyPubkey: Uint8Array; t_k: Uint8Array; k: number; labelNonce?: number;
}

export interface SilentPayment {
  spend?: Uint8Array;          // 32B b_spend (priv) or 33B B_spend (compressed pub)
  scan?: Uint8Array;           // 32B b_scan (priv)  or 33B B_scan (compressed pub)
  sharedSecret?: Uint8Array;   // 33B ser_P(S) (optional fast-path)
  version: number;             // BIP-352 v0
  aSum?: Uint8Array;           // 32B a_sum (sender path)
  outpointL?: Uint8Array;      // 36B txidLE||voutLE (min over ALL inputs)
  k?: number;                  // uint32 output index for this recipient
  labelTweak?: Uint8Array;     // 32B L_m (optional)
  network?: Network;
  address?: string;
  // derived (lazy):
  publicScan?: Uint8Array;     // 33B B_scan
  publicSpend?: Uint8Array;    // 33B B_spend
  outputScript?: {
    scriptPubKey: Uint8Array;  // OP_1 <xonly>
    xOnlyPubkey: Uint8Array;   // 32B
    tweak: Uint8Array;         // 32B t_k or t_k+L
    k: number;
  } | null;
  labeledStealthAddress(m : number) : string | null
  scanFromSummary(
    outputs: Output[],           
    A_serP: Uint8Array,                   // 33B
    outpoint_L: Uint8Array,               // 36B
    opts?: { labelNonces?: number[] }
  ): UTXO[];

  scanFromS(
    outputs: Output[],
    S: Uint8Array,                        // 33B ser_P(S)
    opts?: { labelTweaks?: Uint8Array[] } // precomputed L_m
  ): UTXO[];
}

// ---- The wrapper ----
export function silentPayment(a: SilentPayment, opts?: PaymentOpts): SilentPayment {
  const network = a.network || BITCOIN_NETWORK;
  const o: SilentPayment = { version: 0, network };
  opts = Object.assign({ validate: true }, opts || {});

  // Derive public keys (compressed) from provided priv-or-pub inputs
  lazy.prop(o, 'publicScan', () => {
    if (!a.scan) return undefined;
    return isPrivate(a.scan) ? pubFromPriv(a.scan) : (isPointCompressed(a.scan) ? a.scan : undefined);
  });

  lazy.prop(o, 'publicSpend', () => {
    if (!a.spend) return undefined;
    return isPrivate(a.spend) ? pubFromPriv(a.spend) : (isPointCompressed(a.spend) ? a.spend : undefined);
  });

  // Address (bech32m) — needs both compressed pubs
  lazy.prop(o, 'silentPaymentAddress', () => {
    if (a.address) return a.address;
    if (!o.publicSpend || !o.publicScan) return undefined;
    return encodeSilentPaymentAddress(o.publicSpend, o.publicScan, 0, o.network!);
  });

  // Labeled address helper (requires b_scan)
  lazy.prop(o, 'labeledStealthAddress', () => {
    const m = (a as any).labelNonce as number | undefined; // or however you pass it
    if (m === undefined) return null;

    // need b_scan (private) to compute L, but publish with B_scan from the lazy prop
    if (!a.scan || !isPrivate(a.scan)) return null;
    if (!o.publicSpend || !o.publicScan) return null;

    const L = createLabelTweak(a.scan, m >>> 0);           // range-checked
    const B_m = pointAddScalar(o.publicSpend, L, true);
    if (!B_m) return null;

    return encodeSilentPaymentAddress(B_m, o.publicScan, 0, o.network!);
  });
  // or as function 
  
  o.labeledStealthAddress = (m ) => {
    if (!a.scan || !isPrivate(a.scan)) return null;
    if (!o.publicSpend || !o.publicScan) return null;

    const L = createLabelTweak(a.scan, m >>> 0);           // range-checked
    const B_m = pointAddScalar(o.publicSpend, L, true);
    if (!B_m) return null;

    return encodeSilentPaymentAddress(B_m, o.publicScan, 0, o.network!);
  };
  

  // Output script (two forms: S, or aSum+outpointL+B_scan)
 // TODO: Make a clear distinction between sender and receiver information as you can see i alreaydy made the mistake of creating an outputScript for our own key, we may want to expose this as a function instead.
  lazy.prop(o, 'outputScript', () => {
    if (!o.publicSpend || a.k === undefined) return null;

    if (a.sharedSecret) {
      return createOutputScriptFromS(a.sharedSecret, o.publicSpend, a.k >>> 0, a.labelTweak);
    }

    if (a.aSum && a.outpointL && o.publicScan) {
      return createOutputScriptFromASum(a.aSum, a.outpointL, o.publicScan, o.publicSpend, a.k >>> 0, a.labelTweak);
    }
    return null;
  });
  


  o.scanFromS = (outputsXOnly, S, opts) => {
    if (!o.publicSpend) return [];
    return scanFromS(outputsXOnly, S, o.publicSpend)
  };

  o.scanFromSummary = (outputsXOnly, A_serP, outpoint_L, opts) => {
    if (!o.publicSpend || !a.scan || !isPrivate(a.scan)) return [];
    return scanFromSummary(outputsXOnly, A_serP, outpoint_L, o.publicSpend, a.scan)
  };


// Validation (light)
  if (opts.validate) {
    if (a.scan && !isPrivate(a.scan) && !isCompressedPub(a.scan)) {
      throw new TypeError('scan must be 32B priv or 33B compressed pub');
    }
    if (a.spend && !isPrivate(a.spend) && !isCompressedPub(a.spend)) {
      throw new TypeError('spend must be 32B priv or 33B compressed pub');
    }
    if (a.sharedSecret && a.sharedSecret.length !== 33) {
      throw new TypeError('sharedSecret must be 33 bytes (ser_P(S))');
    }
    if (a.outpointL && a.outpointL.length !== 36) {
      throw new TypeError('outpointL must be 36 bytes (txidLE||voutLE)');
    }
    if (a.labelTweak && !isPrivate(a.labelTweak)) {
      throw new TypeError('labelTweak scalar is 0 or >= n');
    }
  }

  return Object.assign(o, a);
}

```

junderw:
Some related things:

Regarding encoding formats for various combinations of keys:

https://github.com/bitcoin/bips/pull/2026

https://delvingbitcoin.org/t/bip352-private-key-formats/2080

https://btctranscripts.com/bitcoin-core-dev-tech/2024-04/silent-payment-descriptors

PSBT support discussion:

https://delvingbitcoin.org/t/bip352-psbt-support/877

jvgelder:
This is possibly offtopic or for a next iteration but I have been talking to people that are using bitcoin in combination with an HSM. See also https://github.com/macgyver13/bips/blob/bip375-reference-testvectors/bip-0375/examples/hardware_signer/README.md

# Labeled address request

Input:
- bScanIdentifier : string
- labelM : number
- Bspend : string // Can leave this out if we do the addition outside of HSM

Output:
- Bm
or
- L // label tweak if we decide to keep the Bspend addition out of the HSM

```typescript
/**
* @param bScan either a private key or the identifier used by the HSM
* @param m : label
* @param Bspend pub spend key
* @returns Bm tweaked spend pub key 
 */
function createLabeledAddress(bScan: string, m : number, Bspend : string) : Bm
// or 2 step createLabeledAddress -> createLabelTweak
/**
* @param bScan either a private key or the identifier used by the HSM
* @param m : label
* @returns L label tweak
*/
function createLabelTweak(bScan: string, m : number) : L
```
# In and Outputs

Since our outputs depend on our inputs we should provide the user either a function to add new in/outputs
or a parameter to provide all the in/outputs to use (question is do we want to update the state or not?). To work with external signers or HSMs we should keep a list of keys and inputs we want to use. Only at the final stage we can be sure of `OutpointL` and the keys we should be using for `aSum`

## Add input

```typescript
/**
* Store new inputs
* @param hash tx hash
* @param index vout index
* @param private key a (or its identifier) to be used to finally create aSum
*/
function addInput(hash: string, index : number, a : string ) // Stores new 
```

## Add output

Input:
- Bscan
- Bm[]
- value per Bm ? // Not sure at which stage we would want to add the value

```typescript
/**
* @property publicSpend spend public key for the receiver
* @property m label to use
* @property value // the value to send this receiver
*/
interface OutputWithValue{
  publicSpend : Uint8Array
  m?: number
  value : number
}

/**
* Add group pay to
* @param receiverSpAddress // bech32m encoded receiver public silent payment adddress
* @param value // value to pay to this output eventually
*/
function addOutput(silentPaymentReceiverAddress  : string, value : number)
/**
* Add group pay to
* @param Bscan // public scan key of the receiver
* @param Bspend // public spend key of the receiver
* @param m label to use
* @param value // value to pay to this output eventually
*/
function addOutput(Bscan  : string, Bspend : string, m?: number,  value : number)
/**
* Add groups pay to
* @param receiverGroups // map of given receiver Bscan as key and array of  Bspend, label with values 
*/
function addOutput(receiverGroups : map<string, OutputWithValue[]>)
```

## Prepare for spending

Input:
- A list of `a` identifiers to create sumA 
- Mapping of inputIndex/hash/something that ids the input -> priv key identifier

https://github.com/bitcoin/bips/blob/master/bip-0375.mediawiki would actually be a great way to communicate which keys should used where (unless I miss something) but its still in draft unfortunately.

We will need `PSBT_OUT_SP_V0_INFO` `PSBT_OUT_SP_V0_LABEL` which would give us the information which key was used and thus which keys to use for aSum.

There are some test_vectors available but not yet merged: https://github.com/macgyver13/bips/tree/bip375-reference-testvectors/bip-0375

Until that time offer a different structure? or implement it and if it changes we update?

We can also observe how psbt is used for sp here: https://github.com/setavenger/blindbit-wallet-cli/blob/dc555fb7ce2d6cb79f38a8444bfd518e91e35c70/pkg/wallet/send.go#L316

```typescript
/**
* Prepare our transaction for signing
* @returns bip375 PSBT
*/
function prepare() : PSBT
```

## sign and Spend

This is where I am really struggling to comprehend things but I hope its a starter. 

```typescript
/**
* Sign a specific input
* @param input index
* @param bscan either or actual scan priv key or an identifier
* @param bspend either or actual spend priv key or an identifier
*/
function psbt.signSilentInput(input : number, bscan : string, bspend : string){
  // Underwater this would call
  const outpointL = extractOutpointL(psbt.inputs)
  const Asum = extractKeysAndSum(psbt.inputs) // need prevouts here ?
  const S = calculateSharedSecret(OutpountL, Asum, bscan)
  // is k equal to our input number?
  const tK = calculateTweak(S,input)
  const m = psbt.inputs[input].label
  const d = createPrivateKey(tk,bspend, m)
   // Something to sign
  psbt.inputs[input].sign(d)
}

or 
function psbt.signSilentInput(input : number, d: string);
```

# TODO: 

- scanning
- decide how and where to make use of sppriv, spwatch etc

jvgelder:
@junderw should we close this pr an create an issue where we breakdown all steps we need to undertake ?

Played a bit more with some potential interface on PSBT:

# Creating psbt outputs

Using: https://github.com/bitcoin/bips/blob/master/bip-0375.mediawiki
```typescript
addSilentPaymentOutput(value: bigint, address: string): this {
    ... Decoding address etc...
    output.unknownKeyVals.push({
      key: new Uint8Array([PSBT_OUT_SP_V0_INFO]),
      value: metaValue,
    });
    return this;
}
```

```typescript
addSilentChangeOutput(
    value: bigint,
    scanPubKey: Uint8Array,
    spendPubKey: Uint8Array,
    label: number = 0
  ): this {
     ... add intent etc ..
    output.unknownKeyVals.push({
      key: new Uint8Array([PSBT_OUT_SP_V0_LABEL]),
      value: labelBuf,
    });

    return this;
  }
```

# Spending inputs

Below uses : https://gist.github.com/nymius/b3dd0b8a08c6735d617e6216b73c4260
```typescript
addSilentInput(inputData: PsbtInputExtended, tweak: Uint8Array): this {
  ... add regular input ...
  input.unknownKeyVals.push({
      key: new Uint8Array([PSBT_IN_TAP_RAW_TWEAK]),
      value: tweak,
    });
    return this;
}
```

Later we can add:
- DLEQ proofs
- Add scanning in separate code

PR #2270, opened 2025-10-13
Removed unnecessary check for tapInternalKey for signature validations of taproot inputs

summraznboi:
`getTaprootHashesForSig` is an overly reused function for both signing as well as signature validation.
To better make each path clearer, we created two wrappers on top to clearly state its intent, `getTaprootHashesForSigValidation` and `getTaprootHashesForSigning`. Note that `getAllTaprootHashesForSig` is just renamed to `getAllTaprootHashesForSigValidation`, since there isn't a corresponding usage for signing.

This separation allows for us to clearly see that for sig validation, we only need to see the existence of tapKeySig to know that this is a key spend path. This is especially important for nonstandard key spend path patterns such as MuSig2 or other offchain multisig signing protocols.


junderw (approved):

summraznboi:
Thanks for merging this! It would be great to have a bug fix release soon so we can make use of it :)

Armss9936 (approved):

PR #2296, opened 2025-11-12
fix: Handle zero-input PSBTs correctly during transaction parsing

saravanan7mani7:
- Guard SegWit marker detection so unsigned transactions with zero inputs don’t misidentify varint bytes as the witness marker
- Reuse the peeked vin count when the marker is legitimate and ensure the buffer has room for the claimed inputs
- Normalize hash caching to explicit Uint8Array instances to satisfy updated noble/hashes typings
- Add BIP174 fixtures covering 0-in/0-out and 0-in/1-out PSBTs and extend Mocha tests to assert creation and round-trip behaviour


saravanan7mani7:
Accidentally opened.

PR #2297, opened 2025-11-23
p2ms: support up to 20 keys

landabaso:
Extend p2ms to support up to 20 public keys (n-of-20 multisig). The previous 16-key cap was only an internal implementation limit, not a Bitcoin rule. In practice, P2SH multisig cannot handle 16 keys anyway due to script-size limits (the effective cap is ~15), but those constraints do not apply to P2WSH. Since p2ms is often wrapped inside P2WSH, the implementation now supports the full 20-key range.

Added new fixtures and a P2MS under P2WSH integration test.


junderw:
Perhaps we can represent the P2SH limitation in the P2SH payment? Like validate the length of the redeem output etc...?

I'd rather people not get random errors from Bitcoin Core when they broadcast. Those are the most annoying errors to debug.

landabaso:
The library already throws an error if you try to build a P2SH-P2MS with more than ~15 pubkeys.
Those checks were added in this PR: [https://github.com/bitcoinjs/bitcoinjs-lib/pull/1863](https://github.com/bitcoinjs/bitcoinjs-lib/pull/1863)

I verified it with integration tests in bitcoinerlab (P2SH with 16 pubkeys _throws_ as expected):
[https://github.com/bitcoinerlab/descriptors/blob/ae73d02d73a597ce44209e416300fc4e85a661eb/test/integration/sortedmulti.ts#L202](https://github.com/bitcoinerlab/descriptors/blob/ae73d02d73a597ce44209e416300fc4e85a661eb/test/integration/sortedmulti.ts#L202)

[That PR](https://github.com/bitcoinjs/bitcoinjs-lib/pull/1863/files#diff-1fd14555c81109d259df1e768458f3b62822b93e5a29d1ac714ea33ac627ddb2) already included a test for a 16-key P2SH-P2MS case.

If you meant something different, just let me know and I'll address it.

junderw (approved):

PR #2302, opened 2025-12-10
Fix: Ensure toXOnly always returns Uint8Array

### LINKED ISSUE #2249: Buffer.equals() requires Buffer, not Uint8Array
The following comparison fails when `toXOnly()` returns a `Buffer` and `outputKey` is a `Uint8Array`:

Image

Even if both values contain the same underlying bytes, `.equals()` does not accept `Uint8Array` arguments directly.

Using `Buffer.from(outputKey)` works as expected:
```
if ((0, bip371_1.toXOnly)(pubkey).equals(Buffer.from(outputKey))) {
  // works
}
```

Version
`6.1.5`

Expected behavior
`outputKey` must be a `Buffer` when used with `.equals()`. Currently, it’s an `Uint8Array`.

Dairus01:
# Fix: Ensure `toXOnly` Always Returns `Uint8Array` (Fixes #2249)

This PR resolves issue **#2249** by updating the `toXOnly` function so it consistently returns a **Uint8Array**, regardless of whether the input is a `Buffer` or a `Uint8Array`.

Previously, when a `Buffer` was passed into `toXOnly`, the function returned a `Buffer`. This created inconsistencies, especially when users relied on behavior like calling `.equals()` on the returned value, assuming it would always be a `Buffer`. In some environments the function could instead return a sliced `Uint8Array`, causing unexpected failures. Code relying on strict `Uint8Array` behavior also ran into conflicts when a `Buffer` was preserved.

By enforcing `return new Uint8Array(...)` for all cases, we guarantee:

- **Consistency**: The output type is always `Uint8Array`.
- **Safety**: No accidental inheritance of `Buffer` specific methods like `.equals` or `.toString`, which may not exist for Uint8Array or in certain runtimes.

## Changes
- Updated `ts_src/psbt/bip371.ts` so that `toXOnly` always returns a new `Uint8Array`, both for 32 byte inputs and for slices.
- Updated `test/bip371.spec.ts`:
  - Replaced `assert.strictEqual` with `assert.deepStrictEqual` to check content equality rather than reference equality.
  - Added `assert.notStrictEqual` to ensure a fresh instance is created and input `Buffer` identity is never preserved.

## Related Issue
Fixes **#2249**

## Verification
- Executed the `reproduce_issue.ts` script created for this bug and confirmed that `xOnly.equals` is no longer defined, eliminating the type confusion described in the issue report.
- Ran `npm run nobuild:unit` and verified that the full test suite passes with no regressions.


jasonandjay:
I don't think we should fix this issue.

We've already clearly told users that toXOnly accepts Uint8Array.

PR #2305, opened 2025-12-28
docs: clarify Browserify + TypeScript usage

COZYTECH:
## Summary
Adds documentation clarifying how to use bitcoinjs-lib with Browserify and TypeScript.

## Motivation
Some users have reported issues importing `.ts` files after bundling their projects with Browserify.
This is due to misunderstanding how TypeScript source files and JavaScript distribution work.  
This PR provides guidance on the correct import pattern and minimal working example.

## Changes
Added a new section in README: `Using bitcoinjs-lib with Browserify and TypeScript`
Explained why `.ts` imports from the package source will fail after bundling
Provided a minimal TypeScript + Browserify example
Suggested installing `@types/node` when using TypeScript


jasonandjay:
Thank you for your contribution. 
The instructions on how to use it are already provided above, so this part is not really necessary.

PR #2312, opened 2026-02-15
feat: add Pay-to-Merkle-Root (P2MR) support per BIP 360

### LINKED ISSUE #2311: Feature Request: Add P2MR (Pay-to-Merkle-Root) support (BIP 360)
## Summary

[BIP 360](https://github.com/bitcoin/bips/blob/master/bip-0360.mediawiki) introduces a new output type: **Pay-to-Merkle-Root (P2MR)**. The BIP was merged into the official BIP repository on Feb 11, 2026 (PR [bitcoin/bips#1670](https://github.com/bitcoin/bips/pull/1670)) with Draft status. It would be valuable for bitcoinjs-lib to add support for this new output type.

## What is P2MR?

P2MR is a new SegWit **version 2** output type that operates with nearly the same functionality as P2TR (Pay-to-Taproot), but with the **key-path spend removed**. It commits directly to the Merkle root of a script tree (tagged as "TapBranch") without including an internal public key.

### Key characteristics:
- **SegWit version 2** — addresses start with `bc1z` (mainnet)
- **Script-path spend only** — no key-path spend (quantum resistant against long exposure attacks)
- **ScriptPubKey**: `OP_2 OP_PUSHBYTES_32 <32-byte merkle_root>`
- **Control block**: `[control_byte] || [32*m byte merkle_path]` (no internal pubkey, 1 + 32*m bytes)
- **Fully compatible with Tapscript (BIP 342)** — existing tapscript programs can be used without modification
- **Reuses existing tagged hashes**: `TapLeaf`, `TapBranch` (same as BIP 341)

### Differences from P2TR:

| Aspect | P2TR (BIP 341) | P2MR (BIP 360) |
|--------|----------------|-----------------|
| SegWit version | 1 (`OP_1`) | 2 (`OP_2`) |
| Address prefix | `bc1p` | `bc1z` |
| Key-path spend | Yes | No |
| Output commitment | `tweakKey(internalPubkey, merkleRoot)` | `taggedHash("TapBranch", merkleRoot)` |
| Control block size | `33 + 32*m` bytes (includes internal pubkey) | `1 + 32*m` bytes (no internal pubkey) |
| Control block parity bit | 0 or 1 | Always 1 |
| Witness (script-path) | `[stack..., script, controlBlock]` | Same |

## Suggested Implementation Scope

Based on the existing P2TR implementation in this library, P2MR support would involve:

### New files:
- **`ts_src/payments/p2mr.ts`** — P2MR payment type (modeled after `p2tr.ts`, script-path only)

### Modified files:
- **`ts_src/payments/index.ts`** — Export `p2mr`
- **`ts_src/payments/bip341.ts`** — Add P2MR-specific helpers (e.g., control block without internal pubkey, merkle root commitment without key tweaking)
- **`ts_src/address.ts`** — Add P2MR in `toOutputScript()` / `fromOutputScript()` (SegWit v2 → p2mr)
- **`ts_src/psbt/bip371.ts`** or new file — P2MR PSBT handling (`isP2MR`, finalization, signing)
- **`ts_src/psbt.ts`** — P2MR signing/finalization support
- **`ts_src/transaction.ts`** — P2MR sighash computation (if it differs from v1)

### Reusable existing code:
- `tapleafHash`, `tapBranchHash`, `toHashTree`, `findScriptPath` from `bip341.ts`
- Tagged hash infrastructure from `crypto.ts`
- Bech32m encoding/decoding (already supports version ≥ 1)
- Tapscript validation logic

## References

- **BIP 360 Specification**: https://bip360.org/bip360.html
- **BIP 360 on GitHub**: https://github.com/bitcoin/bips/blob/master/bip-0360.mediawiki
- **BIP 360 PR**: https://github.com/bitcoin/bips/pull/1670
- **Test vectors**: https://github.com/bitcoin/bips/blob/master/bip-0360/ref-impl/common/tests/data/P2MR_construction.json
- **Reference implementations**: [Rust](https://github.com/bitcoin/bips/tree/master/bip-0360/ref-impl/rust), [Python](https://github.com/bitcoin/bips/tree/master/bip-0360/ref-impl/python)

## Additional Context

P2MR is designed as a conservative first step toward quantum resistance. While the BIP is still in Draft status and has not been activated on the Bitcoin network, early library support would:

1. Allow developers to experiment with P2MR transaction construction and validation
2. Prepare the ecosystem for potential future activation
3. Provide a foundation for future post-quantum signature integration via tapscript `OP_SUCCESSx` upgrades

jasonandjay:
## Summary

Implements [BIP 360](https://github.com/bitcoin/bips/tree/master/bip-0360) Pay-to-Merkle-Root (P2MR) payment type support. P2MR is a SegWit version 2 output type that commits directly to a script tree's Merkle root without an internal public key, supporting script-path spending only. This provides resistance to long-exposure quantum attacks by removing the key-path spend vector present in P2TR.

Closes #2311

## Changes

**Core implementation:**
- Add `p2mr` payment module (`ts_src/payments/p2mr.ts`) — address encoding (bc1z), scriptPubKey (OP_2), witness construction, and validation
- Add `rootHashFromP2MRPath` to bip341 utilities for P2MR control blocks (no internal pubkey, Merkle path starts at byte offset 1)
- Update `address.ts` to support SegWit v2 encoding/decoding
- Extend PSBT `bip371.ts` to recognize P2MR inputs/outputs via `isTaprootInput`/`isTaprootOutput`
- Add `isP2MR` detection to `psbtutils.ts`

**Key differences from P2TR:**
| | P2TR (BIP 341) | P2MR (BIP 360) |
|---|---|---|
| Witness version | v1 (`OP_1`) | v2 (`OP_2`) |
| Address prefix | `bc1p` | `bc1z` |
| Internal pubkey | Required | None |
| Key-path spend | Yes | No (script-path only) |
| Control block | `[ctrl_byte][32B pubkey][path...]` | `[ctrl_byte][path...]` |

**Tests:**
- 22 unit tests covering all 6 official BIP 360 valid test vectors (address, scriptPubKey, merkleRoot) + 12 witness/controlBlock tests for every leaf in each tree topology + 6 negative tests
- 9 integration tests covering full PSBT workflow (sign → finalize → extract), multi-leaf trees, PSBT serialization roundtrip, key-path rejection, and P2MR vs P2TR comparison

## Notes

- The official BIP 360 test vector for `p2mr_different_version_leaves` has a known bug: control byte `0xC1` instead of the correct `0xFB` for leaf version 250. Our tests use the correct value.
- P2MR transactions cannot be broadcast until BIP 360 activates on mainnet, so integration tests verify the full PSBT workflow locally.
- PSBT signing/finalization reuses existing P2TR script-path code paths — P2MR inputs naturally lack `tapInternalKey` and `tapKeySig`, so the existing logic correctly falls into script-path-only execution.

## Test plan

- [x] All 2695 existing unit tests pass
- [x] 22 new P2MR unit tests pass (BIP 360 test vectors)
- [x] 9 new P2MR integration tests pass
- [x] No regressions in address, PSBT, or P2TR tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,8 @@ Some examples interact (via HTTPS) with a 3rd Party Blockchain Provider (3PBP).
 - [Create (and broadcast via 3PBP) a taproot script-path spend Transaction - OP_CHECKSIG](https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/test/integration/taproot.spec.ts)
 - [Create (and broadcast via 3PBP) a taproot script-path spend Transaction - OP_CHECKSEQUENCEVERIFY](https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/test/integration/taproot.spec.ts)
 - [Create (and broadcast via 3PBP) a taproot script-path spend Transaction - OP_CHECKSIGADD (3-of-3)](https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/test/integration/taproot.spec.ts)
+- [P2MR (BIP360) address generation and script-path spend](https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/test/integration/p2mr.spec.ts)
+- [P2MR multi-leaf script tree with PSBT signing](https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/test/integration/p2mr.spec.ts)
 - [Generate a random address](https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/test/integration/addresses.spec.ts)
 - [Import an address via WIF](https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/test/integration/addresses.spec.ts)
 - [Generate a 2-of-3 P2SH multisig address](https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/test/integration/addresses.spec.ts)
--- a/src/cjs/address.cjs
+++ b/src/cjs/address.cjs
@@ -201,6 +201,9 @@ function fromOutputScript(output, network) {
     return payments.p2tr({ output, network }).address;
   } catch (e) {}
   try {
+    return payments.p2mr({ output, network }).address;
+  } catch (e) {}
+  try {
     return _toFutureSegwitAddress(output, network);
   } catch (e) {}
   throw new Error(bscript.toASM(output) + ' has no matching Address');
@@ -239,6 +242,11 @@ function toOutputScript(address, network) {
       } else if (decodeBech32.version === 1) {
         if (decodeBech32.data.length === 32)
           return payments.p2tr({ pubkey: decodeBech32.data }).output;
+      } else if (
+        decodeBech32.version === 2 &&
+        decodeBech32.data.length === 32
+      ) {
+        return payments.p2mr({ hash: decodeBech32.data }).output;
       } else if (
         decodeBech32.version >= FUTURE_SEGWIT_MIN_VERSION &&
         decodeBech32.version <= FUTURE_SEGWIT_MAX_VERSION &&
--- a/src/cjs/address.d.ts
+++ b/src/cjs/address.d.ts
@@ -17,11 +17,11 @@ export interface Base58CheckResult {
 }
 /** bech32 decode result */
 export interface Bech32Result {
-    /** address version: 0x00 for P2WPKH、P2WSH, 0x01 for P2TR*/
+    /** address version: 0x00 for P2WPKH/P2WSH, 0x01 for P2TR, 0x02 for P2MR */
     version: number;
-    /** address prefix: bc for P2WPKH、P2WSH、P2TR */
+    /** address prefix: bc for P2WPKH/P2WSH/P2TR/P2MR */
     prefix: string;
-    /** address data：20 bytes for P2WPKH, 32 bytes for P2WSH、P2TR */
+    /** address data：20 bytes for P2WPKH, 32 bytes for P2WSH/P2TR/P2MR */
     data: Uint8Array;
 }
 /**
--- a/src/cjs/payments/bip341.cjs
+++ b/src/cjs/payments/bip341.cjs
@@ -46,11 +46,13 @@ var __importStar =
 Object.defineProperty(exports, '__esModule', { value: true });
 exports.MAX_TAPTREE_DEPTH = exports.LEAF_VERSION_TAPSCRIPT = void 0;
 exports.rootHashFromPath = rootHashFromPath;
+exports.rootHashFromP2MRPath = rootHashFromP2MRPath;
 exports.toHashTree = toHashTree;
 exports.findScriptPath = findScriptPath;
 exports.tapleafHash = tapleafHash;
 exports.tapTweakHash = tapTweakHash;
 exports.tweakKey = tweakKey;
+exports.tapBranchHash = tapBranchHash;
 const ecc_lib_js_1 = require('../ecc_lib.cjs');
 const bcrypto = __importStar(require('../crypto.cjs'));
 const bufferutils_js_1 = require('../bufferutils.cjs');
@@ -75,6 +77,36 @@ function rootHashFromPath(controlBlock, leafHash) {
   let kj = leafHash;
   for (let j = 0; j < m; j++) {
     const ej = controlBlock.slice(33 + 32 * j, 65 + 32 * j);
+    if (tools.compare(kj, ej) < 0) {
+      kj = tapBranchHash(kj, ej);
+    } else {
+      kj = tapBranchHash(ej, kj);
+    }
+  }
+  return kj;
+}
+/**
+ * Calculates the root hash from a P2MR control block and leaf hash.
+ * Unlike P2TR, P2MR control blocks do not include an internal public key,
+ * so the Merkle path starts at byte offset 1 (after the control byte).
+ * @param controlBlock - The P2MR control block buffer (1 + 32*m bytes).
+ * @param leafHash - The leaf hash buffer.
+ * @returns The root hash buffer.
+ * @throws {TypeError} If the control block length is invalid.
+ */
+function rootHashFromP2MRPath(controlBlock, leafHash) {
+  if (controlBlock.length < 1)
+    throw new TypeError(
+      `The control-block length is too small. Got ${controlBlock.length}, expected min 1.`,
+    );
+  if ((controlBlock.length - 1) % 32 !== 0)
+    throw new TypeError(
+      `The control-block length of ${controlBlock.length} is incorrect for P2MR!`,
+    );
+  const m = (controlBlock.length - 1) / 32;
+  let kj = leafHash;
+  for (let j = 0; j < m; j++) {
+    const ej = controlBlock.slice(1 + 32 * j, 33 + 32 * j);
     if (tools.compare(kj, ej) < 0) {
       kj = tapBranchHash(kj, ej);
     } else {
--- a/src/cjs/payments/bip341.d.ts
+++ b/src/cjs/payments/bip341.d.ts
@@ -29,6 +29,16 @@ export type HashTree = HashLeaf | HashBranch;
  */
 export declare function rootHashFromPath(controlBlock: Uint8Array, leafHash: Uint8Array): Uint8Array;
 /**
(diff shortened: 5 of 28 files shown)


duncan0k:
Thanks for putting this together. I compared selected construction fields from this PR's fixtures against an independent TypeScript harness aligned to BIP-360 snapshot `620871a7a442e276a058b487cd8743775fb499a4` (the harness is itself differentially tested against the Python reference in that repo). Line numbers below refer to [`bip-0360.mediawiki` at that snapshot](https://github.com/bitcoin/bips/blob/620871a7a442e276a058b487cd8743775fb499a4/bip-0360.mediawiki).

Scope note up front: the harness recorded 89 matching comparisons across the 22 `valid` fixtures (address, merkle root, output, leaf script, control block); for the 6 `invalid` fixtures it recorded three exceptions, one witness-shape rejection and two adapter-detected mismatches. I did not execute the bitcoinjs-lib implementation in that comparison; the code observations below come from reading the diff plus targeted hash recomputation. Corrections welcome.

For the inspected fixtures, the control-block layout (`1 + 32m`, no internal pubkey), `c[0] = v | 1`, `v = c[0] & 0xfe`, the TapLeaf/TapBranch reuse, `OP_2 OP_PUSHBYTES_32 <root>` and the bech32m witness-v2 encoding all match the snapshot. For the inspected script-path flow without an annex or an executed `OP_CODESEPARATOR`, the existing `hashForWitnessV1` supplies the BIP-342 extension (`ext_flag = 1`, `key_version = 0x00`) and `hashScriptPubKeys` commits to the actual prevout script, so the `OP_2` program is committed. A P2MR digest regression test would pin that integration.

Points worth addressing before merge, roughly by impact:

**1. Single-leaf (depth-zero) trees are anyone-can-spend under the pinned text, and the PR constructs them without a diagnostic.** (`ts_src/payments/p2mr.ts:92-96`, `155-173`)

The Design section describes depth-zero trees as anyone-can-spend (line 155), and Script Validation skips script execution once the root matches with `m = 0` (lines 251–252). The changelog dates this to 0.12.0 (2026-06-18, line 422), after this PR's last update, so I read it as the spec moving rather than an oversight. Supplying a single-leaf `scriptTree` currently constructs the output silently; adding `redeem: { output: script }` yields the witness `[script, c1]` for the default leaf version. Seven of the nine integration tests construct single-leaf trees, and both README links point to that test file. Suggestion: a constructor guard or `console.warn` (precedent: `FUTURE_SEGWIT_VERSION_WARNING`, `ts_src/address.ts:78`), wallet examples with at least two leaves (line 171 encourages depth ≥ 1), and keep the upstream depth-zero vectors as explicitly labelled unsafe construction cases.

**2. Leaf version `0` is hashed as `0xC0` but encoded as `0x01` in the control byte.** (`ts_src/payments/bip341.ts:152`, `ts_src/payments/p2mr.ts:132,169,227`)

`tapleafHash` uses `leaf.version || LEAF_VERSION_TAPSCRIPT`, so a tree leaf with `version: 0` is hashed as `0xC0`, while a nested `redeem.redeemVersion: 0` is preserved (`:133-136`) and `c[0] = version | 1` produces `0x01` (`:169`). A verifier following lines 243–244 and 251 recomputes the leaf hash with version 0 and lands on a different root, so a `scriptTree` leaf `{ output, version: 0 }` spent with `redeem: { output, redeemVersion: 0 }` produces a witness that does not match the output it was built for. The same falsy pattern appears at `:132` and `:227`. Relatedly, an odd `redeemVersion` misses the lookup when the supplied tree holds even-version leaves, but that does not establish safe rejection for every input form (`hash + redeem` bypasses the tree). Suggestion: preserve explicit zero versions throughout (`!== undefined`), and add an explicit integer / byte-range / parity check on `redeemVersion` with its own message; the footnote to line 243 explains why the low bit is forced.

**3. `isTapLeafInTree` now accepts either control-block interpretation, which widens an existing P2TR check.** (`ts_src/psbt/bip371.ts:500-524`)

When `tapMerkleRoot` is supplied, the membership check tries the P2TR rule and then the P2MR rule. For a P2TR input whose declared root is `TapBranch(H0, H1)`, a 33-byte control block `c1 || H1` yields `H0` under the previous check (rejected) and the declared root under the new P2MR fallback (accepted). This is a PSBT metadata check rather than a consensus path, but it now accepts control blocks it used to refuse for inputs unrelated to P2MR. Suggestion: resolve the spent output's script from the available UTXO data (`isP2TR` / the `isP2MR` this PR adds in `psbtutils.ts`) and apply exactly one rule; incomplete PSBTs need explicit handling until that script is available.

**4. Root consistency is not cross-checked across input sources.** (`ts_src/payments/p2mr.ts:177-212,288,293`)

The root derived from `address` / `output` is kept in a local `hash`, while the `scriptTree` root is compared against `a.hash` (`:209`) and not against that local value. So `scriptTree` A plus `address` B, `scriptTree` A plus `output` B, or `scriptTree` A plus a witness from another tree (with no address / output / hash supplied, `:288`) can all return an object whose fields disagree. Suggestion: merge every supplied source into one root check and add combined-input negative fixtures.

**5. `findScriptPath` keys on the leaf hash, so a duplicate leaf cannot select its occurrence.** (`ts_src/payments/bip341.ts:130-145`, via `ts_src/payments/p2mr.ts:163`)

The hash-based lookup returns one matching path without exposing a choice of leaf occurrence. A port of this PR's traversal and control-block assembly matched all seven roots and 15 of the 16 control blocks in the successful upstream construction vectors; the remaining difference is `p2mr_duplicate_leaves`, where the expected 33-byte proof at position 0 comes back as the 65-byte proof of the other occurrence. Both proofs reconstruct the root, so this is an API-capability point rather than a funds risk: the shorter proof for a given occurrence is not reachable, and the witness carries 32 extra bytes. Suggestion: an optional occurrence selector, or a helper returning all matching paths, while preserving the current default behaviour.

**6. Fixture notes.**

- The `valid[13]` annotation ("BIP360 test vector has bug: uses 0xC1") is stale relative to the pinned snapshot. Its hardcoded `0xFB` control byte already matches that snapshot and the expected encoding for leaf version `0xFA`; refreshing the annotation to name the vector revision it refers to would avoid a future "fix" in the wrong direction.
- Absent from the fixtures: `p2mr_duplicate_leaves` and the six successful `p2mr_pqc_construction.json` cases. Missing-tree rejection already has a counterpart in `invalid[0]`; explicit null and empty-tree cases would make that coverage clearer.
- The control-block checks the code already implements have no negative fixtures: even `c[0]` (`:279`), length not `1 + 32m` (`:267`), `m > 128` (`:273`).
- When a usable `scriptTree` is supplied, I found no explicit rejection of accompanying `pubkey` / `internalPubkey` fields (the schema at `:53-74` does not list them and `Object.assign(o, a)` at `:293` copies them onto the result); the upstream internal-pubkey-plus-null-tree input already reaches `Not enough data`. A dedicated diagnostic and fixture would make this API boundary clearer.
- The added tests do not assert a known P2MR signature digest; an independently computed digest expectation would catch a future change in `transaction.ts`.

**Smaller notes**

- `ts_src/payments/p2mr.ts:81-90` strips the annex at `witness.length >= 2`. BIP-360 rejects a two-element witness whose final element begins with `0x50` (line 238) and recognises an annex at three or more elements (line 239). The code still rejects the two-element case, but reports a key-path error; an explicit check would clarify the diagnostic. (P2TR's `>= 2` serves its key-path case.)
- Routing witness-v2 addresses to `p2mr` (`ts_src/address.ts:199-206,248-256`) preserves the encoded bytes for valid v2 + 32-byte programs and removes the future-segwit diagnostic. Since the pinned BIP is marked Draft, it is worth stating whether dropping that diagnostic is intentional.
- The nine offline tests in `test/integration/p2mr.spec.ts` sit outside the default unit-test glob (`test/*.ts`). The fixture suite is registered in `test/payments.spec.ts`; moving the offline cases into the default suite would also cover the PSBT examples.
- The README links point to `blob/master/test/integration/p2mr.spec.ts`, a file this PR adds.
- Pinning the bips commit this implementation targets in the `p2mr.ts` header would help readers, since 0.12.0 and 0.12.1 both changed behaviour this PR touches.

The construction figures come from the adapted fixture harness and a port of the tree traversal; the PSBT observations are from source inspection and targeted root calculations. I have not run the bitcoinjs-lib test suite. The harness is at https://github.com/duncan0k/p2mr-ts (construction and digests, no network validation). Happy to be corrected on any of this.

PR #2320, opened 2026-03-01
Fix #2249: Use slice() instead of subarray() to preserve Buffer type

### LINKED ISSUE #2249: Buffer.equals() requires Buffer, not Uint8Array

R-Panic:
Fixes #2249

## Summary

Fixed Buffer type mismatch issue in PSBT Taproot handling where `Buffer.subarray()` was returning `Uint8Array` instead of `Buffer`, causing `Buffer.equals()` comparisons to fail even when byte content is identical.

## Changes

- Changed `script.subarray(2, 34)` to `script.slice(2, 34)` in `getPrevoutTaprootKey()`
- Changed `signature.subarray(0, 64)` to `signature.slice(0, 64)` in `trimTaprootSig()`

## Root Cause

Node.js `Buffer.subarray()` returns a `Uint8Array` view of the underlying buffer, while `Buffer.slice()` returns a new `Buffer` with its own backing. The issue comment in #2249 explains:
```typescript
Buffer.slice() --> returns a Buffer
Buffer.subarray() --> returns a Uint8Array
```

## Impact

- Fixes spurious `Buffer.equals()` failures in BIP371 PSBT Taproot signature verification
- Ensures consistent type (Buffer) when comparing Taproot output keys
- No behavioral changes - both methods produce identical byte content

## Testing

Manual verification: The fix matches the exact solution provided by the issue reporter and addresses both occurrences identified in the codebase.


junderw:
We are migrating away from Buffer.

The solution is not to move back to Buffer, it's to fix the BIP371 code to not use Buffer.equals().

junderw:
Oh, I see. The issue is from an old version.

The current version no longer has the same issue. If it does, let me know.

PR #2326, opened 2026-03-17
fix: use Buffer.slice() instead of subarray() in getPrevoutTaprootKey

lindsey-comet:
## Summary
- Fixes issue #2249 where `Buffer.equals()` requires `Buffer`, not `Uint8Array`
- Changed `script.subarray(2, 34)` to `script.slice(2, 34)` in `getPrevoutTaprootKey()`
- Node.js `Buffer.subarray()` returns `Uint8Array`, while `Buffer.slice()` returns `Buffer`
- This ensures the returned value can be compared using `.equals()` with other `Buffer` values

## Test Plan
- All 2660 existing tests pass
- Build completes successfully

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jvgelder:
should use https://github.com/bitcoinjs/bitcoinjs-lib/pull/2322 as a base for now ?

PR #2329, opened 2026-03-26
feat: add taproot support for HD signing methods (signInputHD, signAllInputsHD)

Crystora:
## Summary

Taproot inputs with `tapBip32Derivation` now work with `signInputHD`, `signInputHDAsync`, `signAllInputsHD`, and `signAllInputsHDAsync`. Previously these methods only handled legacy/segwit via `bip32Derivation` and would fail on taproot inputs.

- `getSignersFromHD` detects taproot inputs and delegates to a new `getTaprootSignersFromHD` function
- Key-path spend (empty `leafHashes`): derives child key, tweaks it using `tapMerkleRoot` from the PSBT input
- Script-path spend (non-empty `leafHashes`): derives child key untweaked, existing `_signTaprootInput` handles the rest
- `HDSigner` / `HDSignerAsync` get optional `signSchnorr?` and `tweak?` - existing code unaffected
- `sighashTypes` on HD methods changed from `= [SIGHASH_ALL]` to optional, matching `signAllInputs` - lets inner methods pick the right default per input type

Closes #2132

PR #2331, opened 2026-04-05
feat(psbt): add taproot support for HD signing methods

### LINKED ISSUE #2132: signAllInputsHD not working for taproot key

paulee:
## Summary

`signInputHD` / `signAllInputsHD` (and their async variants) silently fail on taproot inputs because they only look at `bip32Derivation`, not `tapBip32Derivation`. This patch teaches the HD signing flow to recognize taproot inputs and derive the correct signers.

### What changed

- `getSignersFromHD` now detects taproot inputs via `isTaprootInput()` and delegates to a new `getTaprootSignersFromHD` helper
- **Key-path spend** (`leafHashes` is empty): derives the child key from `tapBip32Derivation`, then tweaks it with `tapTweakHash(internalKey, tapMerkleRoot)` — the tweaked signer produces a valid `tapKeySig`
- **Script-path spend** (`leafHashes` is non-empty): derives the child key *untweaked* and hands it off to the existing `_signTaprootInput` machinery, which matches the pubkey against `tapLeafScript` entries
- `sighashTypes` parameter on all four HD methods changed from `= [SIGHASH_ALL]` to `optional` — this removes a conflict where taproot's `SIGHASH_DEFAULT (0x00)` was being rejected by a hardcoded `SIGHASH_ALL` filter. Inner methods (`_signInput` / `_signTaprootInput`) already default correctly per input type
- `HDSigner` / `HDSignerAsync` interfaces extended with optional `signSchnorr?` and `tweak?` — no breaking change for existing implementations

### Why this approach

PR #2329 tackled the same problem. This takes a slightly different approach:

1. Only modifies `ts_src/psbt.ts` — compiled `src/cjs/` and `src/esm/` output is generated by the build, not hand-edited
2. Cleaner filtering — `.filter()` instead of `.map().filter()` for fingerprint matching
3. Avoids redundant `checkForInput` call — passes the already-resolved input directly to `getTaprootSignersFromHD`
4. Uses `Uint8Array` consistently in tests (v7-compatible)

## Test plan

- [x] Key-path signing with `signInputHD`
- [x] Key-path signing with merkle root present (script tree exists, but signing via key path)
- [x] Script-path signing with `signInputHD`
- [x] Mixed taproot + legacy (P2WPKH) inputs with `signAllInputsHD`
- [x] Error: missing `tapBip32Derivation`
- [x] Error: fingerprint mismatch
- [x] Error: pubkey / derivation path mismatch
- [x] All 2667 existing tests still pass

Closes #2132


paulee:
@jasonandjay @junderw Would you check this PR?

PR #2333, opened 2026-04-15
fix(p2ms): fix signature validation bypass due to operator precedence

deepview-autofix:
The expression `(opts.allowIncomplete && x === OPS.OP_0) !== undefined`
always evaluated to `true` when `allowIncomplete` was set, because
`!== undefined` was applied to the boolean result of `&&` (which is
never `undefined`). This caused `isAcceptableSignature()` to accept any
arbitrary buffer as a valid signature when `allowIncomplete` was enabled,
completely bypassing signature validation.

Fixed by replacing `!== undefined` with `!!` to correctly coerce the
result to boolean without altering the intended logic: accept only
canonical script signatures or OP_0 placeholders (when allowIncomplete).


kewdex (approved):
utACK

